PR #20, opened 2025-11-05
Add support for streaming JSON formatted records.

elenagaljak-db:

Added support for ingest records in JSON format. This approach significantly simplifies initial client setup. For production work it is still recommended to use Protobuf as it is faster.

## How is this tested?

Added one test to tests suite.

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -16,7 +16,7 @@ documentation = "https://docs.rs/databricks-zerobus-ingest-sdk"
 async-trait = "0.1"
 prost = "0.13.3"
 prost-types = "0.13.3"
-reqwest = { version = "0.12", features = ["json", "rustls-tls"] }
+reqwest = { version = "0.12", features = ["json", "rustls-tls"], default-features = false }
 serde_json = "1.0"    
 thiserror = "1.0"
 tokio = { version = "1.42.0", features = ["macros", "rt-multi-thread", "fs"] }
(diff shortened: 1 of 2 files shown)


examples/README.md
```diff
@@ -238,68 +373,71 @@ let mut stream = sdk_handle
...
-    updated_at: Some(chrono::Utc::now().timestamp()),
+for i in 0..100 {
+    let json_record = format!(
+        r#"{{"id": {}, "customer_name": "Customer {}"}}"#,
+        i, i
+    );
+    let ack_future = stream.ingest_record(json_record).await?;
+
+    // Option 1: Fire and forget
+    tokio::spawn(ack_future);
```
teodordelibasic-db: Why is this needed?

teodordelibasic-db (approved):
Looks good, we could just do with covering error paths in unit tests.

sdk/src/lib.rs
```diff
@@ -729,10 +774,28 @@ impl ZerobusStream {
             .map_err(ZerobusError::CreateStreamError)?
             .into_inner();
 
+        let descriptor_proto = match record_type {
+            RecordType::Proto => Some(
+                table_properties
+                    .descriptor_proto
+                    .as_ref()
+                    .ok_or(ZerobusError::InvalidArgument(
```
teodordelibasic-db: I am thinking how we can have this check at the start of `create_stream_with_headers_provider` in order to fail as soon as possible.

PR #25, opened 2025-11-25
Add batch ingestion support

pavledjuric-db:

This PR adds batch ingestion support to the Zerobus SDK, enabling clients to ingest multiple records in a single API call with improved throughput and all-or-nothing semantics.

**Key Changes:**

- Added `ingest_batch()` method to `ZerobusStream` that accepts iterables of proto or JSON records
- Implemented all-or-nothing semantics: entire batch succeeds or fails as a unit
- Updated `zerobus_service.proto` with `IngestRecordBatchRequest` and related batch messages
- Merged codepaths for record and batch ingestion
- Used `smallvec` to avoid heap allocations for single records & single-record batches
- Changed `get_unacked_records` to preserve batch structure instead of flattening batches into individual records
- **Changed backpressure mechanism to track in-flight requests instead of in-flight records** for more accurate flow control with batch operations

**Documentation & Examples:**

- Added `basic_example_json_batch/` - demonstrates batch ingestion with JSON serialization
- Added `basic_example_proto_batch/` - demonstrates batch ingestion with Protocol Buffers
- Updated examples README with comprehensive guidance on choosing between single-record and batch ingestion
- Documented all-or-nothing semantics and performance considerations

## How is this tested?

Tested with 15+ new test cases covering:

- Basic batch ingestion (small and large batches up to 100 records)
- Concurrent batch ingestion from multiple tasks
- Backpressure enforcement with in-flight request tracking (not record counts)
- Type validation (rejecting mismatched proto/JSON batches)
- Recovery flow preserving batch structure
- Mixed single-record and batch operations
- All-or-nothing semantics verification

All tests use the existing mock gRPC infrastructure.


sdk/src/landing_zone.rs
```diff
@@ -4,12 +4,32 @@ use std::sync::Arc;
 use thiserror::Error;
 use tokio::sync::{Notify, OwnedSemaphorePermit, Semaphore};
 
+use crate::{ZerobusError, ZerobusResult};
+
 #[derive(Debug, Error)]
 pub enum LandingZoneError {
     #[error("Attempted to remove non-observed element")]
     RemovingNonObservedElement,
 }
```
elenagaljak-db: I think it would be best if we could not import anything from the _core zerobus logic_ in landing zone, since it is supposed to be an independent module. I would remove this from here and add a new error to the `LandingZoneError` enum below that we could catch in `ingest_record` and convert to `ZerobusError`.

sdk/src/landing_zone.rs
```diff
@@ -75,7 +75,7 @@ impl<T: Clone> LandingZone<T> {
     pub fn remove_all(&self) -> Vec<T> {
         let mut state = self.state.lock().expect("Lock poisoned");
 
-        let mut all_items = Vec::with_capacity(state.observed_items.len() + state.queue.len());
+        let mut all_items = Vec::with_capacity(state.queue.len() + state.observed_items.len());
```
teodordelibasic-db: Interesting :D

sdk/src/lib.rs
```diff
@@ -77,49 +80,142 @@ pub type ProtoEncodedRecord = Vec<u8>;
...
+                    offset_id: Some(offset_id),
+                })
+            }
+        }
+    }
 }
 
-impl IntoRecord for RecordPayload {
-    fn into_record(self) -> Record {
+impl EncodedBatch {
```
teodordelibasic-db: Duplicated `impl` block I think.

teodordelibasic-db (commented):
Looks good, mainly concerned with the misinterpretation of a batch of size 1.

sdk/src/lib.rs
```diff
@@ -77,49 +80,142 @@ pub type ProtoEncodedRecord = Vec<u8>;
...
+                    },
+                )
+                .map(EncodedBatch::Proto),
+            _ => None, // todo
+        }
+    }
+
+    fn into_request_payload(self, offset_id: OffsetId) -> RequestPayload {
+        match self {
+            EncodedBatch::Proto(records) if records.len() == 1 => {
```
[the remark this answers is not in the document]
pavledjuric-db: semantically there are no differences at all.. Either way, back on the server, single records are treated as batches of size 1

sdk/src/lib.rs
```diff
@@ -890,33 +989,110 @@ impl ZerobusStream {
...
+    /// the entire batch has been durably written.
+    ///
+    /// # Arguments
+    ///
+    /// * `payload` - An iterator of protobuf-encoded records (each item should be convertible to `EncodedRecord`)
+    ///
+    /// # Returns
+    ///
+    /// A future that resolves to the logical offset ID of the last acknowledged batch.
+    /// If the batch is empty, and nothing has been ingested yet, the future resolves to -1.
```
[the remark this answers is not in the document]
elenagaljak-db: I thought it started from 0??

sdk/src/stream_configuration.rs
```diff
@@ -11,22 +11,22 @@ use crate::databricks::zerobus::RecordType;
...
 /// };
 /// ```
 #[derive(Clone)]
 pub struct StreamConfigurationOptions {
-    /// Maximum number of records that can be sending or pending acknowledgement at any given time.
+    /// Maximum number of requests that can be sending or pending acknowledgement at any given time.
     ///
     /// This limit controls memory usage and backpressure. When this limit is reached,
     /// `ingest_record()` calls will block until acknowledgments free up space.
     ///
```
elenagaljak-db: nit: ingest_record and ingest_records

sdk/src/lib.rs
```diff
@@ -890,33 +1024,109 @@ impl ZerobusStream {
     /// ```
     pub async fn ingest_record(
         &self,
-        payload: impl Into<RecordPayload>,
+        payload: impl Into<EncodedRecord>,
+    ) -> ZerobusResult<impl Future<Output = ZerobusResult<i64>>> {
```
elenagaljak-db: Can you use type `OffsetId` instead of i64

sdk/src/lib.rs
```diff
@@ -890,33 +1024,109 @@ impl ZerobusStream {
...
+    ///     Some(offset) => println!("Batch written at offset: {}", offset),
+    ///     None => println!("Empty batch - no records written"),
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn ingest_records<I, T>(
+        &self,
+        payload: I,
+    ) -> ZerobusResult<impl Future<Output = ZerobusResult<Option<i64>>>>
```
elenagaljak-db: Here too, `OffsetId`

PR #27, opened 2025-12-16
Graceful stream close handling

elenagaljak-db:

Right now in Zerobus proto contract there is a stream closing message which returns a Duration in ms of how much time will pass until server forcefully closes the stream.
Currently, in the SDK we handle this as a typical retryable error, we shut down current stream immediately and open a new one, resending all unacked records.
The idea behind the new approach would be to not break the stream immediately, but rather wait for acks in that duration, and pause sending new records. If we receive all acks or the time is up, we will trigger recovery and proceed as usual.

## How is this tested?

Added new test suite with 7 new tests:

- Default graceful close - waits for full server duration
- Immediate recovery
- Client max less than server
- Client max greater than server
- Early recovery - all acks received during pause
- Sending buffered records
- Error during pause

### DIFF
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -1483,6 +1483,7 @@ impl ZerobusStream {
                                 }
                             };
                             let mut last_logical_acked_offset = -2;
+                            let mut map = oneshot_map.lock().await;
                             for _offset_to_ack in
                                 (last_acked_offset + 1)..=durability_ack_up_to_offset
                             {
@@ -1490,12 +1491,12 @@ impl ZerobusStream {
                                     let logical_offset = record.offset_id;
                                     last_logical_acked_offset = logical_offset;
 
-                                    let mut map = oneshot_map.lock().await;
                                     if let Some(sender) = map.remove(&logical_offset) {
                                         let _ = sender.send(Ok(logical_offset));
                                     }
                                 }
                             }
+                            drop(map);
                             last_acked_offset = durability_ack_up_to_offset;
                             if last_logical_acked_offset != -2 {
                                 let _ignore_on_channel_break = last_received_offset_id_tx


sdk/src/lib.rs
```diff
@@ -1246,9 +1300,12 @@ impl ZerobusStream {
                         return Err(ZerobusError::StreamClosedError(status));
                     }
                     Err(_timeout) => {
-                        // No message received for ack_timeout_ms
-                        if !landing_zone.is_observed_empty() {
-                            error!("Server ack timeout: no response for {}ms", ack_timeout_ms);
+                        // No message received for server_lack_of_ack_timeout_ms.
+                        if pause_deadline.is_none() && !landing_zone.is_observed_empty() {
```
teodordelibasic-db: Why did we add the `pause_deadline.is_none()` clause?

sdk/src/lib.rs
```diff
@@ -1264,10 +1321,16 @@ impl ZerobusStream {
     fn spawn_sender_task(
```
teodordelibasic-db: `Zerobus` instead of `Ingest API` on line 1319, a remnant of the past. :-)

NEXT_CHANGELOG.md
```diff
@@ -25,10 +25,12 @@
 
 ### Internal Changes
 
-- [**BREAKING**] Changed backpressure mechanism to track in-flight requests instead of in-flight records
+- Improved graceful close mechanism: when server signals stream closure, SDK now continues processing acknowledgments for in-flight records while pausing new record transmission until timeout.
+- [**BREAKING**] Changed backpressure mechanism to track in-flight requests (individual records + batches) instead of individual record count
 
 ### API Changes
 
+- [**BREAKING**] Added `stream_paused_max_wait_time_ms` to `StreamConfigurationOptions` to configure maximum wait time during graceful stream close (`None` = wait for full server duration, `Some(0)` = immediate recovery, `Some(x)` = wait up to min(x, server_duration) milliseconds)
```
teodordelibasic-db: I wouldn't say this is breaking?

sdk/src/lib.rs
```diff
@@ -810,15 +810,17 @@ impl ZerobusStream {
             landing_zone_recovery.reset_observe();
 
             // 3. Spawn receiver and sender task.
+            let is_paused = Arc::new(AtomicBool::new(false));
```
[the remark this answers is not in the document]
elenagaljak-db: We already had is_closed so I didn't want to add more fields. I think we should restructure this anyway, maybe as a part of that code redesign PR.

sdk/src/lib.rs
elenagaljak-db: Fixed

NEXT_CHANGELOG.md
teodordelibasic-db: Fair enough.

teodordelibasic-db (approved):
Looks good, let's just wait for Nikola to merge his changes.

PR #42, opened 2025-12-29
Add alternative ingestion API with direct offset return

elenagaljak-db:

Adds alternative ingestion API methods (`ingest_record_offset()` and `ingest_records_offset()`) that return logical offset IDs directly instead of wrapped in Futures. Includes new `wait_for_offset()` method to explicitly wait for acknowledgment of specific offsets. Refactors `flush()` and `wait_for_offset()` to share common waiting logic.
These are benchmarks runs with ingest_record vs ingest_record_offset
[results.rtf](https://github.com/user-attachments/files/24613254/something.rtf)
## How is this tested?

Tests Added:

- test_ingest_record_offset_single_record - Basic single record ingestion
- test_ingest_record_offset_multiple_records - Multiple sequential records (10 records)
- test_ingest_records_offset_batch - Batch ingestion (3 records in one batch)
- test_ingest_records_offset_empty_batch - Empty batch edge case (returns None)
- test_mixed_ingest_record_api_usage - Mixing both APIs (backward compatibility)
- test_ingest_record_offset_with_json - JSON record type with ingest_record_offset
- test_ingest_records_offset_with_json_batch - JSON batch with ingest_records_offset
- test_ingest_record_offset_after_close_fails - Error handling after close
- test_ingest_records_offset_after_close_fails - Batch error handling after close
- test_ingest_record_offset_blocks_on_inflight_limit - Backpressure/blocking behavior

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -15,7 +15,7 @@
 - **Deprecated `ingest_record()` and `ingest_records()` methods**: Use `ingest_record_offset()` and `ingest_records_offset()` instead
   - The new methods return offsets directly (after queuing) without Future wrapping for a cleaner API
   - Use with `wait_for_offset()` to explicitly wait for acknowledgments when needed
-  - Old methods will continue to work but may be removed in a future major version
+  - Old methods will continue to work but may be removed in a future version
 
 ### Bug Fixes
 


examples/basic_example_json/src/main.rs
```diff
@@ -62,10 +62,31 @@ async fn main() -> Result<(), Box<dyn Error>> {
         now, now
     );
 
+    // Example 1: Using v1 API, ingest_record returns a future.
     let ack_future = stream.ingest_record(json_record).await.unwrap();
+    let offset_id = ack_future.await.unwrap();
+    println!("Record acknowledged with offset Id: {}", offset_id);
+
+    // Example 2: Using v2 API, ingest_record_v2 returns offset immediately.
+    let json_record_2 = format!(
```
gotocoding-DB: This part is not in line with Example 1 – there, we define the json record first and then have the comment.

elenagaljak-db: Fixed

sdk/src/lib.rs
```diff
@@ -1039,6 +1039,55 @@ impl ZerobusStream {
...
+    /// // Ingest and get offset immediately
+    /// let offset = stream.ingest_record_v2(my_record).await?;
+    ///
+    /// // Later, wait for acknowledgment
+    /// stream.wait_for_offset(offset).await?;
+    /// println!("Record at offset {} has been acknowledged", offset);
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn ingest_record_v2(
```
teodordelibasic-db: Let's then just rename this to what Atomic proposed, mark previously existing methods as deprecated and change all the examples/documentation to primarily show this new API.

README.md
```diff
@@ -421,6 +421,43 @@ let record = YourMessage {
...
+// Wait for specific offsets.
+for offset in offsets {
+    stream.wait_for_offset(offset).await?;
+}
+```
+
+**When to use `ingest_record_offset()` vs `ingest_record()`:**
+- Use `ingest_record_offset()` when you want the logical offset as an integer and will call `wait_for_offset()` explicitly
+- Use `ingest_record()` when you want the Future returned directly for immediate chaining
+- Both methods have the same performance; the choice depends on your acknowledgment tracking pattern
```
gotocoding-DB: > Both methods have the same performance

Is it really correct? I expected `_offset` approach to be faster compared to spawning futures.

gotocoding-DB: Based on your numbers, there are a few that become much faster, including the 10Kb case.

Msg Size | Streams | Before (Rec/s) | After (Rec/s) | % Change
-- | -- | -- | -- | --
20 B | 1 | 769,891 | 1,163,572 | +51.1%
  | 2 | 1,430,806 | 2,014,497 | +40.8%
  | 4 | 2,532,661 | 3,267,739 | +29.0%
  | 8 | 3,410,913 | 3,565,046 | +4.5%
  | 16 | 3,690,303 | 4,054,323 | +9.8%
220 B | 1 | 666,951 | 970,554 | +45.5%
  | 2 | 1,232,906 | 1,561,771 | +26.7%
  | 4 | 1,856,131 | 2,095,292 | +12.9%
  | 8 | 1,910,700 | 2,021,231 | +5.8%
  | 16 | 1,838,483 | 2,189,667 | +19.1%
793 B | 1 | 75,388 | 74,920 | -0.6%
  | 16 | 141,466 | 140,395 | -0.7%
10.7 KB | 1 | 56,372 | 54,297 | -3.7%
  | 4 | 73,030 | 86,229 | +18.1%
  | 16 | 87,276 | 87,833 | +0.6%

sdk/src/lib.rs
```diff
@@ -1711,25 +1836,114 @@ impl ZerobusStream {
...
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[instrument(level = "debug", skip_all, fields(table_name = %self.table_properties.table_name))]
+    pub async fn flush(&self) -> ZerobusResult<()> {
+        if self.is_closed.load(Ordering::Relaxed) {
+            return Err(ZerobusError::StreamClosedError(tonic::Status::internal(
+                "Stream closed during flush.".to_string(),
+            )));
+        }
```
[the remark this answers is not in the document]
elenagaljak-db: It looks redundant but if we remove it and stream is close we could be potentially checking the last generated offset on something that doesn't exist anymore so I had to put the check there too. There is probably a better way to write this.

PR #52, opened 2026-01-16
Improved error classification in OAuth token retrieval

elenagaljak-db:

Improved error classification in OAuth token retrieval: 5xx server errors and network failures are now retryable, while 4xx client errors (invalid credentials, etc.) are non-retryable

## How is this tested?
not tested

### DIFF
--- a/sdk/src/errors.rs
+++ b/sdk/src/errors.rs
@@ -39,9 +39,9 @@ pub enum ZerobusError {
     /// Returned when a connection or setup operation times out.
     #[error("Connection timeout: {0}")]
     ConnectionTimeout(String),
-    /// Returned when a transient network or server error occurs.
-    #[error("Transient error: {0}")]
-    TransientError(String),
+    /// Returned when OAuth token fetching fails due to network or server errors.
+    #[error("Token fetch failed: {0}")]
+    TokenFetchError(String),
 }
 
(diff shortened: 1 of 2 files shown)


davidtosovic-db (approved):

sdk/src/errors.rs
```diff
@@ -79,6 +82,7 @@ impl ZerobusError {
             ZerobusError::UnexpectedStreamResponseError(_) => true,
             ZerobusError::InvalidStateError(_) => false,
             ZerobusError::ConnectionTimeout(_) => true,
+            ZerobusError::TransientError(_) => true,
```
teodordelibasic-db: Could we rename to i.e. `TokenFetchError` just to be more descriptive and precise?

teodordelibasic-db (commented):
Looks good, just a suggestion to rename the new error to be less generic. Also I think we should revisit the error handling as a whole and try to give actionable/descriptive errors and messages when possible. I feel like a lot of errors fall under the same enum variant and don't provide much info.

sdk/src/errors.rs
elenagaljak-db: That makes perfect sense. I agree with the logic of the error being thrown from only one place, it makes it easier to debug later. I will fix it now.

PR #55, opened 2026-01-19
Improve error propagation for faster feedback

elenagaljak-db:
Improve error propagation for faster feedback
- Add tokio::select! to detect server errors immediately in wait_for_offset_internal
- Pass server_error_tx to receiver and sender tasks for error broadcasting

## How is this tested?
Existing tests and added 1 more test.

### DIFF
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -53,7 +53,7 @@ use proto_zerobus::{
 };
 use smallvec::{smallvec, SmallVec};
 use tokio::sync::RwLock;
-use tokio::time::{sleep, Duration};
+use tokio::time::Duration;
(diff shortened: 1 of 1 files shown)


sdk/src/lib.rs
```diff
@@ -1129,6 +1143,9 @@ impl ZerobusStream {
                     )
                     .await;
                     return Err(error);
+                } else {
```
teodordelibasic-db: I am not sure if this is the correct place to clear the channel. Alternative is to clear it only after a successful recovery. With the current implementation when an error occurs, we clear it and start recovery, if recovery fails there is a period where there was no error, but the stream was never actually healthy. We can discuss offline further.

teodordelibasic-db (approved):

PR #56, opened 2026-01-20
Callback support

elenagaljak-db:

This PR adds acknowledgment callback support, enabling users to receive notifications when records/batches are acknowledged by the server or encounter errors. 
Created new sdk/src/callbacks.rs module with:                                                                                                    
    - AckCallback trait defining on_ack(offset_id) and on_error(offset_id, error_message) methods                                                    
    - Trait requires Send + Sync for safe use across async tasks 

Also renamed `arrow_config` to `arrow_configuration`

## How is this tested?

Added comprehensive test suite in tests/src/rust_tests.rs with 4 test cases:                                                                     
- test_callbacks_on_successful_acks - verifies callbacks fire for acknowledged records                                                           
- test_callbacks_on_stream_errors - verifies error callbacks fire when stream fails                                                              
- test_callbacks_with_batches - validates batch ingestion callback behavior                                                                      
- test_callback_called_per_offset - confirms callbacks fire once per offset in cumulative acks

### DIFF
--- a/README.md
+++ b/README.md
@@ -718,6 +718,8 @@ match stream.close().await {
 | `record_type` | `RecordType` | `RecordType::Proto` | Record serialization format (Proto or Json) |
 | `stream_paused_max_wait_time_ms` | `Option<u64>` | `None` | Max time to wait during graceful close (`None` = full server duration, `Some(0)` = immediate, `Some(x)` = min(x, server_duration)) |
 | `ack_callback` | `Option<Arc<dyn AckCallback>>` | `None` | Optional callback for acknowledgment notifications |
+| `callback_max_wait_time_ms` | `Option<u64>` | `None` | Maximum time to wait for callback processing to complete after closing the stream (`None` = wait indefinitely, `Some(x)` = wait up to `x` ms) |
+
 
 **Example:**
 
@@ -975,6 +977,7 @@ pub struct StreamConfigurationOptions {
     pub record_type: RecordType,
     pub stream_paused_max_wait_time_ms: Option<u64>,
     pub ack_callback: Option<Arc<dyn AckCallback>>,
+    pub callback_max_wait_time_ms: Option<u64>
 }
 ```
 
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -45,6 +45,7 @@
 
 ### API Changes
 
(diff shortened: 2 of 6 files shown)


teodordelibasic-db:
As discussed offline, let's introduce a separate task that handles callbacks in the Rust SDK and have the receiver task send messages through a channel to that task.

sdk/src/callbacks.rs
```diff
@@ -0,0 +1,118 @@
...
+//! ```
+
+use crate::offset_generator::OffsetId;
+
+/// Callback trait for receiving acknowledgment notifications.
+///
+/// Implement this trait to receive callbacks when records/batches are acknowledged
+/// by the server or when errors occur.
+///
+/// # Thread Safety and Performance
```
teodordelibasic-db: This isn't updated I think.

sdk/src/callbacks.rs
```diff
@@ -0,0 +1,118 @@
...
+///
+///     fn on_error(&self, offset_id: OffsetId, error_message: &str) {
+///         eprintln!("Error: {}", error_message);
+///     }
+/// }
+/// ```
+pub trait AckCallback: Send + Sync {
+    /// Called when a record/batch is successfully acknowledged by the server.
+    ///
+    /// **Warning**: This runs synchronously in the receiver task. Keep it lightweight
```
teodordelibasic-db: This isn't updated I think.

README.md
```diff
@@ -894,11 +974,48 @@ pub struct StreamConfigurationOptions {
...
+    fn on_error(&self, offset_id: OffsetId, error_message: &str);
+}
+```
+
+- `on_ack()` - Called when a record/batch is successfully acknowledged
+- `on_error()` - Called when a record/batch encounters an error
+
+**Important:** Callbacks run synchronously in the receiver task. Keep implementations lightweight to avoid blocking acknowledgment processing.
+
+### `HeadersProvider`
```
teodordelibasic-db: I initially didn't want to add too much information regarding this to avoid confusing clients since we support only one authentication methods and you can't really do too much with the headers provider, it's still mostly for our testing, and possibly some future-proofing.

elenagaljak-db: I agree but it is a public struct so it should exist in the API part of the readme. I think its reasonably small here

sdk/src/callbacks.rs
elenagaljak-db: fixed

PR #57, opened 2026-01-21
Optimize gRPC channel reuse across multiple streams

elenagaljak-db:

Modified `ZerobusSdk` to reuse a single gRPC channel across multiple stream creations instead of creating a new channel (and TCP connection) for each stream.
Also,  added `Drop` implementation for `ZerobusStream` to make sure supervisor tasks gets properly closed, along with receiver and sender task

## How is this tested?
Added 6 new tests to `stream_initialization_and_basic_lifecycle_tests` module:

- `test_close_waits_for_graceful_supervisor_shutdown` - Verifies graceful shutdown with timeout
- `test_drop_immediately_shuts_down` - Verifies drop uses immediate abort
- `test_close_idempotent_after_error` - Tests idempotency after errors
- `test_supervisor_responds_to_cancellation` - Tests cancellation token responsiveness
- `test_concurrent_close_operations` - Tests multiple tasks racing to close
- `test_stream_cleanup_on_drop` - Tests proper cleanup on drop

### DIFF
--- a/sdk/src/lib.rs
+++ b/sdk/src/lib.rs
@@ -62,6 +62,8 @@ use tonic::metadata::MetadataValue;
 use tonic::transport::{Channel, ClientTlsConfig, Endpoint};
 use tracing::{debug, error, info, instrument, span, warn, Level};
 
+const SHUTDOWN_TIMEOUT_SECS: u64 = 2;
+
 /// The type of the stream connection created with the server.
 /// Currently we only support ephemeral streams on the server side, so we support only that in the SDK as well.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -1020,6 +1022,11 @@ impl ZerobusStream {
         loop {
             debug!("Supervisor task loop");
 
+            if cancellation_token.is_cancelled() {
+                debug!("Supervisor task cancelled, exiting");
+                return Ok(());
+            }
+
             let landing_zone_sender = Arc::clone(&landing_zone);
             let landing_zone_receiver = Arc::clone(&landing_zone);
             let landing_zone_recovery = Arc::clone(&landing_zone);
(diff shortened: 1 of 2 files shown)


teodordelibasic-db (commented):
Looks good regarding the channel reuse, but I'd revert the task cancellation changes in this PR and leave it for one of the next PRs. I don't think this approach works fully and I think a fully correct solution would require a bit more changes so it is probably easier to keep this PR contained to channel reuse.

Why I think this approach is faulty is that we check the `is_closed` flag at the beginning of the sender and receiver tasks, so if i.e. the sender is blocked on `observe().await` waiting for new records it will never reach the next iteration to check the flag.

I think the best approach is using a cancellation token which is set during `close` and `drop`. Those seem to be the standard approach and we use it on server side as well.

sdk/src/lib.rs
```diff
@@ -998,15 +1012,13 @@ impl ZerobusStream {
...
         server_error_tx: tokio::sync::watch::Sender<Option<ZerobusError>>,
+        cancellation_token: CancellationToken,
     ) -> ZerobusResult<()> {
         let mut initial_stream_creation = true;
         let mut stream_init_result_tx = Some(stream_init_result_tx);
 
         loop {
             debug!("Supervisor task loop");
-            if is_closed.load(Ordering::Relaxed) {
-                return Ok(());
```
teodordelibasic-db: Now that we are removing this, I think we should also use the cancellation token in the supervisor task and check it at the beginning and the end, so here we can do:

```rust
if cancellation_token.is_cancelled() {                                            
    debug!("Supervisor task cancelled, exiting");                                 
    return Ok(());                                                                
}  
```

teodordelibasic-db (commented):
Looks good, I think there could be a few minor improvements potentially, but I think we should revisit this in any case.

PR #85, opened 2026-01-28
[Arrow] Refactor ack logic to handle record batch autochunking.

nikolaobradovic-db:

Add record-based acknowledgment tracking for Arrow Flight streams to support correct recovery when batches are auto-chunked. We are now tracking offsets of ingested records per stream, which allows us to map which portions of the Arrow record batches are made durable regardless of the autochunking method applied by ArrowFlight when the record batch is larger than the configured max GRPC message size.

## How is this tested?

Added new tests.

Closes https://github.com/databricks/zerobus-sdk-rs/issues/47

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -17,6 +17,7 @@
 
 ### Bug Fixes
 
+- **[Experimental] Record-based acknowledgment tracking for Arrow Flight streams**: Added cumulative record counting to support proper ack tracking and correct recovery when batches are auto-chunked.
 
 ### Documentation
 


tests/src/rust_tests.rs
```diff
@@ -369,10 +369,10 @@ mod stream_initialization_and_basic_lifecycle_tests {
...
 
         let test_record = b"test record data".to_vec();
-        let _ = stream.ingest_record(test_record).await?;
+        let _offset = stream.ingest_record(test_record).await?;
         let test_record = b"test record data 2".to_vec();
-        let _ = stream.ingest_record(test_record).await?;
-        let _ = stream.close().await;
+        let _offset = stream.ingest_record(test_record).await?;
+        let _result = stream.close().await;
         let flush_result = stream.flush().await;
```
elenagaljak-db: I'm going to guess these were a by-product of a different formatter?

elenagaljak-db (dismissed):
Looks good to me. I would maybe add short description of the changes in the NEXT_CHANGELOG.md but since this feature is still under development, maybe it's not needed?

teodordelibasic-db (dismissed):

PR #87, opened 2026-02-05
Add configurable TLS

teodordelibasic-db:

Previously TLS was exposed as a public field of the SDK. This PR adds an SDK builder API which also exposes an option to pass a struct implementing the `TlsConfig` trait which knows how to configure and return a gRPC endpoint.

Also fixes the header name for the identifying header from `user-agent` to `x-zerobus-sdk` because the `tonic` gRPC library overrides the standard header.

## How is this tested?

Unit tests, manual on local server.

### DIFF
--- a/NEXT_CHANGELOG.md
+++ b/NEXT_CHANGELOG.md
@@ -4,6 +4,15 @@
 
 ### New Features and Improvements
 
+- **Builder Pattern for SDK Initialization**: Added `ZerobusSdk::builder()` for fluent SDK configuration
+  - `.endpoint()` - Set the Zerobus endpoint (scheme is optional, defaults work with or without `https://`)
+  - `.unity_catalog_url()` - Set the Unity Catalog URL (optional when using custom headers providers)
+  - `.tls_config()` - Provide a custom `TlsConfig` implementation (defaults to `SecureTlsConfig`)
+- **Configurable TLS via `TlsConfig` trait**: TLS is now configured through a strategy pattern
+  - `SecureTlsConfig` (default) - Production TLS with system CA certificates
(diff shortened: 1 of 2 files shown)


sdk/src/headers_provider.rs
```diff
@@ -92,7 +89,7 @@ impl HeadersProvider for OAuthHeadersProvider {
         let mut headers = HashMap::new();
         headers.insert("authorization", format!("Bearer {}", token));
         headers.insert("x-databricks-zerobus-table-name", self.table_name.clone());
-        headers.insert("user-agent", self.user_agent.clone());
+        headers.insert("x-zerobus-sdk", DEFAULT_X_ZEROBUS_SDK.to_string());
```
elenagaljak-db: Just for my knowledge, why did we call it x-zerobus-sdk?

sdk/src/lib.rs
```diff
@@ -83,9 +57,40 @@ use tokio_retry::RetryIf;
...
+/// supported for production use. The API may change in future releases.
+#[cfg(feature = "arrow-flight")]
+pub use arrow_configuration::ArrowStreamConfigurationOptions;
+#[cfg(feature = "arrow-flight")]
+pub use arrow_stream::{
+    ArrowSchema, ArrowTableProperties, DataType, Field, RecordBatch, ZerobusArrowStream,
+};
+pub use builder::ZerobusSdkBuilder;
+pub use callbacks::AckCallback;
+pub use default_token_factory::DefaultTokenFactory;
```
elenagaljak-db: Any special reason why you rearranged the imports?

PR #141, opened 2026-03-12
[Java] Add experimental Arrow Flight support

danilonajkov-db:
Adds Arrow Flight ingestion to the Java SDK. The Java side accepts VectorSchemaRoot directly, serializes to IPC bytes internally, and passes them across the JNI boundary where the Rust SDK handles encoding, framing, and transmission over the Arrow Flight gRPC protocol.

  - New ZerobusArrowStream with ingestBatch(), waitForOffset(), flush(), close(), getUnackedBatches()
  - New ArrowStreamConfigurationOptions with builder pattern
  - New SDK methods: createArrowStream(), recreateArrowStream()
  - Opt-in: add arrow-vector and arrow-memory-netty as dependencies (provided scope, >= 15.0.0)
  - Integration tests and example in examples/arrow/

## How is this tested?
Added arrow flight integration tests

### DIFF
--- a/java/examples/README.md
+++ b/java/examples/README.md
@@ -9,6 +9,7 @@ The examples are organized by stream type and demonstrate both single-record and
 **Features demonstrated:**
 - `ZerobusProtoStream` - Protocol Buffer ingestion with method-level generics
 - `ZerobusJsonStream` - JSON ingestion with flexible serialization
+- `ZerobusArrowStream` (experimental) - Arrow Flight columnar ingestion
 - `ZerobusStream` (deprecated) - Legacy Future-based API for backward compatibility
 
 ## Directory Structure
@@ -25,6 +26,9 @@ examples/
 │   ├── README.md
 │   ├── SingleRecordExample.java
 │   └── BatchIngestionExample.java
+├── arrow/                 (Arrow Flight examples - ZerobusArrowStream, experimental)
+│   ├── README.md
+│   └── ArrowIngestionExample.java
 └── legacy/                (Legacy examples - ZerobusStream)
     └── LegacyStreamExample.java
 ```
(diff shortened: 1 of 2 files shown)


java/src/main/java/com/databricks/zerobus/ZerobusArrowStream.java
```diff
@@ -0,0 +1,326 @@
...
+  /**
+   * Closes the stream, flushing all pending batches first.
+   *
+   * <p>After closing, unacknowledged batches can still be retrieved via {@link
+   * #getUnackedBatches()} for use in stream recreation.
+   *
+   * @throws ZerobusException if an error occurs during close
+   */
+  @Override
+  public void close() throws ZerobusException {
```
teodordelibasic-db: We can investigate this further and fix in follow up PR since it exists for regular stream as well, but LLM found two issues in close:

1. Two threads calling `close()` concurrently can both read a non-zero handle, leading to double `nativeDestroy` (use-after-free). Same risk if `ingestBatch` races with `close` — `ensureOpen()` passes, then the handle gets freed underneath. I don't think this is of a high priority since we say streams are not supposed to be used by multiple threads.
2. If `nativeDestroy` throws, the native resource leaks forever. `nativeHandle` is set to 0, but underlying native memory allocated by the Rust SDK is never freed.

Something like this should fix maybe:

```rust
private final AtomicLong nativeHandle = new AtomicLong(0);

public void close() throws ZerobusException {
    long handle = nativeHandle.getAndSet(0);
    if (handle == 0) return;
    try {
        nativeClose(handle);
        try {
            cachedUnackedBatches = nativeGetUnackedBatches(handle);
        } catch (Exception e) {
            logger.warn("Failed to cache unacked batches: {}", e.getMessage());
            cachedUnackedBatches = new ArrayList<>();
        }
    } finally {
        nativeDestroy(handle);
    }
}
```

java/src/main/java/com/databricks/zerobus/ZerobusArrowStream.java
```diff
@@ -0,0 +1,326 @@
...
+  }
+
+  // ==================== Serialization Utilities ====================
+
+  /**
+   * Serializes an Arrow schema to IPC stream format.
+   *
+   * <p>Package-private for use by {@link ZerobusSdk#createArrowStream}.
+   */
+  static byte[] serializeSchemaToIpc(Schema schema) throws ZerobusException {
```
teodordelibasic-db: Improvement could be to reuse the `RootAllocator` since it's heavyweight. Since it's created only during stream creation the performance impact is negligible probably. We can either pass a `BufferAllocator` to this function from the caller or have a static one.

PR #155, opened 2026-03-23
[Rust] Add stream builder API

teodordelibasic-db:

This PR adds a `StreamBuilder` API to the Rust SDK that replaces the existing `create_stream*` family of methods with a fluent builder.

All setters can be called in any order. The builder validates at `build()` time that table name, authentication, and format have been configured.

```rust
let stream = sdk
    .stream_builder()
    .table("catalog.schema.table")
    .oauth("client-id", "client-secret")
    .json()
    .max_inflight_requests(500_000)
    .build()
    .await?;
```

Motivation:
- The current `create_stream` API is a wide parameter list where adding or removing arguments is a breaking change.
- The builder gives us a stable public surface where individual setters can be added/removed/deprecated independently.
- This is the first step toward providing similar builder APIs in the wrapper SDKs (Python, Java, Go, TypeScript), each idiomatic for their language.

Design decisions:
- No typestate, no generics — simple `StreamBuilder<'a>` with runtime validation at `build()`. Same pattern as `ZerobusSdkBuilder`. Keeps the public API surface small (one exported type) and allows setters in any order.
- `table_name` is a setter (`.table()`) not a constructor parameter, consistent with all other configuration.
- `build()` for gRPC streams (JSON/Proto), `build_arrow()` for Arrow streams — separate methods because the return types differ (`ZerobusStream` vs `ZerobusArrowStream`).
- `validate()` lets users check configuration without opening a stream, useful for fail-fast during startup.
- `#[must_use]` on `StreamBuilder` warns if a builder is created but never built.
- Deprecated methods are kept functional for backward compatibility and for wrapper SDKs that still depend on them.

## How is this tested?

1. 12 unit tests in `stream_builder.rs`.
2. 4 integration tests in `rust/tests/src/rust_tests.rs`.
3. CI.

## Follow-up items

These are scoped out of this PR and tracked for subsequent work.

1. Add `NoOpHeadersProvider` and `.no_auth()` builder method for local testing.

2. Add public getter methods on `ZerobusStream` and `ZerobusSdk` for fields that are currently public, prepares users for field privatization in version 2.0.0.

3. Unify `StreamConfigurationOptions` and `ArrowStreamConfigurationOptions` shared fields into a common config struct. Currently each shared setter dual-writes to both config structs.

4. Unify `ZerobusStream` and `ZerobusArrowStream` into a single stream type so `build()` can return one type regardless of format.

5. Add idiomatic stream builder APIs to wrapper SDKs (Python, Java, Go, TypeScript). Each language should use its own idiomatic pattern (kwargs in Python, traditional builder in Java, functional options in Go, options object in TypeScript).

6. Rust 2.0 breaking changes: privatize all public fields on `ZerobusStream`, `ZerobusSdk`, `StreamConfigurationOptions`, `ArrowStreamConfigurationOptions`, `TableProperties` (add getters). Remove deprecated `create_stream*` methods, `ZerobusSdk::new()`, `use_tls`, `ingest_record`/`ingest_records` (non-offset variants), and `record_type` from config.

### DIFF
--- a/rust/README.md
+++ b/rust/README.md
@@ -688,14 +688,18 @@ match stream.close().await {
 **Example:**
 
 ```rust
-let options = StreamConfigurationOptions {
-    max_inflight_requests: 50000,
-    recovery: true,
-    recovery_timeout_ms: 20000,
-    recovery_retries: 5,
-    flush_timeout_ms: 600000,
-    ..Default::default()
-};
+let stream = sdk
+    .stream_builder()
+    .table("catalog.schema.orders")
+    .oauth(client_id, client_secret)
+    .json()
+    .max_inflight_requests(50_000)
+    .recovery(true)
+    .recovery_timeout_ms(20_000)
+    .recovery_retries(5)
+    .flush_timeout_ms(600_000)
+    .build()
+    .await?;
 ```
 
 ## Error Handling
@@ -902,43 +906,9 @@ Returns unacknowledged records grouped by batch, preserving the original batch s
 
 Only call after stream failure.
 
-### `TableProperties`
+### `StreamBuilder`
 
-Configuration for the target table.
-
-**Fields:**
-```rust
-pub struct TableProperties {
-    pub table_name: String,
-    pub descriptor_proto: Option<prost_types::DescriptorProto>,
-}
-```
-
-- `table_name` - Full table name (e.g., "catalog.schema.table")
-- `descriptor_proto` - Optional Protocol buffer descriptor loaded from generated files (required for Proto record type, None for JSON)
-
-### `StreamConfigurationOptions`
-
-Stream behavior configuration.
-
-**Fields:**
-```rust
-pub struct StreamConfigurationOptions {
-    pub max_inflight_requests: usize,
-    pub recovery: bool,
-    pub recovery_timeout_ms: u64,
-    pub recovery_backoff_ms: u64,
-    pub recovery_retries: u32,
-    pub server_lack_of_ack_timeout_ms: u64,
-    pub flush_timeout_ms: u64,
-    pub record_type: RecordType,
-    pub stream_paused_max_wait_time_ms: Option<u64>,
-    pub ack_callback: Option<Arc<dyn AckCallback>>,
-    pub callback_max_wait_time_ms: Option<u64>
-}
-```
-
-See [Configuration Options](#configuration-options) for details.
+Configure stream parameters via fluent setters; all configuration goes through the builder. See [Create a Stream](#4-create-a-stream) for usage and [Configuration Options](#configuration-options) for the full list of available setters and their defaults.
 
 ### `AckCallback`
 
--- a/rust/README.md
+++ b/rust/README.md
@@ -42,7 +42,6 @@ The Zerobus Rust SDK provides a robust, async-first interface for ingesting larg
 - **High Throughput** - Configurable inflight record limits for optimal performance
 - **Batch Ingestion** - Ingest multiple records at once with all-or-nothing semantics for maximum throughput
 - **Flexible Serialization** - Support for both JSON (simple) and Protocol Buffers (type-safe) data formats
-- **Stream Builder** - Fluent builder API for configuring and creating ingestion streams
 - **Type Safety** - Protocol Buffers ensure schema validation at compile time
 - **Schema Generation** - CLI tool to generate protobuf schemas from Unity Catalog tables
 - **Flexible Configuration** - Fine-tune timeouts, retries, and recovery behavior
@@ -590,15 +589,15 @@ impl AckCallback for MyCallback {
 let mut stream = sdk
     .stream_builder().table("catalog.schema.orders")
     .oauth(client_id, client_secret)
-    .json()
+    .compiled_proto(descriptor_proto)
     .max_inflight_requests(10_000)
     .ack_callback(Arc::new(MyCallback))
     .build()
     .await?;
 
 for i in 0..1000 {
-    let record = serde_json::json!({"id": i, "name": format!("order-{}", i)});
-    stream.ingest_record_offset(record.to_string()).await?;
+    let record = YourMessage { id: Some(i), /* ... */ };
+    stream.ingest_record_offset(ProtoMessage(record)).await?;
     // Callback fires when this record is acknowledged
 }
 
--- a/rust/README.md
+++ b/rust/README.md
@@ -42,6 +42,7 @@ The Zerobus Rust SDK provides a robust, async-first interface for ingesting larg
 - **High Throughput** - Configurable inflight record limits for optimal performance
 - **Batch Ingestion** - Ingest multiple records at once with all-or-nothing semantics for maximum throughput
 - **Flexible Serialization** - Support for both JSON (simple) and Protocol Buffers (type-safe) data formats
+- **Stream Builder** - Fluent builder API for configuring and creating ingestion streams
 - **Type Safety** - Protocol Buffers ensure schema validation at compile time
 - **Schema Generation** - CLI tool to generate protobuf schemas from Unity Catalog tables
 - **Flexible Configuration** - Fine-tune timeouts, retries, and recovery behavior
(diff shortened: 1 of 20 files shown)


rust/sdk/src/lib.rs
```diff
@@ -267,8 +280,8 @@ pub struct ZerobusSdk {
     pub use_tls: bool,
     pub unity_catalog_url: String,
     shared_channel: tokio::sync::Mutex<Option<ZerobusClient<Channel>>>,
-    workspace_id: String,
-    tls_config: Arc<dyn TlsConfig>,
+    pub(crate) workspace_id: String,
+    pub(crate) tls_config: Arc<dyn TlsConfig>,
```
elenagaljak-db, line 265: nit: I saw claude put this pub(crate) before in my own code, I don't really know whats the purpose of it, but I think it's not necessary

rust/sdk/src/lib.rs
```diff
@@ -291,6 +304,43 @@ impl ZerobusSdk {
...
+    ///     .max_inflight_requests(500_000)
+    ///     .build()
+    ///     .await?;
+    /// ```
+    pub fn stream_builder(
+        &self,
+        table_name: impl Into<String>,
+    ) -> StreamBuilder<'_, NoFormat, NoAuth> {
+        StreamBuilder::new(self, table_name)
+    }
```
elenagaljak-db: Is there any specific reason why the table name is a method parameter and not another builder method `.table()` ?

rust/sdk/src/lib.rs
teodordelibasic-db: It's necessary. `pub(crate)` means "visible within this crate but not to external consumers." These fields were previously private (only accessible within `lib.rs`), but `StreamBuilder` lives in a different module (`builder/stream_builder.rs`) and needs to read `self.sdk.workspace_id`, `self.sdk.tls_config`, etc. to construct `OAuthHeadersProvider` and `ZerobusArrowStream`. Without `pub(crate)`, the builder can't access them. The same applies to `get_or_create_channel_zerobus_client()` and `ZerobusStream::new_stream()`.

rust/sdk/src/lib.rs
teodordelibasic-db: Eh not really except that I don't expect a stream to not need a table name in the future, but it probably may be better to have it as a method as well.

danilonajkov-db: yeah its a central part of ingestion. But still I would say its worth verifying the contract with Vicky

rust/sdk/src/builder/stream_builder.rs
```diff
@@ -0,0 +1,665 @@
...
+    /// Select JSON record format.
+    pub fn json(self) -> StreamBuilder<'a, Json, HasAuth> {
+        let mut b = self.transition();
+        b.grpc_config.record_type = RecordType::Json;
+        b.descriptor_proto = None;
+        b
+    }
+
+    /// Select compiled protobuf record format.
+    pub fn compiled_proto(
```
danilonajkov-db: maybe just `.proto()` (just personal preference, sounds more straightforward to me)

rust/sdk/src/builder/stream_builder.rs
```diff
@@ -0,0 +1,665 @@
+//! Typestate builder for creating Zerobus ingestion streams.
```
danilonajkov-db: Some validations from LLM that would be good to add:
```
  - No runtime assertion that descriptor_proto.is_some() when building
  CompiledProto. If a future change lets descriptor_proto be cleared independently
  of the typestate, this silently constructs a broken stream. Today it's safe
  because the only writers to descriptor_proto are .json() (sets None) and
  .compiled_proto(desc) (sets Some(desc)), and each pins the typestate accordingly — but that invariant is a code convention, not a compiler guarantee. Compare to
  the Arrow path, which does .expect("Arrow format guarantees schema is set")
  (stream_builder.rs:919) for the analogous invariant.

  - No check that self.auth.is_some() before resolve_headers_provider — relies on
  the HasAuth typestate. The .expect("HasAuth guarantees auth is set") at
  stream_builder.rs:852 is the runtime fallback if the type system is ever
  bypassed.
```

Also might be worth adding a method `is_valid_config` inside ZerobusStream::new_stream() that checks all invariants. To make sure the customer can't weirdly configure something while this conversion from old -> new stream creation API is happening.

danilonajkov-db (commented):
Should we mark the old methods as depricated?

rust/sdk/src/builder/stream_builder.rs
teodordelibasic-db: Idea here was to differentiate between `compiled protobuf` and `dynamic protobuf`. Dynamic Protobuf is essentially used to construct a Descriptor in-code, without having to have a `.proto` file and then manually compile it before running the actual Rust program. Taking this into consideration, I don't have strong feelings regarding the naming, if you think `proto` and `dynamic_proto`, if it ever gets added, makes more sense, I can change it.

rust/sdk/src/builder/stream_builder.rs
teodordelibasic-db: Hm so I think most of this comment is now outdated since I removed compile-time validations from the builder API. It just during `build` checks if all the fields are set and if there aren't any invalid combinations. Also `validate` is available to be called without calling `build`.

rust/sdk/src/lib.rs
```diff
@@ -814,31 +892,41 @@ impl ZerobusSdk {
...
 
-        Ok(new_stream)
+                Ok(new_stream)
+            }
+            Err(e) => {
+                error!("Arrow Flight stream recreation failed: {}", e);
+                Err(e)
+            }
+        }
     }
```
elenagaljak-db, line 907: Should we wait for these records to get acked before we return the new stream or no? I remember there was some discussion recently about it, maybe even a PR. Imo we don't have to, if ingestion fails it will show up in flush or close anyway, and it slows down the method. But I'm fine with both approaches

rust/README.md
```diff
@@ -42,6 +42,7 @@ The Zerobus Rust SDK provides a robust, async-first interface for ingesting larg
 - **High Throughput** - Configurable inflight record limits for optimal performance
 - **Batch Ingestion** - Ingest multiple records at once with all-or-nothing semantics for maximum throughput
 - **Flexible Serialization** - Support for both JSON (simple) and Protocol Buffers (type-safe) data formats
+- **Stream Builder** - Fluent builder API for configuring and creating ingestion streams
```
elenagaljak-db: Is this really a feature

rust/README.md
```diff
@@ -576,22 +587,18 @@ impl AckCallback for MyCallback {
...
+    .build()
+    .await?;
 
 for i in 0..1000 {
-    let record = YourMessage { id: Some(i), /* ... */ };
-    stream.ingest_record_offset(record.encode_to_vec()).await?;
+    let record = serde_json::json!({"id": i, "name": format!("order-{}", i)});
+    stream.ingest_record_offset(record.to_string()).await?;
     // Callback fires when this record is acknowledged
 }
```
elenagaljak-db, line 603: Any special reason this example is changed to json?

rust/README.md
```diff
@@ -753,18 +760,18 @@ let sdk = ZerobusSdk::builder()
...
+    .json()
+    .build()
+    .await?;
 
 // Ingest data...
 match stream.close().await {
     Err(_) => {
         // Stream failed, recreate with unacked records
-        stream = sdk.recreate_stream(stream).await?;
+        stream = sdk.recreate_stream(&stream).await?;
```
elenagaljak-db, line 777: Was this a mistake in the README?

elenagaljak-db:
Code looks much cleaner 😃

rust/sdk/src/lib.rs
teodordelibasic-db: The issue was that in the Scala SDK the recreate method was doing fire and forget for the records. Here the behavior is fine, we wait for the records to be enqueued, waiting for acks is on the client via calling `flush`.

rust/README.md
teodordelibasic-db: Yup.

rust/README.md
teodordelibasic-db: By mistake, reverted.

rust/README.md
teodordelibasic-db: Eh not really, removed.

elenagaljak-db (dismissed):

elenagaljak-db (approved):
Great work

PR #159, opened 2026-03-24
[Rust] Zero-copy Arrow IPC batch ingestion

elenagaljak-db:

Adds `ZerobusArrowStream::ingest_ipc_batch(Bytes)` so callers that already have Arrow IPC stream bytes (e.g. from Go/Python) can send them over Flight without decoding to RecordBatch and re-encoding

New API: `ingest_ipc_batch` accepts `bytes::Bytes` and builds `FlightData` from the IPC payload.
Pipeline: The Arrow Flight path now sends pre-encoded `FlightData` (schema + data messages) instead of feeding `RecordBatch` through `FlightDataEncoderBuilder`; ingest_batch still takes `RecordBatch` but encodes to `FlightData` before send.
IPC contract: Validates IPC with `StreamReader` - exactly one `RecordBatch` in the stream (trailing stream metadata after finish() is OK). .

## How is this tested?
New tests. Moved all arrow tests into a separate file.


rust/sdk/src/arrow_stream.rs
```diff
@@ -1029,98 +1294,70 @@ impl ZerobusArrowStream {
...
+    /// hold IPC-serialised bytes. Forwards bytes directly to the Flight wire format
+    /// without deserialising to a [`RecordBatch`] — eliminating one IPC round-trip
+    /// compared to calling `ingest_batch`.
+    ///
+    /// The `ipc_bytes` must be a valid Arrow IPC *stream* containing exactly one
+    /// RecordBatch (i.e. the output of `pyarrow.RecordBatch.serialize()`,
+    /// `tableToIPC(table, 'stream')`, etc.). Trailing stream metadata (such as an
+    /// end-of-stream marker after `finish()`) is allowed after that batch.
+    #[instrument(level = "debug", skip_all, fields(table_name = %self.table_properties.table_name))]
+    pub async fn ingest_ipc_batch(&self, ipc_bytes: Bytes) -> ZerobusResult<OffsetId> {
```
danilonajkov-db: in ingest_batch we validate the schema with:
```

        // Validate schema matches.
        if batch.schema() != self.table_properties.schema {
            return Err(ZerobusError::InvalidArgument(format!(
                "RecordBatch schema does not match stream schema. Expected: {:?}, Got: {:?}",
                self.table_properties.schema,
                batch.schema()
            )));
        }
```
So now as we can't do it here, we need every downstream SDK to implement this in its own way right?

danilonajkov-db (commented):
cool!

teodordelibasic-db (commented):
**Nit (not in diff):** Module doc at line 6 mentions only `RecordBatch` ingestion. Should mention `ingest_ipc_batch` as an alternative entry point for FFI callers.

rust/sdk/src/arrow_stream.rs
```diff
@@ -69,44 +91,227 @@ struct PendingBatch {
...
+        let meta_end = meta_start + meta_len as usize;
+        if meta_end > bytes.len() {
+            return Err(ZerobusError::InvalidArgument(
+                "IPC: truncated metadata".into(),
+            ));
+        }
+        Ok((meta_start, meta_end))
+    }
+
+    let (ms, me) = read_meta_range(bytes, 0)?;
```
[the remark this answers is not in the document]
elenagaljak-db: Fixed

rust/sdk/src/arrow_stream.rs
```diff
@@ -69,44 +94,281 @@ struct PendingBatch {
...
+                    records_already_acked as usize,
+                    remaining_rows as usize,
+                ))))
+            }
+        }
+    }
+}
+
+/// Deserialises Arrow IPC stream bytes into a [`RecordBatch`].
+/// Enforces the same single-batch contract as [`validate_ipc_stream_exactly_one_record_batch`].
```
teodordelibasic-db: nit: `validate_ipc_stream_exactly_one_record_batch` doesn't exist anymore, should it be `ipc_bytes_to_flight_data`?

teodordelibasic-db (dismissed):
Looks great!

teodordelibasic-db (approved):

PR #187, opened 2026-03-30
Add .NET Sdk

sp-jcberleur:

Initial Databricks Zerobus Ingest SDK for .NET to ingest data from .NET applications

## How is this tested?

```
dotnet test
```


dotnet/src/Zerobus/Native/HeadersProviderBridge.cs
```diff
@@ -0,0 +1,71 @@
...
+            {
+                result.Headers = IntPtr.Zero;
+                result.Count = 0;
+                result.ErrorMessage = IntPtr.Zero;
+                return result;
+            }
+
+            // Allocate an array of CHeader structs in unmanaged memory.
+            var headerSize = Marshal.SizeOf<CHeader>();
+            var arrayPtr = Marshal.AllocHGlobal(headerSize * headers.Count);
```
teodordelibasic-db: I think there is an allocator mismatch with the Rust free path. Rust's `zerobus_free_headers` (`rust/ffi/src/lib.rs:798`) frees this pointer with `libc::free()`, which expects memory from the C allocator's `malloc`.

- On Linux/macOS, `AllocHGlobal` uses `malloc`, so this works by accident.
- On Windows, `AllocHGlobal` uses `LocalAlloc`, which is a different heap than MSVCRT's `free`. This is undefined behavior — heap corruption, sporadic crashes.

The Go SDK allocates with `C.malloc` (`go/ffi.go:368`), which matches the Rust free path.

We can replace with `NativeMemory.Alloc((nuint)(headerSize * headers.Count))`, which calls `malloc` on all platforms.

dotnet/src/Zerobus/Native/HeadersProviderBridge.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            int idx = 0;
+            foreach (var (key, value) in headers)
+            {
+                var headerPtr = arrayPtr + idx * headerSize;
+                var cHeader = new CHeader
+                {
+                    Key = Marshal.StringToCoTaskMemUTF8(key),
```
teodordelibasic-db: This might be a same allocator mismatch. Rust frees these with `CString::from_raw()` (`lib.rs:792-795`), which calls Rust's global dealloc = `libc::free` on the default allocator. `StringToCoTaskMemUTF8` uses `CoTaskMemAlloc`, which on Windows is the COM allocator — not `malloc`.

Go uses `C.CString()` (`go/ffi.go:379-380`), which calls `malloc`.

We can encode strings manually with `Encoding.UTF8.GetBytes`, allocate with `NativeMemory.Alloc`, copy bytes in and add the NUL terminator. Something like:

```csharp
private static IntPtr AllocUtf8String(string s)
{
  var byteCount = Encoding.UTF8.GetByteCount(s);
  var ptr = (byte*)NativeMemory.Alloc((nuint)(byteCount + 1));
  Encoding.UTF8.GetBytes(s, new Span<byte>(ptr, byteCount));
  ptr[byteCount] = 0;
  return (IntPtr)ptr;
}
```

dotnet/src/Zerobus/Native/HeadersProviderBridge.cs
```diff
@@ -0,0 +1,71 @@
...
+
+            result.Headers = arrayPtr;
+            result.Count = (nuint)headers.Count;
+            result.ErrorMessage = IntPtr.Zero;
+        }
+        catch (Exception ex)
+        {
+            result.Headers = IntPtr.Zero;
+            result.Count = 0;
+            result.ErrorMessage = Marshal.StringToCoTaskMemUTF8(ex.Message);
```
teodordelibasic-db: Same issue potentially, Rust frees `error_message` with `CString::from_raw()` (`lib.rs:803`). We could use the C allocator by applying the same `NativeMemory.Alloc` + manual UTF-8 encoding fix as above.

dotnet/src/Zerobus/ZerobusStream.cs
```diff
@@ -0,0 +1,308 @@
...
+            ObjectDisposedException.ThrowIf(_disposed != 0, this);
+            return _ptr;
+        }
+    }
+
+    /// <summary>
+    /// Attempts to retrieve the bridge handle and callback reference for the stream.
+    /// Internal use only.
+    /// </summary>
+    internal ZerobusStream Recreate(IntPtr newPtr)
```
teodordelibasic-db, line 426: The old native stream pointer (`_ptr`) is leaked here I think. After `Recreate()`:

- `_disposed` is set to `1`, so `Dispose(bool)` will early-return at line 244
- `_ptr` still holds the old pointer, but `StreamFree` is never called for it
- The finalizer also no-ops because `_disposed != 0`

Rust's `zerobus_sdk_recreate_stream` takes the old stream by reference - it reads from it but does not free it. The caller should eventually call `zerobus_stream_free`.

Fix might look something like this:

```csharp
internal ZerobusStream Recreate(IntPtr newPtr)
{
    var disposed = Interlocked.CompareExchange(ref _disposed, 1, 0);
    ObjectDisposedException.ThrowIf(disposed != 0, this);

    var oldPtr = Interlocked.Exchange(ref _ptr, IntPtr.Zero);

    var newStream = _bridgeHandle.IsAllocated
        ? new ZerobusStream(newPtr, _bridgeHandle, _callbackRef!)
        : new ZerobusStream(newPtr);

    // Prevent accidental double-free of the shared handle from the old wrapper
    _bridgeHandle = default;

    // Free the old native stream (don't Close — it's already failed/closed)
    if (oldPtr != IntPtr.Zero)
        NativeMethods.StreamFree(oldPtr);

    return newStream;
}

File: dotnet/src/Zerobus/Native/NativeInterop.cs [hunk 5]
```diff
@@ -0,0 +1,422 @@
...
+    {
+        if (records.Length == 0)
+            return -1;
+
+        var result = new CResult();
+        var numRecords = (nuint)records.Length;
+
+        // Pin all record buffers and collect pointers
+        var handles = new GCHandle[records.Length];
+        var ptrs = stackalloc byte*[records.Length];
```
teodordelibasic-db: `stackalloc` with user-controlled size can overflow the stack. On 64-bit, this allocates `records.Length * 16` bytes. For 100k records that's 1.6MB — exceeding the default 1MB .NET thread stack. A `StackOverflowException` cannot be caught.

Go doesn't have this issue because CGO uses separate stack management.

We can add a threshold and fall back to heap allocation for large batches:
```csharp
const int StackAllocThreshold = 4096;
byte** ptrs = records.Length <= StackAllocThreshold
    ? stackalloc byte*[records.Length]
    : (byte**)NativeMemory.Alloc((nuint)(records.Length * sizeof(byte*)));
```
(And free the heap allocation in the `finally` block.)

dotnet/src/Zerobus/Native/NativeInterop.cs
```diff
@@ -0,0 +1,422 @@
...
+        if (records.Length == 0)
+            return -1;
+
+        var result = new CResult();
+        var numRecords = (nuint)records.Length;
+
+        // Encode each string as null-terminated UTF-8 and pin
+        var encoded = new byte[records.Length][];
+        var handles = new GCHandle[records.Length];
+        var ptrs = stackalloc byte*[records.Length];
```
teodordelibasic-db: Same `stackalloc` overflow risk as the proto batch path above.

dotnet/src/Zerobus/Native/NativeInterop.cs
```diff
@@ -0,0 +1,422 @@
...
+
+        try
+        {
+            for (int i = 0; i < records.Length; i++)
+            {
+                // Encode with null terminator
+                var utf8 = Encoding.UTF8.GetBytes(records[i] + '\0');
```
teodordelibasic-db: More of a nit. The `records[i] + '\0'` allocates a new string per record just to append a NUL character. 

Minor optimization - encode without concatenation and append the NUL byte directly:

```csharp
var byteCount = Encoding.UTF8.GetByteCount(records[i]);
var utf8 = new byte[byteCount + 1];
Encoding.UTF8.GetBytes(records[i], utf8);
// utf8[byteCount] is already 0 from array initialization
```

dotnet/src/Zerobus/Native/NativeInterop.cs
```diff
@@ -0,0 +1,422 @@
...
+    public static CStreamConfigurationOptions ConvertConfig(StreamConfigurationOptions? options)
+    {
+        if (options is null)
+            return NativeMethods.GetDefaultConfig();
+
+        var defaults = StreamConfigurationOptions.Default;
+
+        return new CStreamConfigurationOptions
+        {
+            MaxInflightRequests = (nuint)(options.MaxInflightRequests > 0
```
teodordelibasic-db: Setting `RecoveryRetries = 0` (meaning "don't retry") silently falls back to the default of 4. The Go SDK has the same bug (comment at `go/ffi.go:193`). We can add a code comment acknowledging this or use nullable types (`uint?`) to distinguish "unset" from "explicitly zero".

dotnet/src/Zerobus/ZerobusSdk.cs
```diff
@@ -0,0 +1,210 @@
...
+    {
+        var ptr = Interlocked.Exchange(ref _ptr, IntPtr.Zero);
+        if (ptr != IntPtr.Zero)
+        {
+            NativeMethods.SdkFree(ptr);
+        }
+    }
+
+    /// <inheritdoc />
+    public void Dispose()
```
teodordelibasic-db, line 632: `_disposed` is a plain `bool` with non-atomic reads/writes. If `Dispose()` is called from two threads simultaneously, both could read `_disposed == false` before either writes `true`. We can consider aligning to the same `Interlocked.CompareExchange` pattern used in `ZerobusStream.Dispose(bool)`.

teodordelibasic-db (commented):
Thanks a lot for contributing and sorry for the delayed review, we've been quite overwhelmed with product stabilization these past few weeks. I'll prioritize now so that we can get it in, a few other customers also want to use it. I am nowhere near a C#/.NET expert so I am reviewing mostly with LLM help.

A few more general comments below as well:

---

We should integrate the `.NET` CI into general repo CI. Currently PRs touching only `dotnet/**` pass CI without any checks and Rust FFI changes don't trigger cross-SDK .NET tests. In `.github/workflows/push.yml` we should add:

1. `dotnet` output in the `changes` job.
2. `dotnet` path filter (`dotnet/**` + `.github/workflows/ci-dotnet.yml`)
3. `dotnet` job calling `ci-dotnet.yml`
4. `dotnet` in the `gate` job's `needs` array + result check loop
5. `cross-sdk-dotnet` job (mirroring `cross-sdk-go`) triggered when `rust/**` changes

---

A couple of missing misc files in the `dotnet/` directory that we have in other SDKs:

1. `CONTRIBUTING.md` - guidelines how to setup a devloop for making changes and contributing them for this SDK.
2. `NEXT_CHANGELOG.md` - used by the release process for unreleased changes
3. `NOTICE` - license/attribution file

---

As far as I understand, this is a sync-only API, LLM suggests we could benefit from an API returning `Task` object for methods `CreateStream`, `WaitForOffset`, `Flush`, and `Close`. ASP.NET Core users calling current API on request threads will block thread pool threads.

Not a blocker obviously, just want to cross-check with you if it makes sense to add it in the future.

RichardCupples:
Any progress, PR has been parked for 2 months now?

RichardCupples:
Hi, I understood this PR was to be prioritized, could it please be merged soon?

at-besa:
Also waiting for this some time now 👍🏽

teodordelibasic-db:
[`rust/ffi/src/lib.rs:1092`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/rust/ffi/src/lib.rs#L1092)

The ingest and flush paths keep the retryable flag by building the result with `CResult::error(err)`, which reads `err.is_retryable()` ([`lib.rs:692`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/rust/ffi/src/lib.rs#L692)). But `create_stream` ([`lib.rs:1092`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/rust/ffi/src/lib.rs#L1092)), `create_stream_with_headers_provider` ([`lib.rs:1160`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/rust/ffi/src/lib.rs#L1160)) and `recreate_stream` ([`lib.rs:1206`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/rust/ffi/src/lib.rs#L1206)) turn the error into a plain `String` first and then call `write_error_result(result, &err, false)`, which hardcodes the flag to false.

So a .NET caller always sees `ZerobusException.IsRetryable` as false for a failed create or recreate, even when Rust considered the error retryable. Create and recreate are exactly where a caller decides whether to retry, so this looks like a real bug rather than just cosmetic.

Could we keep the typed error all the way to the result and use `CResult::error(err)` for create, create with headers provider, and recreate, the same way ingest does? A test that injects a retryable failure on create and checks that `IsRetryable` is true would cover it.

teodordelibasic-db:
[`rust/ffi/src/lib.rs:573`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/rust/ffi/src/lib.rs#L573)

Only `zerobus_arrow_stream_is_closed` exists ([`lib.rs:573`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/rust/ffi/src/lib.rs#L573)). There is no `is_closed` for the proto or JSON stream, and the .NET wrapper does not expose one. The Rust stream and the Go SDK both have `is_closed()`, and it is the natural way to decide when to call `RecreateStream`. Without it, .NET callers can only guess from caught exceptions. We could add it (a small FFI addition).

teodordelibasic-db:
[`rust/ffi/src/lib.rs:833`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/rust/ffi/src/lib.rs#L833)

`CallbackHeadersProvider::get_headers` sets an `in_use` flag and returns the error `"Concurrent headers provider callback detected - Go callback must be thread-safe"` if it sees a second call at the same time ([`lib.rs:835`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/rust/ffi/src/lib.rs#L835)). This is existing shared FFI code, but the recreate path added in this PR makes it easier to hit: `recreate_stream` calls `Arc::clone(&stream.headers_provider)` ([`lib.rs:578`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/rust/sdk/src/lib.rs#L578)), so the old and new streams share one provider during recovery, and if both call it at once a user could get this error through no fault of their own. Two small suggestions: document on `IHeadersProvider` that implementations must be thread-safe, and change the user visible error text so it does not mention Go.

teodordelibasic-db:
[`rust/ffi/src/lib.rs:1226`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/rust/ffi/src/lib.rs#L1226)

None of the FFI entry points wrap their body in `std::panic::catch_unwind` (zero occurrences in the file). A Rust panic that unwinds across the `extern "C"` boundary is undefined behavior and in practice takes down the whole .NET process. This is existing shared FFI code and also affects Go, so it may be out of scope for this PR, but since .NET now calls into it, it would be worth a guard on the entry points (or a tracking issue) so a panic becomes an error result instead of a crash.

dotnet/src/Zerobus/ZerobusStream.cs
```diff
@@ -0,0 +1,318 @@
...
+    /// <code>
+    /// // JSON
+    /// long offset = stream.IngestRecord("{\"id\": 1, \"message\": \"Hello\"}");
+    ///
+    /// // Protobuf
+    /// byte[] protoBytes = SerializeMyProto(myMessage);
+    /// long offset = stream.IngestRecord(protoBytes);
+    /// </code>
+    /// </example>
+    public long IngestRecord(string payload)
```
teodordelibasic-db, line 92: Every operation here checks `_disposed` and then reads the raw `_ptr` and passes it to native code, with nothing stopping `Dispose` from running in between. [`ZerobusStream.cs:242`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/dotnet/src/Zerobus/ZerobusStream.cs#L242) swaps `_ptr` to zero and then calls `StreamClose` and `StreamFree` on the old pointer.

So this sequence looks possible: thread A passes the disposed check and reads `_ptr`, thread B disposes and frees the native stream, thread A then calls into Rust with a pointer that has already been freed. That would be a use-after-free.

It also looks unsafe on the Rust side. `zerobus_stream_close` takes the pointer as `&mut ZerobusStream` ([`lib.rs:1613`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/rust/ffi/src/lib.rs#L1613)) while `ingest`, `flush` and `wait` take it as `&ZerobusStream`. If `close` runs at the same time as `ingest`, Rust ends up with both a `&mut` and a `&` to the same object, which is undefined behavior on its own.

This matters because the docs say the stream is thread-safe and can be used from many threads ([`README.md:98`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/dotnet/README.md#L98), [`ZerobusStream.cs:12`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/dotnet/src/Zerobus/ZerobusStream.cs#L12)), and disposing during shutdown or error handling is common. Could we guard the native handle so it cannot be freed while a call is using it? For example a ref counted `SafeHandle`, or a reader/writer lock where `ingest` takes the read side and `Dispose` takes the write side and waits for in-flight calls to finish. And we should avoid calling the Rust `close` (`&mut`) while any other call can still be running.

dotnet/src/Zerobus/Native/HeadersProviderBridge.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                result.Headers = IntPtr.Zero;
+                result.Count = 0;
+                result.ErrorMessage = IntPtr.Zero;
+                return result;
+            }
+
+            // Allocate an array of CHeader structs in unmanaged memory.
+            var headerSize = Marshal.SizeOf<CHeader>();
+            var arrayPtr = (nint)NativeMemory.Alloc((nuint)(headerSize * headers.Count));
```
teodordelibasic-db: The callback allocates the header array and the key, value and error strings with `NativeMemory.Alloc`, which uses the C runtime `malloc`. Rust then frees them in [`zerobus_free_headers`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/rust/ffi/src/lib.rs#L786): the array with `libc::free`, and each string by turning it back into a `CString` and dropping it.

Freeing memory that C# allocated with `malloc` by dropping a Rust `CString` only works when both use the same heap. On Linux and macOS they happen to line up, so it works today. On Windows the C runtime heap and Rust's allocator are not guaranteed to be the same, so this can corrupt the heap. Windows x64 is a shipped target ([`Zerobus.csproj:104`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/dotnet/src/Zerobus/Zerobus.csproj#L104)).

The Go SDK gets away with this because its strings come from cgo's `C.CString`, which uses the same `libc` that Rust's `CString` drop expects. This port allocates through `NativeMemory` and frees through `CString`, so that no longer matches. Could we pick one owner for this memory? Either allocate the strings on the Rust side so Rust also frees them, or free everything the callback returns with `libc::free` instead of `CString`. It would also be good to test this on Windows under a heap checker with `CreateStreamWithHeadersProvider`.

dotnet/src/Zerobus/ZerobusStream.cs
```diff
@@ -0,0 +1,318 @@
...
+    /// The stream cannot be used after calling this method.
+    /// </summary>
+    /// <remarks>
+    /// This is automatically called by <see cref="Dispose()"/>, but you may call
+    /// it explicitly if you need to inspect the close error.
+    /// </remarks>
+    /// <exception cref="ZerobusException">
+    /// Thrown if flush or close fails.
+    /// </exception>
+    public void Close()
```
teodordelibasic-db, line 288: `Close()` calls `Dispose(true)`, which marks the wrapper disposed and always frees the native stream in the finally block ([`ZerobusStream.cs:258`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/dotnet/src/Zerobus/ZerobusStream.cs#L258)). After that, `GetUnackedRecords()` and `NativePointer` both throw `ObjectDisposedException`, so we cannot call `GetUnackedRecords` or `sdk.RecreateStream(stream)` on a stream we closed.

This seems to conflict with the docs at [`ZerobusStream.cs:16`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/dotnet/src/Zerobus/ZerobusStream.cs#L16), which tell users to check for unacknowledged records after a failure, and with recovery in general: the Rust `recreate_stream(&stream)` ([`lib.rs:578`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/rust/sdk/src/lib.rs#L578)) needs the old stream to still be alive so it can read its unacked batches. Right now there is no public way to close or flush gracefully and still keep the stream readable and recreatable.

We can see this in the test at [`IntegrationTests.cs:460`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/dotnet/tests/Zerobus.IntegrationTests/IntegrationTests.cs#L460): it calls `NativeInterop.StreamClose(stream.NativePointer)` directly instead of `stream.Close()`, because the public `Close()` would free the pointer and make `RecreateStream` impossible. So the recovery path is only tested in a way real users cannot reproduce.

Could we split graceful close from freeing the stream? A closed or failed stream would stay readable (`GetUnackedRecords`) and recreatable until the caller explicitly disposes it. A test for close or flush failure followed by `GetUnackedRecords` and `RecreateStream` would lock this down.

dotnet/README.md
```diff
@@ -0,0 +1,346 @@
...
+```csharp
+using var stream = sdk.CreateStreamWithHeadersProvider(
+    new TableProperties("catalog.schema.table"),
+    new MyHeadersProvider(),
+    options);  // optional
+```
+
+### `ZerobusStream`
+
+An active bidirectional gRPC stream for record ingestion. Thread-safe.
```
teodordelibasic-db: [`README.md:98`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/dotnet/README.md#L98) says the stream is thread-safe, and the XML docs ([`ZerobusStream.cs:12`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/dotnet/src/Zerobus/ZerobusStream.cs#L12)) say you can call `IngestRecord` from many threads at once. There is no test that actually ingests from several threads on one stream, and because `Dispose` can race with an in-flight call, the claim is not currently safe.

Whatever the final design is, it would be good to make the docs match it. For example, we could say that concurrent ingest is safe but ingest at the same time as `Dispose` or `Close` is not, or make it fully safe and keep the promise. It would be better not to ship the claim as it stands.

dotnet/src/Zerobus/ZerobusStream.cs
```diff
@@ -0,0 +1,318 @@
...
+    /// <exception cref="ZerobusException">Thrown if the wait fails.</exception>
+    /// <exception cref="ObjectDisposedException">Thrown if the stream has been disposed.</exception>
+    /// <example>
+    /// <code>
+    /// long offset = stream.IngestRecord(data);
+    /// // ... do other work ...
+    /// stream.WaitForOffset(offset);
+    /// </code>
+    /// </example>
+    public void WaitForOffset(long offset)
```
teodordelibasic-db, line 197: Every operation is a blocking call over an async Rust operation running on a shared Tokio runtime (`RUNTIME.block_on`). `WaitForOffset`, `Flush`, `Close`, stream creation, and ingest under backpressure all hold the calling .NET thread for the whole network wait, and there is no `CancellationToken`. In ASP.NET Core or worker services this ties up thread pool threads and can starve them under load.

Matching the Go SDK's blocking style is reasonable for a first version, but it would be good to make it a clear choice rather than a surprise. At a minimum, we could say in the `README` that these are blocking calls and should not run on request handling threads. As a follow up, we could add async versions (`CreateStreamAsync`, `IngestRecordAsync` returning `ValueTask<long>`, `WaitForOffsetAsync`, `FlushAsync`, `CloseAsync`), `IAsyncDisposable`, and `CancellationToken` support. Real async needs the FFI to signal completion through a callback; wrapping in `Task.Run` only moves the blocking, it does not remove it. If that is out of scope for now, it would help to say so.

dotnet/src/Zerobus/ZerobusSdk.cs
```diff
@@ -0,0 +1,208 @@
...
+    /// (e.g. <c>https://zerobus.databricks.com</c>).
+    /// </param>
+    /// <param name="unityCatalogUrl">
+    /// The Unity Catalog URL for OAuth token acquisition
+    /// (e.g. <c>https://workspace.databricks.com</c>).
+    /// </param>
+    /// <exception cref="ZerobusException">
+    /// Thrown if the SDK cannot be initialised (invalid URLs, etc.).
+    /// </exception>
+    public ZerobusSdk(string zerobusEndpoint, string unityCatalogUrl)
```
teodordelibasic-db: The way we construct the SDK is worth settling now, while the NuGet API is still unpublished. For streams the current design is good: the `StreamConfigurationOptions` record with `Default` and the `with` expression is idiomatic C#, so no stream builder is needed there. The concern is only the SDK constructor.

The two argument constructor `ZerobusSdk(endpoint, unityCatalogUrl)` maps to `zerobus_sdk_new`, which decides TLS purely from the `http://` prefix and gives no place to pass custom TLS, an application name or user agent, a proxy, or future options. The Rust SDK deprecated its own two argument `new` in favor of `ZerobusSdk::builder()` for this reason. Adding options later would mean either many constructor overloads or a breaking change.

Also, `unityCatalogUrl` is required even for `CreateStreamWithHeadersProvider`, but with a custom headers provider there is no Unity Catalog token to fetch, and the Rust builder makes it optional there.

Suggestion: add an extension point now, either an options object passed to the constructor or a small builder such as `ZerobusSdk.CreateBuilder().Endpoint(...).UnityCatalogUrl(...).Build()`, with `UnityCatalogUrl` optional. Even if TLS cannot be passed through the FFI yet, having it there would avoid a breaking change later. A full stream builder is a follow up, not a blocker.

dotnet/tests/Zerobus.IntegrationTests/IntegrationTests.cs
```diff
@@ -0,0 +1,507 @@
+using System.Diagnostics;
+using Grpc.Core;
+using NUnit.Framework;
+
+namespace Databricks.Zerobus.IntegrationTests;
+
+[TestFixture]
+[Parallelizable(ParallelScope.Children)]
```
teodordelibasic-db: The tests run sequentially (the `Parallelizable` attribute only parallelizes independent fixtures). It would be good to add before merge:

- Many threads ingesting on one stream while another thread disposes, closes, or recreates it, to exercise the use-after-free path.
- A close or flush failure, then `GetUnackedRecords` still returns the records, then `RecreateStream` still works.
- Assertions that `IsRetryable` is true for a retryable create or recreate failure.
- A headers provider test that runs the real C# callback allocation path on Windows, Linux and macOS.

The current unit tests (`ZerobusSdkTests`, `NativeInteropTests`) are fairly thin, mostly null argument checks.

dotnet/src/Zerobus/Native/NativeInterop.cs
```diff
@@ -0,0 +1,439 @@
...
+                ? options.MaxInflightRequests
+                : defaults.MaxInflightRequests),
+            Recovery = options.Recovery,
+            RecoveryTimeoutMs = options.RecoveryTimeoutMs > 0
+                ? options.RecoveryTimeoutMs
+                : defaults.RecoveryTimeoutMs,
+            RecoveryBackoffMs = options.RecoveryBackoffMs > 0
+                ? options.RecoveryBackoffMs
+                : defaults.RecoveryBackoffMs,
+            RecoveryRetries = options.RecoveryRetries > 0
```
teodordelibasic-db: `ConvertConfig` uses `> 0` checks: `RecoveryRetries = options.RecoveryRetries > 0 ? options.RecoveryRetries : defaults.RecoveryRetries` ([`NativeInterop.cs:421`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/dotnet/src/Zerobus/Native/NativeInterop.cs#L421)), and the same for `MaxInflightRequests` ([`NativeInterop.cs:411`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/dotnet/src/Zerobus/Native/NativeInterop.cs#L411)) and the timeout fields. So a caller who explicitly sets `RecoveryRetries` to `0`, meaning no retries, silently gets `4` instead.

`StreamPausedMaxWaitTimeMs` already handles this correctly with a nullable `ulong` and a `Has` flag. Could we do the same for any field where `0` is a real value, or validate and throw? At the very least, it would help to document that `0` means use the default.

dotnet/src/Zerobus/ZerobusSdk.cs
```diff
@@ -0,0 +1,208 @@
...
+    /// <summary>
+    /// Recreates a new bidirectional gRPC stream from an existing stream.
+    /// This is used for recovery scenarios where a stream needs to be re-established
+    /// using the existing stream's configuration and state.
+    /// </summary>
+    /// <param name="stream">The existing stream to recreate from.</param>
+    /// <returns>A new <see cref="ZerobusStream"/> ready for record ingestion.</returns>
+    /// <exception cref="ZerobusException">Thrown if the stream cannot be recreated.</exception>
+    /// <exception cref="ObjectDisposedException">Thrown if the SDK has been disposed.</exception>
+    public ZerobusStream RecreateStream(ZerobusStream stream)
```
teodordelibasic-db, line 507: `RecreateStream(stream)` returns a new stream and also marks the `stream` passed in as disposed, freeing its native pointer (`stream.Recreate`, [`ZerobusStream.cs:299`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/dotnet/src/Zerobus/ZerobusStream.cs#L299)). Passing an object in and having it become unusable is a little surprising in C#, and it does not work well with `using var stream = ...`. We could either return the new stream without changing the argument, or make the behavior clear in the method's XML docs and the `README` so callers know not to use the old handle afterward.

dotnet/src/Zerobus/ZerobusStream.cs
```diff
@@ -0,0 +1,318 @@
...
+    ///     stream.Flush();
+    /// }
+    /// catch (ZerobusException)
+    /// {
+    ///     var unacked = stream.GetUnackedRecords();
+    ///     Console.WriteLine($"Failed to acknowledge {unacked.Length} records");
+    /// }
+    /// </code>
+    /// </example>
+    public object[] GetUnackedRecords()
```
teodordelibasic-db: Returning `object[]`, where each element is either a `string` or a `byte[]`, forces boxing and type checks at the call site and is hard to change once the API is public. We could use a typed shape instead, for example a `readonly record struct UnackedRecord(byte[] Data, bool IsJson)` with an `AsJson()` helper, or return `ReadOnlyMemory<byte>[]` and let the caller decode. Worth doing before this is public.

dotnet/src/Zerobus/Native/NativeInterop.cs
```diff
@@ -0,0 +1,439 @@
...
+
+        if (!success)
+            ThrowIfFailed(ref result);
+    }
+
+    /// <summary>
+    /// Converts managed <see cref="StreamConfigurationOptions"/> to the native struct,
+    /// applying defaults for unset values.
+    /// </summary>
+    public static CStreamConfigurationOptions ConvertConfig(StreamConfigurationOptions? options)
```
teodordelibasic-db, line 1001: `ConvertConfig` calls the native `GetDefaultConfig()` when `options` is null ([`NativeInterop.cs:405`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/dotnet/src/Zerobus/Native/NativeInterop.cs#L405)), but when `options` is set it fills the unset fields from the hardcoded C# defaults (`StreamConfigurationOptions.Default`, [`NativeInterop.cs:407`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/dotnet/src/Zerobus/Native/NativeInterop.cs#L407)). The record's default values ([`StreamConfigurationOptions.cs`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/dotnet/src/Zerobus/StreamConfigurationOptions.cs)) copy the Rust `zerobus_get_default_config` values ([`lib.rs:1651`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/rust/ffi/src/lib.rs#L1651)). If Rust changes a default, the C# record would quietly fall out of sync. We could seed the record from `GetDefaultConfig()` once, or add a test that asserts the two sets of defaults are equal so any drift is caught.

dotnet/src/Zerobus/Native/HeadersProviderBridge.cs
```diff
@@ -0,0 +1,81 @@
...
+                };
+                Marshal.StructureToPtr(cHeader, headerPtr, false);
+                idx++;
+            }
+
+            result.Headers = arrayPtr;
+            result.Count = (nuint)headers.Count;
+            result.ErrorMessage = IntPtr.Zero;
+        }
+        catch (Exception ex)
```
teodordelibasic-db, line 81: In `NativeCallback` the header array is allocated ([`HeadersProviderBridge.cs:44`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/dotnet/src/Zerobus/Native/HeadersProviderBridge.cs#L44)) and then each key and value string is allocated in the loop. If one of those allocations throws (for example out of memory in `AllocUtf8String`), the catch block ([`HeadersProviderBridge.cs:63`](https://github.com/databricks/zerobus-sdk/blob/7e2c27264bf22cc4a6a469142141d48b4d0f6b37/dotnet/src/Zerobus/Native/HeadersProviderBridge.cs#L63)) sets `Headers` to zero and only allocates an error message. The array and strings already allocated are leaked, because Rust can only free what is returned in `CHeaders`. In an auth callback that may run again and again on transient failures, this adds up. We could track the pointers allocated and free them in the catch or a finally before returning the error.

dotnet/src/Zerobus/StreamConfigurationOptions.cs
```diff
@@ -0,0 +1,83 @@
...
+    /// Flush operation timeout in milliseconds.
+    /// Default: 300,000 (5 minutes).
+    /// </summary>
+    public ulong FlushTimeoutMs { get; init; } = 300_000;
+
+    /// <summary>
+    /// Type of record to ingest (Proto, Json, or Unspecified).
+    /// Default: <see cref="Zerobus.RecordType.Proto"/>.
+    /// </summary>
+    public RecordType RecordType { get; init; } = RecordType.Proto;
```
teodordelibasic-db, line 73: A caller can make a stream with `RecordType.Json` and still call the `byte[]` (proto) overload, or make a proto stream with no descriptor, and only find out when Rust rejects it later. Since one `ZerobusStream` exposes every ingest overload, examples cannot catch this at compile time. Not a blocker, but before publishing we could add typed factories such as `CreateJsonStream` and `CreateProtoStream(descriptor)` that return stream types exposing only the matching overloads. That would also remove the overlap between `TableProperties.DescriptorProto` and `RecordType`. If that is too much, we could at least document that JSON streams must set `RecordType.Json` and use the string overloads.

dotnet/src/Zerobus/Native/NativeInterop.cs
sp-jcberleur: fixed in 227f9d3

dotnet/src/Zerobus/Native/NativeInterop.cs
sp-jcberleur: fixed in 227f9d3

dotnet/src/Zerobus/ZerobusStream.cs
sp-jcberleur: added reader/writer lock in d42ee27

dotnet/src/Zerobus/Native/HeadersProviderBridge.cs
sp-jcberleur: fixed in ae79b44

dotnet/src/Zerobus/ZerobusStream.cs
sp-jcberleur: fixed in a697429

dotnet/README.md
sp-jcberleur: I think the reader/writer lock in a697429 fixes this

dotnet/src/Zerobus/ZerobusStream.cs
sp-jcberleur: added in 514c7f9, c52ed4f & 38f7c0d

It did require ffi changes, if you want I can remove them from this PR and do a follow up

dotnet/src/Zerobus/ZerobusSdk.cs
sp-jcberleur: fixed in 6633302

dotnet/tests/Zerobus.IntegrationTests/IntegrationTests.cs
sp-jcberleur: added more tests in e47ad12

dotnet/src/Zerobus/ZerobusSdk.cs
sp-jcberleur: improved docs in e185a24

dotnet/src/Zerobus/Native/NativeInterop.cs
sp-jcberleur: fixed in 3dd542d

dotnet/src/Zerobus/Native/HeadersProviderBridge.cs
sp-jcberleur: fixed in f32ff50

dotnet/src/Zerobus/StreamConfigurationOptions.cs
sp-jcberleur: added in fbd1c96, f75a54b

teodordelibasic-db (dismissed):
LGTM to merge, we should just rebase, if you think there are any follow ups feel free to create issues so we can track.

We also have to see how we will be able to merge this, CI is broken for forks for the last couple of months, we are fine with granting you write access but can't seem to find the GitHub user for some reason.

teodordelibasic-db:
> LGTM to merge, we should just rebase, if you think there are any follow ups feel free to create issues so we can track.
> 
> We also have to see how we will be able to merge this, CI is broken for forks for the last couple of months, we are fine with granting you write access but can't seem to find the GitHub user for some reason.

Okay I can't grant you write access because you are not a member of the `databricks` GitHub org,  so we will probably have to go with me pushing your branch to main and opening a PR so that CI can run normally. Are you fine with that? If yes, we can proceed once conflicts are resolved.

sp-jcberleur:
> > LGTM to merge, we should just rebase, if you think there are any follow ups feel free to create issues so we can track.
> > We also have to see how we will be able to merge this, CI is broken for forks for the last couple of months, we are fine with granting you write access but can't seem to find the GitHub user for some reason.
> 
> Okay I can't grant you write access because you are not a member of the `databricks` GitHub org, so we will probably have to go with me pushing your branch to main and opening a PR so that CI can run normally. Are you fine with that? If yes, we can proceed once conflicts are resolved.

yes that's fine, conflicts are fixed

zlata-stefanovic-db:
Hi @sp-jcberleur our CI can't run on PRs from forks, so I've re-opened your commits (authorship preserved) as #536 where CI can run. Along the way I made a few fixes to get it green: SHA-pinned the GitHub Actions (repo security policy), ran cargo fmt, and sorted out the CI registry/runner setup.

Follow #536 for updates. Thanks for the contribution!

teodordelibasic-db:
We have merged PR https://github.com/databricks/zerobus-sdk/pull/536. Thank you a lot for contributing and sorry for this being so delayed! We will look into publishing this in the following days and get back to you just to confirm we have done everything needed.

teodordelibasic-db:
Feel free to close this PR 🙂

PR #188, opened 2026-04-02
[Rust] Multiplexed streams core

danilonajkov-db:
Added a MultiplexedStream abstraction that manages N sub-streams behind the scenes. Records are distributed
  across streams via round-robin, giving customers higher throughput with a simple API.

- Bitshift-encoded MessageId: Instead of maintaining a HashMap to map mux-level IDs to (stream, offset) pairs, we encode the stream index in the upper 6 bits of the offset returned by the sub-stream. This gives O(1) encode/decode with zero allocations and no locks on the hot path. Supports up to 64 sub-streams.
- Poison-on-failure semantics: If any sub-stream fails, the entire MultiplexedStream shuts down — flushes
  healthy streams best-effort, collects all unacked records, and blocks further ingestion. Callers can
  retrieve unacked records via get_unacked_records().
- MessageId API: The multiplexed stream returns opaque MessageId values (not raw offsets) from ingest_record
   / ingest_records. Callers use wait_for_message_id to await acknowledgment. This makes it clear that these
  IDs are not ordered offsets.

1. Once stream builder API is merged , ill expand it to be the only way of creating multiplexed streams. It will handle creating the streams and managing this multiplex object. For now this PR just has the core logic for multiplexing
2. Once the PR for better rust mock tests is merged Ill expand the testing sute with more and better tests.

dont worry most of the LOC is tests.

All current changes are currently guarded by "testing" flag

## How is this tested?

New unit test suite, benchmarked performance increase


danilonajkov-db:
> I thought that we decided to go into the approach of this being like a default thing? With a global ZerobusStream that would act as a single or as MultiplexStream depending on the feature? Is there any specific reason you decided not to?

@elenagaljak-db yes, but we decided to slightly change the contracts for multiplexed streams. I think a light wrapper over streams would make most sense now, for maintainability as well
[See the doc](https://docs.google.com/document/d/1DKBi_67_eUX629Zc0m56s0SPoGvUVnnw0svePe-uP00/edit?tab=t.0#heading=h.4cz970y1mk93)

rust/sdk/src/lib.rs
```diff
@@ -41,6 +41,8 @@ mod errors;
 mod headers_provider;
 mod landing_zone;
 mod offset_generator;
+#[cfg(feature = "testing")]
```
danilonajkov-db, line 51: Guarded all new stuff with testing feature until I integrate it with the stream builder API in a following PR

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -0,0 +1,402 @@
...
+/// `to_string()` per record in a hot loop will have a performance penalty.
+impl std::fmt::Display for MessageId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "MessageId(stream={}, offset={})", self.stream_index(), self.sub_offset())
+    }
+}
+
+impl MessageId {
+    fn new(stream_index: usize, sub_offset: OffsetId) -> Self {
+        Self(((stream_index as i64) << (64 - STREAM_BITS)) | (sub_offset & OFFSET_MASK))
```
danilonajkov-db, line 56: I'm not sure if we need some protections here as we can use only 58 bits out of the 64. I don't really want to add checks as this message_id is on the hot path and constructed for every ingest_record.

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -0,0 +1,409 @@
...
+
+    pub async fn ingest_record(
+        &self,
+        payload: impl Into<EncodedRecord>,
+    ) -> ZerobusResult<MessageId> {
+        self.check_closed()?;
+        let record = payload.into();
+        let idx = self.pick_substream();
+        let stream = &self.streams[idx];
+        self.wait_for_capacity(stream, idx).await?;
```
danilonajkov-db, line 233: so basically once a stream is chosen for a record it stays assigned to it. Meaning if that stream is full it waits until there is space.
I generally think this is okay, as streams getting filled up would be quite rare and they empty their queue quite fast as acks come in.
We can further improve this in the future if needed.

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -0,0 +1,407 @@
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
```
teodordelibasic-db, line 21: Let's add a top-level file doc comment explaining this functionality.

rust/sdk/src/multiplexed_stream.rs
teodordelibasic-db: Eh, since at 1M records per second it would take 90 years to hit the limit, or something along those lines, I think we should be fine for now. We can think about guarding somehow.

rust/sdk/src/multiplexed_stream.rs
teodordelibasic-db: Sounds good. You can leave a TODO comment to maybe first check if stream is full and if it is go to the next one.

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -0,0 +1,407 @@
...
+        }
+
+        match first_error {
+            Some(e) => Err(e),
+            None => Ok(()),
+        }
+    }
+
+    pub fn is_closed(&self) -> bool {
+        self.is_closed.load(Ordering::Relaxed)
```
teodordelibasic-db: If one of the inner streams dies asynchronously and no caller subsequently touches it (no `ingest_*`, `flush`, or `wait_for_message_id` to that stream), this `is_closed()` will keep returning false and round-robin writes that hit healthy inner streams are accepted.

We can either check all inner streams with `self.is_closed.load(...) || self.streams.iter().any(|s| s.is_closed())` or have some sort of supervisor task / callback that flips the outer flag.

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -0,0 +1,407 @@
...
+        if self.is_closed.swap(true, Ordering::Relaxed) {
+            return;
+        }
+
+        warn!(
+            num_streams = self.streams.len(),
+            "MultiplexedStream poisoned due to sub-stream failure"
+        );
+
+        let flush_results = join_all(self.streams.iter().map(|s| s.flush())).await;
```
teodordelibasic-db, line 143: We should probably try to call `close()` on each inner stream here to stop all the tasks and free resources.

`close` needs `mut` on `self` so it may be a bit tricky, LLM suggests interior mutability with `tokio::sync::Mutex<Vec<ZerobusStream>>` or per-stream `Mutex`, but I am not sure that's okay from a perf standpoint.

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -0,0 +1,407 @@
...
+        } else {
+            warn!(stream_index = idx, error = %e, "Ingest errored but sub-stream still alive");
+        }
+    }
+
+    pub async fn ingest_record(
+        &self,
+        payload: impl Into<EncodedRecord>,
+    ) -> ZerobusResult<MessageId> {
+        self.check_closed()?;
```
teodordelibasic-db, line 229: If `check_closed()` passes here, then while we are waiting inside `wait_for_capacity()` another task calls `shutdown_on_failure`, we will ingest a record through a healthy inner stream.

We can check `self.is_closed` inside the `wait_for_capacity` loop and check again after it returns, before calling `ingest_record_offset`. Same fix in `ingest_records`.

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -0,0 +1,407 @@
...
+    pub async fn ingest_records<I, T>(&self, payload: I) -> ZerobusResult<Option<MessageId>>
+    where
+        I: IntoIterator<Item = T>,
+        T: Into<EncodedRecord>,
+    {
+        let records: Vec<EncodedRecord> = payload.into_iter().map(Into::into).collect();
+        if records.is_empty() {
+            return Ok(None);
+        }
+        self.check_closed()?;
```
teodordelibasic-db: nit: We can move this to the top.

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -0,0 +1,407 @@
...
+                    "Sub-stream {} closed unexpectedly",
+                    idx
+                )));
+            }
+
+            if stream.has_capacity() {
+                return Ok(());
+            }
+
+            tokio::time::sleep(std::time::Duration::from_millis(backoff_ms)).await;
```
teodordelibasic-db, line 185: We can look into exposing a `Notify` or watch channel from `ZerobusStream`/`LandingZone` that signals that capacity is available or an inner stream is closed etc. Then await on that instead of doing a sleep. Can be a follow up as well.

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -0,0 +1,407 @@
...
+    fn check_closed(&self) -> ZerobusResult<()> {
+        if self.is_closed.load(Ordering::Relaxed) {
+            return Err(ZerobusError::InvalidStateError(
+                "MultiplexedStream is closed".to_string(),
+            ));
+        }
+        Ok(())
+    }
+
+    async fn shutdown_on_failure(&self) {
```
teodordelibasic-db: Can we try returning the specific error or at least log it and the stream index or something like that?

teodordelibasic-db (commented):
Looks good mostly, I think we can iterate for some of the follow ups. We can also update the `NEXT_CHANGELOG.md` and add doc comments to all public facing fields/methods etc.

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: done

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: done, added logging and return of the error.

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: yep i tested with the mutex and there is a difference in performance.
So I added this `stream.signal_shutdown()` API. It works like a close but without needing the mut self. 
Signal shutdown will free up the tasks, grpc connection. Only the JoinHandles remain which are lightweight. The drop impl would then clean them up as well.

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: I think a good v2 would be to expose a Notify and implement the thing from the other comment about ingest to a stream that is free.

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: done, thanks

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: yes this is like a lazy is_closed. I didn't want to go through checking N streams on every ingest.

I think the best of both worlds would be this:
- public is_closed - which does basically this `self.is_closed.load(...) || self.streams.iter().any(|s| s.is_closed())`
- private is_closed - with the current behavior that is used throughout this file.
 
So that the customer can have the current state, but we can remain fast in the checks. WDYT?

danilonajkov-db:
> Looks good mostly, I think we can iterate for some of the follow ups. We can also update the `NEXT_CHANGELOG.md` and add doc comments to all public facing fields/methods etc.

I was planning to do this as part of the next PR where i integrate it with the stream builder API. For now it doesn't have any public methods

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -0,0 +1,471 @@
...
+    pub async fn ingest_record(
+        &self,
+        payload: impl Into<EncodedRecord>,
+    ) -> ZerobusResult<MessageId> {
+        self.check_closed()?;
+        let record = payload.into();
+        let idx = self.pick_substream();
+        let stream = &self.streams[idx];
+        self.wait_for_capacity(stream, idx).await?;
+        self.check_closed()?;
```
teodordelibasic-db, line 234: The point in one line: the contract says "once poisoned, no more ingests succeed,"
but there's a tiny timing window where a record still gets accepted with an `Ok`,
even though it'll never be acked.

I think there is a tiny window for poisoning where record gets accepted with `Ok`, but it can never be acked:

- `ingest_record` checks `is_closed`, waits for capacity, checks `is_closed`
  again, then queues the record onto the chosen lane.
- Between that second check and the actual queue, another task can poison the mux
  (set `is_closed = true`). The check already passed, so this call proceeds anyway.
- The caller gets back `Ok(MessageId)`, but the poison path may have already
  flushed past this record — so it never acks.

Second, smaller version of the same gap:

- `wait_for_capacity` only checks `landing_zone.len() < max_inflight_requests`; it
  doesn't reserve the slot.
- So several concurrent callers can all pass that check, then queue past the
  limit / block inside `LandingZone::add` — and that wait doesn't notice poison.

We can treat it as follow up and implement some sort of sub-stream reserve / try-enqueue API (reserve capacity → re-check closed → enqueue, with no new blocking point).

rust/sdk/src/multiplexed_stream.rs
teodordelibasic-db: Sounds good, we can go with that.

teodordelibasic-db (dismissed):

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: yep, the main problem with this is that the individual streams don't provide a way to get a peek in their landing zones. I'll create some new pub(crate) APIs for ingesting with reserved capacity. Something like this:

```rust
  struct MuxLane {
      stream: ZerobusStream,
      admission: tokio::sync::RwLock<()>,
  }

  let slot = lane.stream.reserve_capacity().await?
  let _stream_order = lane.stream.ingest_lock().await;

  let _admit = lane.admission.read().await;
  self.check_closed()?;
  lane.stream.check_open()?;

  let off = lane.stream.enqueue_reserved(encoded_batch, slot);
  Ok(MessageId::new(i, off))
```

I'll treat it as a must do followup before we ungate this.

teodordelibasic-db (approved):
LGTM, would probably be good if you create follow up issues so we don't lose track.

Issue #99, opened 2026-01-12
[All] Support dynamic Descriptor Proto


teodordelibasic-db:
**Problem Statement**
Creating/reading/passing a Descriptor Proto is a pain currently in Java.

**Proposed Solution**
Look into an easier dynamic way for clients to build their descriptors in code.

**Additional Context**
N/A

PR #197, opened 2026-04-16
feat: Add schema module for UC → proto descriptor conversion

flaviofcruz:
Exposes a new `schema` module with `descriptor_from_uc_columns` / `descriptor_from_uc_schema` that convert a Unity Catalog table schema — including nested STRUCT, ARRAY, and MAP columns via `type_json` — into a `prost_types::DescriptorProto`. Callers can build descriptors at runtime instead of pre-generating `.proto` files offline, and downstream consumers (e.g. Vector's Zerobus sink) can drop their own copy of this conversion in favor of the shared implementation.

Co-authored-by: Isaac

### DIFF
--- a/rust/sdk/src/schema.rs
+++ b/rust/sdk/src/schema.rs
@@ -439,6 +439,15 @@ impl MessageCollector {
         }
     }
 
+    /// Return `base` if unused in this scope, otherwise append an incrementing
+    /// suffix (`base2`, `base3`, …).
+    ///
+    /// The suffix order tracks the order in which the caller registers names,
+    /// which is ultimately the order UC returns columns / struct fields. Callers
+    /// that regenerate descriptors and compare the output bit-for-bit need to
+    /// feed this function in the same order on every run — top-level columns
+    /// are already sorted by `position` in [`descriptor_from_uc_columns`], and
+    /// struct fields preserve the `type_json` order (which UC returns stably).
     fn unique_name(&mut self, base: String) -> String {
         if self.used.insert(base.clone()) {
             return base;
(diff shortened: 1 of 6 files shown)


rust/sdk/src/schema.rs
```diff
@@ -0,0 +1,893 @@
...
+    let inner = match json.as_object() {
+        Some(obj) if obj.contains_key("name") && obj.contains_key("type") => {
+            obj.get("type").unwrap().clone()
+        }
+        _ => json,
+    };
+    parse_complex_type(&inner)
+}
+
+fn parse_complex_type(json: &JsonValue) -> Result<ComplexType, String> {
```
teodordelibasic-db: We can consider introducing a recursion limit here.

rust/sdk/src/schema.rs
```diff
@@ -0,0 +1,893 @@
...
+            for f in fields_json {
+                let fo = f
+                    .as_object()
+                    .ok_or_else(|| format!("expected field object, got {}", f))?;
+                let name = fo
+                    .get("name")
+                    .and_then(|v| v.as_str())
+                    .ok_or_else(|| "struct field missing 'name'".to_string())?
+                    .to_string();
+                let nullable = fo.get("nullable").and_then(|v| v.as_bool()).unwrap_or(true);
```
teodordelibasic-db: Nullability for top level `UcColumn.nullable` defaults to `false` if missing from REST API response, while here we default to `true`. Should we align these two?

teodordelibasic-db (dismissed):
Looks good, thanks for the change! We also have an issue for introducing support for dynamic Descriptor ([[All] Support dynamic Descriptor Proto](https://github.com/databricks/zerobus-sdk/issues/99)), which is kind of similar, eases usage without having to precompile `.proto` files.

rust/sdk/src/schema.rs
elenagaljak-db: In `generate proto` tool we put nesting level max at 100 ([code](https://github.com/databricks/zerobus-sdk/blob/main/rust/tools/generate_files/src/generate.rs#L36)), so we can use that one here too

flaviofcruz: Added the same limit

rust/sdk/src/schema.rs
flaviofcruz: Done!

flaviofcruz:
I updated the CLI tool to use the new library and added more validations in the SDK for restricted names so that SDK and CLI work in the same way.

teodordelibasic-db (approved):
LGTM, the CI failures are unrelated to this PR, but for it to pass I think you would need to push this branch directly to `databricks/zerobus-sdk` instead of from the fork.

PR #199, opened 2026-04-17
Fix ci to align with releases

elenagaljak-db:

Fix ci to align with releases

## How is this tested?

n/a


.github/workflows/release-python.yml
```diff
@@ -20,8 +17,54 @@ jobs:
       fail-fast: false
       matrix:
         target: [x86_64, aarch64]
+    permissions:
+      contents: read
+      id-token: write
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - name: Get JFrog OIDC token
```
teodordelibasic-db, line 26: This JFrog OIDC token + cargo config block is copy-pasted ~35 lines × 10+ times across these workflows. The `secure-public-registry-releases-eng` repo has shared scripts for this (`jfrog-oidc-token.sh`, `configure-cargo.sh`, `configure-pip.sh`, `configure-npm.sh`). Let's add similar scripts to `zerobus-sdk/.github/scripts/` and call them.

.github/workflows/release-go.yml
```diff
@@ -1,35 +1,113 @@
...
+          - runner: { group: databricks-protected-runner-group, labels: linux-ubuntu-latest }
+            target: x86_64-unknown-linux-gnu
+            lib_dir: linux_amd64
+            cross: false
+          - runner: { group: databricks-protected-runner-group, labels: linux-ubuntu-latest }
+            target: aarch64-unknown-linux-gnu
+            lib_dir: linux_arm64
+            cross: true
+          - runner: { group: databricks-protected-runner-group, labels: windows-server-latest }
+            target: x86_64-pc-windows-gnu
```
teodordelibasic-db, line 22: Other flows use `x86_64-pc-windows-msvc`, why do we use GNU here?

.github/workflows/release-java.yml
```diff
@@ -2,70 +2,57 @@ name: Build Java
...
-            native_dir: windows-x86_64
-            cross: false
-
-    name: Build JNI - ${{ matrix.native_dir }}
-    runs-on: ${{ matrix.runner }}
-
-    defaults:
-      run:
-        working-directory: rust
+  build-jar:
```
teodordelibasic-db, line 13: This workflow downloads JNI artifacts, builds a JAR, and uploads it as `java-jar`. But the corresponding registry workflow (`release-zerobus-sdk-java.yml` in `secure-public-registry-releases-eng`) doesn't download `java-jar`, it downloads `jni-*` artifacts directly and rebuilds the JAR itself. So the `java-jar` artifact from this workflow is never consumed. Either:
- The registry workflow should download `java-jar` instead of rebuilding (simpler, faster), or
- This workflow should only upload JNI artifacts (which `release-jni.yml` already does), making `release-java.yml` redundant.

.github/workflows/release-rust.yml
```diff
@@ -1,4 +1,6 @@
 name: Release Rust
+# LEGACY: This workflow is no longer used.
```
teodordelibasic-db: Why don't we remove this file?

.github/workflows/release-go.yml
elenagaljak-db: CGO on Windows uses GCC (MinGW), not MSVC, so the library must be built with the GNU target for ABI compatibility. It has been this way since Go was first added to the repo.

.github/workflows/release-java.yml
elenagaljak-db: I already fixed this, i just forgot to push to the branch

.github/workflows/release-rust.yml
elenagaljak-db: Eh, fair point

.github/workflows/release-python.yml
elenagaljak-db: Fixed

teodordelibasic-db (approved):

PR #200, opened 2026-04-20
feat: Add schema module for UC → proto descriptor conversion 

flaviofcruz:

Exposes a new `schema` module with `descriptor_from_uc_columns` / `descriptor_from_uc_schema` that convert a Unity Catalog table schema — including nested STRUCT, ARRAY, and MAP columns via `type_json` — into a `prost_types::DescriptorProto`. Callers can build descriptors at runtime instead of pre-generating `.proto` files offline, and downstream consumers (e.g. Vector's Zerobus sink) can drop their own copy of this conversion in favor of the shared implementation.

Co-authored-by: Isaac

## How is this tested?

Added unit tests.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -587,7 +587,7 @@ dependencies = [
  "tokio-stream",
  "tokio-util",
  "tonic 0.13.1",
- "tonic-build 0.13.1",
+ "tonic-build",
  "tracing",
 ]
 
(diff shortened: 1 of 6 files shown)


teodordelibasic-db (dismissed):

teodordelibasic-db:
@flaviofcruz I think you need to rebase onto newest main to get rid of these build failures.

teodordelibasic-db (approved):

PR #201, opened 2026-04-20
Pin python dependencies

elenagaljak-db:

Pin python dependencies

## How is this tested?

n/a


python/requirements-dev.lock
```diff
@@ -0,0 +1,718 @@
+#
+# This file is autogenerated by pip-compile with Python 3.12
```
teodordelibasic-db: The doc suggests `uv pip compile --exclude-newer
  '2026-03-19T00:00:00Z'`, did you try that?

teodordelibasic-db (dismissed):
Unsigned commits FYI.

python/requirements-dev.lock
elenagaljak-db: I added it, thanks

teodordelibasic-db (approved):

PR #224, opened 2026-04-21
feat: add connector factory for programmatic proxy configuration

flaviofcruz:

Vector provides ability to set the proxy configuration on the sink, which needs to be plumbed when building the SDK. This PR provides that ability and continues to support environment variables for proxy setup.

## How is this tested?

Built and tested with vector

### DIFF
--- a/rust/sdk/src/proxy.rs
+++ b/rust/sdk/src/proxy.rs
@@ -1,8 +1,68 @@
-use hyper_http_proxy::{Intercept, Proxy, ProxyConnector};
+use std::sync::Arc;
+
+use hyper_http_proxy::{Intercept, Proxy, ProxyConnector as HyperProxyConnector};
 use hyper_util::client::legacy::connect::HttpConnector;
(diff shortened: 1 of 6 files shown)


rust/sdk/src/proxy.rs
```diff
@@ -1,8 +1,85 @@
...
+            )),
+        })
+    }
+
+    pub(crate) fn into_inner(self) -> ProxiedConnector {
+        self.0
+    }
+}
+
+enum BuildError {
```
teodordelibasic-db: Why are we introducing this and not using `ZerobusError` directly?

teodordelibasic-db (dismissed):
LGTM, minor comment.

rust/sdk/src/proxy.rs
flaviofcruz: Fixed

teodordelibasic-db (approved):

PR #229, opened 2026-04-22
[Java] Release hotfix version 1.1.1

teodordelibasic-db:

The JNI `nativeIsClosed` implementation only checked whether the stream wrapper
was explicitly closed (via `nativeClose`), but did not check the Rust stream's
internal `is_closed` flag. This meant `isClosed()` returned `false` even after
the stream internally shut down due to non-retryable errors.

The fix checks both conditions, wrapper cleared or stream internally closed.

- Bump Java SDK version from 1.1.0 to 1.1.1
- Bump JNI crate version from 1.1.0 to 1.1.1

## How is this tested?

Compiled and verified via `cargo check -p zerobus-jni`.

### DIFF
--- a/java/CHANGELOG.md
+++ b/java/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Version changelog
 
+## Release v1.1.1
+
+### Bug Fixes
+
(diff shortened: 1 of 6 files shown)


elenagaljak-db (dismissed):
Thanks for such a quick fix. Lgtm

elenagaljak-db (approved):

PR #235, opened 2026-04-26
[Java] Add ingestRecordNoWait fire-and-forget ingestion API

### LINKED ISSUE #152: [Java] Ingest record nowait
### SDK

Java

### Description

Rust and Python SDKs have support for `ingest_record_nowait` as a fire and forget ingestion. We should have it in all SDKs.

### Proposed Solution

Make a wrapper around existing Rust implementation and expose it as a new API

### Additional Context

_No response_

danilonajkov-db:
## Summary

- Adds `ingestRecordNoWait` and `ingestRecordsNoWait` (batch) overloads on `ZerobusProtoStream` and `ZerobusJsonStream`, plus a single-record `ingestRecordNoWait` on the deprecated `ZerobusStream`.
- New JNI `nativeIngestRecordNoWait` bridge that spawns ingestion on the tokio runtime and discards the offset; batch nowait reuses the existing `nativeIngestRecordsOffset` and discards the offset.
- `flush()`, `close()`, `waitForOffset`, and the offset-returning ingest paths block on a `PendingNoWait` counter so previously spawned no-wait tasks complete first — preserves ordering between `ingestRecordNoWait` and subsequent offset / flush / close calls (a guarantee the Python wrapper does not provide).
- All ingest paths (including nowait) go through `ensureOpen()`, so a closed stream surfaces synchronously instead of accepting silently-dropped writes.
- README, NEXT_CHANGELOG, and examples updated. Reflection-based API surface test (`NoWaitApiTest`) plus a new integration test case (`IntegrationTest` Test 17) covering flush + ordering for both proto and JSON streams.

Closes #152.

## Test plan

- [x] `NoWaitApiTest` passes
- [x] `IntegrationTest` Test 17 passes against staging
- [x] Manual smoke: `ingestRecordNoWait` followed by `flush()` produces all records in the destination table
- [x] Manual smoke: `ingestRecordNoWait(A)` then `ingestRecordOffset(B)` lands A before B
- [x] Manual smoke: `ingestRecordNoWait` after the stream is closed throws `ZerobusException` synchronously


java/src/main/java/com/databricks/zerobus/ZerobusStream.java
```diff
@@ -136,6 +136,23 @@ public CompletableFuture<Void> ingestRecord(RecordType record) throws ZerobusExc
...
+   * have been acknowledged.
+   *
+   * @param record the protobuf record to ingest
+   * @throws ZerobusException if the stream is already closed, the payload is invalid, or the record
+   *     could not be queued
+   */
+  public void ingestRecordNoWait(RecordType record) throws ZerobusException {
+    ensureOpen();
+    nativeIngestRecordNoWait(nativeHandle, record.toByteArray(), false);
+  }
```
elenagaljak-db: I don't think we need to add this to `ZerobusStream` if it's deprecated (line 36)?

PR #256, opened 2026-04-28
[Go] Add dynamic proto descriptor generation from UC schema

elenagaljak-db:

 Adds runtime conversion of Unity Catalog table metadata into a protobuf `DescriptorProto`, eliminating the need for offline `.proto` file generation.
                                                                                                                                                    
  - New `DescriptorFromUcColumns` and `DescriptorFromUcSchema` functions build a `*descriptorpb.DescriptorProto` from UC column metadata at runtime 
  - New `MessageDescriptorFromUcColumns` and `MessageDescriptorFromUcSchema` return both the wire bytes (for `TableProperties.DescriptorProto`) and a `protoreflect.MessageDescriptor` (for `dynamicpb`) in a single call                                                                             
  - Supports all UC scalar types plus `STRUCT`, `ARRAY`, and `MAP` complex types                                                                  
  - New `go/examples/proto/dynamic/` example demonstrates end-to-end usage with `dynamicpb`                                                         
  - Adds `google.golang.org/protobuf` as a direct dependency 

## How is this tested?
30 unit tests added covering all type aliases, complex type nesting, error paths, `UcColumn` JSON unmarshalling, and `sanitizeMessageName`


teodordelibasic-db:
This is the descriptor-generation half of the Go dynamic-protobuf work now tracked in #376 (under umbrella #99).

PR #258, opened 2026-04-29
[All] Graceful close on arrow streams

elenagaljak-db:
 - Adds stream_paused_max_wait_time_ms config option to ArrowStreamConfigurationOptions across all SDKs (Rust, Go, Python, TypeScript, Java)
  - Implements client-side graceful close state machine: when the server signals it will close the stream, the SDK enters a paused state, stops sending new batches but drains in-flight ack responses before triggering recovery                                                                                                                                                                                          
  - Graceful close recoveries don't count against the recovery_retries budget, since they're intentional server-side restarts (e.g. rolling deploys)   

## How is this tested?

 - 5 new integration tests in rust/tests/src/arrow_tests.rs covering: full server duration wait, immediate recovery, client-capped wait, early exit when all batches acked, ingest accepted during grace period

### DIFF
--- a/go/NEXT_CHANGELOG.md
+++ b/go/NEXT_CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### New Features and Improvements
 
-- **Arrow Flight — graceful stream close**: When the server signals an impending close, the client pauses sends, drains in-flight acks within a bounded wait, then recovers; graceful-close recoveries do not count toward `RecoveryRetries`.
+- **Arrow Flight — graceful stream close**: When the server signals an impending close, the client pauses sends, drains in-flight acks within a bounded wait, then recovers.
 - **`ArrowStreamConfigurationOptions.StreamPausedMaxWaitTimeMs`**: Optional `*uint64` limiting how long to wait (ms) while paused (`nil` = full server duration, `0` = immediate recovery).
 
 ### Deprecations
--- a/go/NEXT_CHANGELOG.md
+++ b/go/NEXT_CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ### New Features and Improvements
 
+- **Arrow Flight — graceful stream close**: When the server signals an impending close, the client pauses sends, drains in-flight acks within a bounded wait, then recovers; graceful-close recoveries do not count toward `RecoveryRetries`.
+- **`ArrowStreamConfigurationOptions.StreamPausedMaxWaitTimeMs`**: Optional `*uint64` limiting how long to wait (ms) while paused (`nil` = full server duration, `0` = immediate recovery).
+
 ### Deprecations
 
 ### Bug Fixes
--- a/java/NEXT_CHANGELOG.md
+++ b/java/NEXT_CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ### New Features and Improvements
 
+- **Arrow Flight — graceful stream close**: On server signaled close, the client stops sending new batches, drains in-flight acknowledgments within a bounded wait, then recovers; recoveries driven only by graceful close do not count against `recoveryRetries`.
+- **`ArrowStreamConfigurationOptions`**: Added `streamPausedMaxWaitTimeMs` for the maximum time (milliseconds) to wait in the paused state during graceful close (`null` = full server duration, `0` = immediate recovery).
+
 ### Bug Fixes
 
 ### Documentation
--- a/java/NEXT_CHANGELOG.md
+++ b/java/NEXT_CHANGELOG.md
@@ -7,7 +7,7 @@
 ### New Features and Improvements
 
 - **Arrow Flight — graceful stream close**: On server signaled close, the client stops sending new batches, drains in-flight acknowledgments within a bounded wait, then recovers.
-- **`ArrowStreamConfigurationOptions`**: Added `streamPausedMaxWaitTimeMs` for the maximum time (milliseconds) to wait in the paused state during graceful close (`null` = full server duration, `0` = immediate recovery).
+- **`ArrowStreamConfigurationOptions`**: Added `streamPausedMaxWaitTimeMs` for the maximum time (milliseconds) to wait in the paused state during graceful close (`-1` = full server duration, `0` = immediate recovery).
 
 ### Bug Fixes
 
--- a/java/NEXT_CHANGELOG.md
+++ b/java/NEXT_CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### New Features and Improvements
 
-- **Arrow Flight — graceful stream close**: On server signaled close, the client stops sending new batches, drains in-flight acknowledgments within a bounded wait, then recovers; recoveries driven only by graceful close do not count against `recoveryRetries`.
+- **Arrow Flight — graceful stream close**: On server signaled close, the client stops sending new batches, drains in-flight acknowledgments within a bounded wait, then recovers.
 - **`ArrowStreamConfigurationOptions`**: Added `streamPausedMaxWaitTimeMs` for the maximum time (milliseconds) to wait in the paused state during graceful close (`null` = full server duration, `0` = immediate recovery).
 
 ### Bug Fixes
--- a/java/src/main/java/com/databricks/zerobus/ArrowStreamConfigurationOptions.java
+++ b/java/src/main/java/com/databricks/zerobus/ArrowStreamConfigurationOptions.java
@@ -151,7 +151,7 @@ public class ArrowStreamConfigurationOptions {
    * Returns the maximum time in milliseconds to wait during graceful stream close.
    *
    * <p>When the server sends a close stream signal, the SDK enters a "paused" state where it stops
-   * accepting new batches but continues processing acknowledgments for in-flight batches.
+   * sending new batches but continues processing acknowledgments for in-flight batches.
    *
    * <p>Values: -1 = wait full server duration (default), 0 = immediate recovery, &gt;0 = wait up to
    * min(this, server_duration) milliseconds.
--- a/java/src/main/java/com/databricks/zerobus/ArrowStreamConfigurationOptions.java
+++ b/java/src/main/java/com/databricks/zerobus/ArrowStreamConfigurationOptions.java
@@ -30,6 +30,7 @@ public class ArrowStreamConfigurationOptions {
   private long flushTimeoutMs = 300000;
   private long connectionTimeoutMs = 30000;
   private IPCCompressionType ipcCompression = IPCCompressionType.NONE;
+  private long streamPausedMaxWaitTimeMs = -1;
 
   private ArrowStreamConfigurationOptions() {}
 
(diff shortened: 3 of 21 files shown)


rust/sdk/src/arrow_stream.rs
```diff
@@ -1119,19 +1141,105 @@ impl ZerobusArrowStream {
...
+                                    if wait_duration_ms == 0 {
+                                        info!("Server will close the stream. Triggering immediate recovery.");
+                                        return Err(ZerobusError::StreamClosedError(
+                                            tonic::Status::unavailable(
+                                                "Immediate recovery on close signal",
+                                            ),
+                                        ));
+                                    }
+
+                                    is_paused.store(true, Ordering::Relaxed);
```
teodordelibasic-db, line 1215: I think the `is_paused` flag is not actually checked where it should be checked. We flip it to true here once a close signal is received, but `send_flight_data_internal` never reads it. Ingest calls during the grace period continue to push `FlightData` onto the existing `batch_tx` `mpsc` channel and out over the DoPut stream that the server has already announced it is closing.

In `send_flight_data_internal`, after appending to `pending_batches`, we can skip the `sender.send(...)` loop while `is_paused.load(Relaxed)` is true (still return `Ok(offset_id)`). The supervisor will initialize a fresh `batch_tx` on recovery and the pending batches will be replayed.

java/src/main/java/com/databricks/zerobus/ArrowStreamConfigurationOptions.java
```diff
@@ -144,6 +147,21 @@ public IPCCompressionType ipcCompression() {
     return this.ipcCompression;
   }
 
+  /**
+   * Returns the maximum time in milliseconds to wait during graceful stream close.
+   *
+   * <p>When the server sends a close stream signal, the SDK enters a "paused" state where it stops
+   * accepting new batches but continues processing acknowledgments for in-flight batches.
```
teodordelibasic-db: I think this is a bit confusingly worded, the SDK doesn't stop accepting batches, it just stops sending them.

typescript/src/lib.rs
```diff
@@ -1190,6 +1194,7 @@ impl From<ArrowStreamConfigurationOptions> for RustArrowStreamOptions {
             flush_timeout_ms: opts.flush_timeout_ms.map(|v| v as u64).unwrap_or(default.flush_timeout_ms),
             connection_timeout_ms: opts.connection_timeout_ms.map(|v| v as u64).unwrap_or(default.connection_timeout_ms),
             ipc_compression,
+            stream_paused_max_wait_time_ms: opts.stream_paused_max_wait_time_ms.map(|v| v as u64),
```
teodordelibasic-db: Since TS SDK depends on a pinned Rust SDK version, I don't think we should update TS here, but instead first release a new Rust SDK version and then use that version in TS.

python/rust/src/arrow.rs
```diff
@@ -136,6 +136,11 @@ pub struct ArrowStreamConfigurationOptions {
     /// IPC compression codec. Default: IPCCompression.NONE
     #[pyo3(get, set)]
     pub ipc_compression: IPCCompression,
+
+    /// Maximum time in milliseconds to wait during graceful stream close.
+    /// None = wait full server duration, 0 = immediate recovery, >0 = wait up to min(this, server_duration).
+    #[pyo3(get, set)]
+    pub stream_paused_max_wait_time_ms: Option<i64>,
```
teodordelibasic-db, line 143: Let's just add a negative check in `to_rust` like we do for other fields.

rust/sdk/src/arrow_stream.rs
```diff
@@ -794,9 +802,18 @@ impl ZerobusArrowStream {
...
                     server_error_tx.clone(),
                     Arc::clone(&last_acked_records),
+                    Arc::clone(&is_paused),
+                    &options,
                 )
                 .await;
 
+                // Capture whether this was a graceful close recovery before resetting.
+                // Graceful close recoveries should not consume the retry budget.
+                let was_graceful_close = is_paused.load(Ordering::Relaxed);
```
teodordelibasic-db: The supervisor decides whether to skip retry-budget accounting by reading
`is_paused` after `process_acks` returns:

But for two of the three close-signal branches in `process_acks`,
`is_paused` is never set to `true` before the function returns:

- `stream_paused_max_wait_time_ms == Some(0)` - returns at line 1205 without
  touching `is_paused`.
- `wait_duration_ms == 0` (server says "close now", or
  `min(client_max, server_duration) == 0`) - returns at line 1218 without
  touching `is_paused`.

In both cases `was_graceful_close` is observed `false`, so the supervisor
falls through to the `else` branch at line 836 and runs
`recovery_attempts.fetch_add(1, Relaxed)` - consuming a retry slot.

rust/sdk/src/arrow_stream.rs
elenagaljak-db: Fixed

java/src/main/java/com/databricks/zerobus/ArrowStreamConfigurationOptions.java
elenagaljak-db: Fixed

typescript/src/lib.rs
elenagaljak-db: Ok, removing it

python/rust/src/arrow.rs
elenagaljak-db: Added

rust/sdk/src/arrow_stream.rs
elenagaljak-db: To be honest, I'm not sure why that check was even there. I removed it since imo its perfectly fine for recovery to take 1 attemp, thats how it works in regular grpc stream too

teodordelibasic-db: Let's just update the `NEXT_CHANGELOG.md` file to not say that recovery doesn't count towards the retry budget.

python/NEXT_CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 ### New Features and Improvements
 
+- **Arrow Flight — graceful stream close**: On server signaled close, the client pauses sending, drains in-flight acks within a bounded wait, then recovers; those recoveries do not consume the `recovery_retries` budget.
```
teodordelibasic-db: Same as for Rust.

java/NEXT_CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 ### New Features and Improvements
 
+- **Arrow Flight — graceful stream close**: On server signaled close, the client stops sending new batches, drains in-flight acknowledgments within a bounded wait, then recovers; recoveries driven only by graceful close do not count against `recoveryRetries`.
```
teodordelibasic-db: Same as for Rust.

go/NEXT_CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 ### New Features and Improvements
 
+- **Arrow Flight — graceful stream close**: When the server signals an impending close, the client pauses sends, drains in-flight acks within a bounded wait, then recovers; graceful-close recoveries do not count toward `RecoveryRetries`.
```
teodordelibasic-db: Same as for Rust.

java/NEXT_CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
...
+- **`ArrowStreamConfigurationOptions`**: Added `streamPausedMaxWaitTimeMs` for the maximum time (milliseconds) to wait in the paused state during graceful close (`null` = full server duration, `0` = immediate recovery).
```
teodordelibasic-db: I think this is also incorrect, `-1` is for full duration from code.

rust/sdk/src/arrow_stream.rs
```diff
@@ -1456,15 +1572,15 @@ impl ZerobusArrowStream {
                 let current_ack = *offset_rx.borrow_and_update();
                 if let Some(ack_offset) = current_ack {
                     if ack_offset >= offset_to_wait {
-                        debug!(
+                        info!(
```
teodordelibasic-db: I think this may be unintentional, Flavio changed to debug in his PR.

rust/sdk/src/arrow_stream.rs
```diff
@@ -1456,15 +1572,15 @@ impl ZerobusArrowStream {
...
                             ack_offset = ack_offset,
                             target_offset = offset_to_wait,
                             "{} completed",
                             operation_name
                         );
                         return Ok(());
                     }
-                    trace!(
+                    debug!(
```
teodordelibasic-db, line 1582: Same here.

rust/sdk/src/arrow_stream.rs
```diff
@@ -794,9 +802,14 @@ impl ZerobusArrowStream {
                     ack_timeout,
                     server_error_tx.clone(),
                     Arc::clone(&last_acked_records),
+                    Arc::clone(&is_paused),
+                    &options,
                 )
                 .await;
 
+                // Reset paused state after process_acks returns (before recovery).
+                is_paused.store(false, Ordering::Relaxed);
```
teodordelibasic-db: Resetting `is_paused` here leaves a window between this line and the
`*tx_guard = None` at line 857 (which spans the `recovery_backoff_ms`
sleep at line 849) where `ingest_batch` will skip the pause check at
`send_flight_data_internal:1351` and write to the old, about-to-be-closed
`batch_tx`. The batch is also pushed to `pending_batches`, so reconnect
will replay it and correctness is preserved (the server should dedupe by
`offset_id`), but the duplicate send to a stream the server has already
announced it is closing is avoidable.

We can move `is_paused.store(false, Ordering::Relaxed)` to right
after the successful reconnect (next to `recovery_attempts.store(0, ...)`
at line 880) so the pause gate stays closed until the new sender is
installed.

python/NEXT_CHANGELOG.md
elenagaljak-db: Fixed

java/NEXT_CHANGELOG.md
elenagaljak-db: Fixed

go/NEXT_CHANGELOG.md
elenagaljak-db: Fixed

java/NEXT_CHANGELOG.md
elenagaljak-db: Fixed

rust/sdk/src/arrow_stream.rs
elenagaljak-db: Fixed

rust/sdk/src/arrow_stream.rs
elenagaljak-db: Fixed

teodordelibasic-db (approved):

PR #267, opened 2026-05-07
Fix h2 protocol errors

elenagaljak-db:
Zerobus server was getting h2 protocol errors sporadically on stream closure from client side, causing noise in logs. The issue was caused by how we handled receiver and sender task abortion, creating a race condition between `END_STREAM` and `RST_STREAM` signals over which one was sent to the server first. If `END_STREAM` is sent first everything is handled properly, but if `RST_STREAM` is sent first, the h2 protocol errors would occur. The fix introduces another cancellation token to ensure proper ordering

## How is this tested?

Tested on staging server.

### DIFF
--- a/rust/sdk/src/lib.rs
+++ b/rust/sdk/src/lib.rs
@@ -1134,6 +1134,12 @@ impl ZerobusStream {
 
             // 3. Spawn receiver and sender task.
             let is_paused = Arc::new(AtomicBool::new(false));
+
+            // Per-stream child token
+            let per_stream_token = cancellation_token.child_token();
+            // Separate token for recv_task's close path
+            let recv_drain_token = CancellationToken::new();
+
             let mut recv_task = Self::spawn_receiver_task(
                 response_grpc_stream,
                 logical_last_received_offset_id_tx.clone(),
@@ -1142,7 +1148,8 @@ impl ZerobusStream {
                 Arc::clone(&is_paused),
                 options.clone(),
                 server_error_tx.clone(),
-                cancellation_token.clone(),
+                per_stream_token.clone(),
+                recv_drain_token.clone(),
                 callback_tx.clone(),
             );
             let mut send_task = Self::spawn_sender_task(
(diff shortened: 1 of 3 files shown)


rust/sdk/src/lib.rs
```diff
@@ -1869,6 +1897,36 @@ impl ZerobusStream {
                     }
                 }
             }
+
+            // Drain remaining server messages so the server sees END_STREAM instead of
+            // the client RST_STREAM-ing the response. Inline on close (runtime may exit
+            // right after); detached on recovery / errors so recovery isn't delayed.
+            if close_initiated {
+                let _ = tokio::time::timeout(Duration::from_millis(500), async {
```
teodordelibasic-db: In the supervisor task, in the branch where `send_task` finishes first the drain is capped at 100ms:

```rust
if matches!(send_result, Ok(Ok(()))) && cancellation_token.is_cancelled() {
    recv_drain_token.cancel();
    let _ = tokio::time::timeout(Duration::from_millis(100), &mut recv_task).await;
}
recv_task.abort();
```

Here the drain is capped at 500 ms.

We can align the two by introducing a constant and we should also correct the `NEXT_CHANGELOG.md` if needed.

rust/sdk/src/lib.rs
```diff
@@ -1706,33 +1722,42 @@ impl ZerobusStream {
         is_paused: Arc<AtomicBool>,
         options: StreamConfigurationOptions,
         server_error_tx: tokio::sync::watch::Sender<Option<ZerobusError>>,
-        cancellation_token: CancellationToken,
+        _cancellation_token: CancellationToken,
```
teodordelibasic-db: Do we need `_cancellation_token` here now that it is unused?

rust/sdk/src/lib.rs
teodordelibasic-db: In the supervisor task, in the branch where `send_task` finishes first the drain is capped at 100ms:

```rust
if matches!(send_result, Ok(Ok(()))) && cancellation_token.is_cancelled() {
    recv_drain_token.cancel();
    let _ = tokio::time::timeout(Duration::from_millis(100), &mut recv_task).await;
}
recv_task.abort();
```

Here the drain is capped at 500 ms.

We can align the two by introducing a constant and we should also correct the `NEXT_CHANGELOG.md` if needed.

rust/sdk/src/lib.rs
elenagaljak-db: No, we don't. Thanks

teodordelibasic-db (approved):

rust/sdk/src/lib.rs
elenagaljak-db: I put them both at 500ms.

PR #270, opened 2026-05-11
feat: add new SDK function to convert UC schema to arrow schema

flaviofcruz:

  **WHAT**

  Adds two new public SDK functions, gated behind the existing `arrow-flight` feature:

  - `schema::arrow_schema_from_uc_columns(&[UcColumn]) -> Result<arrow_schema::Schema, SchemaError>`
  - `schema::arrow_schema_from_uc_schema(&UcTableSchema) -> Result<arrow_schema::Schema, SchemaError>`

  They parallel the existing `descriptor_from_uc_*` functions and produce a native Arrow schema using the Spark/Delta IPC conventions the Databricks Arrow Flight server expects:

  - `STRING` / `VARIANT` / `DECIMAL` → `LargeUtf8`, `BINARY` → `LargeBinary`
  - `DATE` → `Date32`, `TIMESTAMP` → `Timestamp(Microsecond, Some("UTC"))`, `TIMESTAMP_NTZ` → `Timestamp(Microsecond, None)`
  - `ARRAY<T>` → `List(Field::new("item", T, true))`
  - `MAP<K,V>` → `Map(Field::new("key_value", Struct{key,value}, false))`

  Also folded in a small internal refactor so both backends share a single source of truth for top-level UC type names (`parse_uc_top_level_type` → `PrimitiveType` → per-backend projection), and
  extracted `validate_map_key` / `shape_unsupported` helpers to eliminate the literal-duplicate map-key validation and "not supported" error strings between the proto and Arrow paths.

  **WHY**

  Vector's `databricks_zerobus` sink today derives its Arrow schema by going UC → proto descriptor → Arrow (in `vector/src/sinks/databricks_zerobus/proto_to_arrow.rs`). That round-trip loses UC type
  fidelity: `DATE` becomes Arrow `Int32` instead of `Date32`, `TIMESTAMP` becomes `Int64` instead of `Timestamp(Micros, UTC)`. Going UC → Arrow directly inside the SDK gives any wrapper a
  type-correct schema in one call and centralizes the Spark/Delta/Flight conventions (LargeUtf8, LargeBinary, "key_value" map entries) so wrappers can't drift.

  The function lives in the same module as `descriptor_from_uc_*` and reuses the existing `ComplexType` AST / `parse_type_json` pipeline — no new parser, no exposed internals.

### DIFF
--- a/rust/sdk/src/schema.rs
+++ b/rust/sdk/src/schema.rs
@@ -681,7 +681,9 @@ fn sanitize_message_name(name: &str) -> String {
 /// `STRING` / `VARIANT` / `DECIMAL` → `LargeUtf8`, `BINARY` → `LargeBinary`,
 /// `DATE` → `Date32`, `TIMESTAMP` → `Timestamp(Microsecond, Some("UTC"))`,
 /// `TIMESTAMP_NTZ` → `Timestamp(Microsecond, None)`, `ARRAY<T>` → `List` with
-/// item field `"item"`, `MAP<K,V>` → `Map` with entries field `"key_value"`.
+/// item field `"item"`, `MAP<K,V>` → `Map` with entries field `"entries"`
+/// containing `"keys"` and `"values"` (the canonical schema the Databricks
+/// Arrow Flight server builds from Delta).
 #[cfg(feature = "arrow-flight")]
 pub fn arrow_schema_from_uc_columns(
     columns: &[UcColumn],
@@ -738,7 +740,8 @@ fn map_primitive_to_arrow(p: PrimitiveType) -> arrow_schema::DataType {
         PrimitiveType::String => DataType::LargeUtf8,
         PrimitiveType::Long => DataType::Int64,
         PrimitiveType::Integer => DataType::Int32,
-        PrimitiveType::Short | PrimitiveType::Byte => DataType::Int32,
+        PrimitiveType::Short => DataType::Int16,
+        PrimitiveType::Byte => DataType::Int8,
         PrimitiveType::Double => DataType::Float64,
         PrimitiveType::Float => DataType::Float32,
         PrimitiveType::Boolean => DataType::Boolean,
@@ -746,6 +749,10 @@ fn map_primitive_to_arrow(p: PrimitiveType) -> arrow_schema::DataType {
         PrimitiveType::Timestamp => DataType::Timestamp(TimeUnit::Microsecond, Some("UTC".into())),
         PrimitiveType::TimestampNtz => DataType::Timestamp(TimeUnit::Microsecond, None),
         PrimitiveType::Date => DataType::Date32,
+        // TODO: emit Decimal128(precision, scale) once the Databricks Arrow
+        // Flight server accepts native Decimal128. UC carries (p, s) in
+        // `type_text` ("decimal(10,2)") and the `type_json` primitive string,
+        // but `PrimitiveType::Decimal` discards them today.
         PrimitiveType::Decimal => DataType::LargeUtf8,
     }
 }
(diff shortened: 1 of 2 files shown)


rust/sdk/src/schema.rs
```diff
@@ -663,6 +666,157 @@ fn sanitize_message_name(name: &str) -> String {
...
+            let key_primitive = validate_map_key(key, name)?;
+            let value_field = match value.as_ref() {
+                ComplexType::Primitive(p) => Field::new("value", map_primitive_to_arrow(*p), true),
+                ComplexType::Struct(_) => complex_type_to_arrow_field("value", value, true)?,
+                ComplexType::Array(_) | ComplexType::Map { .. } => {
+                    return Err(shape_unsupported("maps with complex value types", name));
+                }
+            };
+            let entries = DataType::Struct(Fields::from(vec![
+                Field::new("key", map_primitive_to_arrow(key_primitive), false),
```
teodordelibasic-db: Here the entries struct is built as `Map("key_value", Struct{"key", "value"})`, but Shinkansen's server-side converter (`universe/shinkansen/core/server/src/type_converter.rs`) builds `Map("entries", Struct{"keys", "values"})`. These should match I think.

rust/sdk/src/schema.rs
```diff
@@ -663,6 +666,157 @@ fn sanitize_message_name(name: &str) -> String {
...
+        PrimitiveType::Integer => DataType::Int32,
+        PrimitiveType::Short | PrimitiveType::Byte => DataType::Int32,
+        PrimitiveType::Double => DataType::Float64,
+        PrimitiveType::Float => DataType::Float32,
+        PrimitiveType::Boolean => DataType::Boolean,
+        PrimitiveType::Binary => DataType::LargeBinary,
+        PrimitiveType::Timestamp => DataType::Timestamp(TimeUnit::Microsecond, Some("UTC".into())),
+        PrimitiveType::TimestampNtz => DataType::Timestamp(TimeUnit::Microsecond, None),
+        PrimitiveType::Date => DataType::Date32,
+        PrimitiveType::Decimal => DataType::LargeUtf8,
```
teodordelibasic-db, line 756: We are currently in the process of adding `DECIMAL` support on server side. For Arrow it will be mapped to `Decimal128(precision, scale)`. Currently this PR drops the precision and scale.

We can try to plumb precision/scale through `PrimitiveType` or special-case `DECIMAL` earlier in `parse_uc_top_level_type` / `parse_primitive_type` to capture the `decimal(p,s)` parameters from `type_name` / `type_text` / the JSON form, then emit `DataType::Decimal128(p, s)` on the Arrow side. Both UC's `type_text` (`"decimal(10,2)"`) and the JSON primitive string `"decimal(10,2)"` carry it.

We can leave this as a follow-up.

rust/sdk/src/schema.rs
```diff
@@ -663,6 +666,157 @@ fn sanitize_message_name(name: &str) -> String {
...
+}
+
+#[cfg(feature = "arrow-flight")]
+fn map_primitive_to_arrow(p: PrimitiveType) -> arrow_schema::DataType {
+    use arrow_schema::{DataType, TimeUnit};
+    match p {
+        PrimitiveType::String => DataType::LargeUtf8,
+        PrimitiveType::Long => DataType::Int64,
+        PrimitiveType::Integer => DataType::Int32,
+        PrimitiveType::Short | PrimitiveType::Byte => DataType::Int32,
```
teodordelibasic-db: Let's switch to this to match server side:

```rust
PrimitiveType::Short => DataType::Int16,
PrimitiveType::Byte => DataType::Int8,
```

teodordelibasic-db (dismissed):

rust/sdk/src/schema.rs
flaviofcruz: Great point, now the server does not need to do the recast.

rust/sdk/src/schema.rs
flaviofcruz: Added a TODO. Will follow up

rust/sdk/src/schema.rs
flaviofcruz: Done

teodordelibasic-db (approved):

PR #288, opened 2026-05-18
[Python] Version 1.3.0 - Arrow Flight Beta

teodordelibasic-db:

Promotes Arrow Flight API for Python to Beta and releases version 1.3.0.

## How is this tested?

Arrow examples and tests.

### DIFF
--- a/python/rust/src/sync_wrapper.rs
+++ b/python/rust/src/sync_wrapper.rs
@@ -1,122 +1,49 @@
-use std::future::Future;
-use std::pin::Pin;
 use std::sync::Arc;
 
 use pyo3::prelude::*;
-use pyo3::types::{PyBytes, PyList};
+use pyo3::types::PyBytes;
 use tokio::runtime::Runtime;
 use tokio::sync::RwLock;
 
 use databricks_zerobus_ingest_sdk::{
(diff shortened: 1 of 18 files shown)


python/rust/src/sync_wrapper.rs
```diff
@@ -442,63 +330,25 @@ impl ZerobusSdk {
...
-                callback_max_wait_time_ms: opts.callback_max_wait_time_ms.map(|v| v as u64),
-                ack_callback,
-                ..Default::default()
-            }
-        } else {
-            RustStreamOptions::default()
-        };
-
+        let opts = options.unwrap_or_default();
+        opts.validate()?;
```
elenagaljak-db, line 334: Claude says that this is changing the default from the rust one which is 1_000_000 to python one which is 50_000.

elenagaljak-db: for max-inflight_records

teodordelibasic-db: Hm, fixed, but I think this was set from before Python transitioned to being a Rust wrapper. I think now there is no reason to bump it to Rust default.

elenagaljak-db (dismissed):

nikolaobradovic-db (dismissed):

elenagaljak-db (approved):

PR #290, opened 2026-05-18
[Go] Go v1.2.0, ingest_record_nowait, Arrow beta

elenagaljak-db:

- **Arrow Flight Beta** — promotes Arrow Flight ingestion from Experimental/Unsupported to Beta. The API is stabilising but may still change before reaching GA.
- **Fire-and-forget ingestion** — new `IngestRecordNowait` and `IngestRecordsNowait` methods on `ZerobusStream`. Both return immediately after spawning a background task; ingestion errors are silently ignored. Useful when you want to queue records without blocking at all. Call `Flush()` or `Close()` before exiting to ensure durability.

## How is this tested?

existing tests and new test for ingest_nowait

### DIFF
--- a/go/ffi.go
+++ b/go/ffi.go
@@ -105,6 +105,22 @@ extern int64_t zerobus_stream_ingest_json_records(CZerobusStream* stream,
                                                     const char** json_records,
                                                     uintptr_t num_records,
                                                     CResult* result);
+extern void zerobus_stream_ingest_proto_record_nowait(CZerobusStream* stream,
+                                                       const uint8_t* data,
+                                                       uintptr_t data_len,
+                                                       CResult* result);
+extern void zerobus_stream_ingest_json_record_nowait(CZerobusStream* stream,
+                                                      const char* json_data,
+                                                      CResult* result);
+extern void zerobus_stream_ingest_proto_records_nowait(CZerobusStream* stream,
+                                                        const uint8_t** records,
+                                                        const uintptr_t* record_lens,
+                                                        uintptr_t num_records,
+                                                        CResult* result);
+extern void zerobus_stream_ingest_json_records_nowait(CZerobusStream* stream,
+                                                       const char** json_records,
+                                                       uintptr_t num_records,
+                                                       CResult* result);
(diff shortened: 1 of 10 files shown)


teodordelibasic-db (approved):
Looks good.

go/ffi.go
teodordelibasic-db, line 6: We are only adding back `darwin_arm64/` in this PR.

Can we either drop this line or cross-compile for Intel Mac as well?

go/ffi.go
```diff
@@ -506,6 +522,134 @@ func streamIngestJSONRecord(streamPtr unsafe.Pointer, jsonData string) (int64, e
...
+
+	recordPtrs := (*[1 << 30]*C.uint8_t)(cPtrArray)[:len(records):len(records)]
+	recordLens := (*[1 << 30]C.size_t)(cLenArray)[:len(records):len(records)]
+
+	// Pin Go memory so it isn't moved while Rust copies it (before returning).
+	var pinner runtime.Pinner
+	defer pinner.Unpin()
+
+	for i, record := range records {
+		if len(record) > 0 {
```
teodordelibasic-db, line 596: If any `records[i]` has `len(record) == 0`, the corresponding slot in `recordPtrs` / `recordLens` is left holding whatever bytes `malloc` returned. The Rust side (`rust/ffi/src/lib.rs:1566-1574`) then runs:

```rust
.map(|(ptr, len)| std::slice::from_raw_parts(*ptr, *len).to_vec())
```

A garbage `*len` can trigger undefined behavior or a segfault.

We can use `C.calloc` instead of `C.malloc`, or explicitly set `recordPtrs[i] = nil; recordLens[i] = 0` in an `else` branch.

This is present in `streamIngestProtoRecords` as well.

teodordelibasic-db (commented):
Looks good.

go/ffi.go
elenagaljak-db: Sure, I will try to cross compile

go/ffi.go
elenagaljak-db: Fixed

teodordelibasic-db (approved):

PR #330, opened 2026-05-28
Add client-side warning to detect non-optimal use of the SDKs

irinatomic-db:

Adds a process-wide client-side warning to `ZerobusStream` and `ZerobusArrowStream` that surface common stream misuse patterns at runtime.

**Stream churn** - fires a `WARN` when 100 or more streams for the same table are opened within a 60-second sliding window. Re-fires after the rate recovers and rises again.

The warning lives in a new `client_warnings` module and can be suppressed with `ZEROBUS_SDK_WARNINGS_ENABLED=false`.

**Why**: Users sometimes accidentally create a new stream per record instead of reusing streams. This pattern causes unnecessary load on the service and is hard to spot from the outside. 

Python and TypeScript pick these up automatically; Go and Java will on the next FFI release.

 ## How is this tested?

Added unit tests.

For manual verification, a temporary test was added that initialised a tracing subscriber and opened streams in a loop to confirm `WARN` line fired correctly. It was removed before this PR.

### DIFF
--- a/rust/NEXT_CHANGELOG.md
+++ b/rust/NEXT_CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### New Features and Improvements
 
+- Added a process-wide stream churn warning: logs a `WARN` when 100 or more streams for the same table are opened within a 60-second sliding window, which may indicate a "one stream per record" misuse pattern. Applies to `ZerobusStream` and `ZerobusArrowStream`. Set `ZEROBUS_SDK_WARNINGS_ENABLED=false` to suppress.
+
 ### Bug Fixes
 
 ### Documentation
--- a/rust/README.md
+++ b/rust/README.md
@@ -18,6 +18,7 @@ A high-performance Rust client for streaming data ingestion into Databricks Delt
   - [5. Ingest Data](#5-ingest-data)
   - [6. Handle Acknowledgments](#6-handle-acknowledgments)
   - [7. Close the Stream](#7-close-the-stream)
+- [Client-side warnings](#client-side-warnings)
 - [Configuration Options](#configuration-options)
 - [Error Handling](#error-handling)
 - [Examples](#examples)
@@ -675,6 +676,18 @@ match stream.close().await {
     Ok(_) => println!("Stream closed successfully"),
 }
 ```
+
+## Client-side warnings
+
+The SDK logs a `WARN`-level message via [`tracing`](https://docs.rs/tracing) when **100 or more** streams for the same table are opened within a 60-second sliding window. This usually indicates a "one stream per record" misuse pattern. The warning fires again if the rate drops below the threshold and later surges again.
+
+To suppress, set the environment variable before starting the process:
+
+```bash
+ZEROBUS_SDK_WARNINGS_ENABLED=false
+```
+
+Also accepts `0` or `no`.
 
 ## Configuration Options
 
--- a/rust/sdk/src/client_warnings.rs
+++ b/rust/sdk/src/client_warnings.rs
@@ -1,0 +1,294 @@
+//! Process-wide client-side warnings that surface common SDK misuse patterns.
+//!
+//! A warning is emitted via [`tracing::warn!`] when 100 or more streams for
+//! the same table are opened within a 60-second sliding window, which usually
+//! indicates a "one stream per record" misuse pattern.
+//!
+//! The warning is process-wide and keyed by table name.
+//!
+//! ## Opt-out
+//!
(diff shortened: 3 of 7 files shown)


rust/sdk/src/lib.rs
```diff
@@ -611,6 +614,10 @@ impl ZerobusStream {
             )
         })??);
 
+        client_warnings::record_stream_opened(&table_properties.table_name);
```
dbatomic: Both `new` and `recreate_stream` hit this path. If I am reading this correctly, we do double counting during `recreate`.

Ideally, `recreate` stream shouldn't count as churn.

rust/NEXT_CHANGELOG.md
```diff
@@ -6,6 +6,12 @@
 
 ### New Features and Improvements
 
+- Added process-wide client-side warnings to surface common misuse patterns:
+  - **Concurrent open streams**: logs a `WARN` when 32 or more ingest streams for the same table are open simultaneously.
```
dbatomic: Do we really need both patterns? They are kind of doing a similar thing.
Can we just have 100 per 60 seconds?

What is the scenario where we care about 32 concurrent streams?

rust/sdk/src/client_warnings.rs
```diff
@@ -0,0 +1,447 @@
...
+    CHURN_MONITOR.get_or_init(|| Mutex::new(StreamChurnState::new()))
+}
+
+/// Records one stream open for `table_name`.
+///
+/// Maintains a per-table sliding window of open timestamps. Logs a `WARN`-level
+/// message the first time the count within the window reaches
+/// [`CHURN_WARN_THRESHOLD`] (100). The warning re-fires if the rate drops below
+/// the threshold and later surges again in a new window.
+pub(crate) fn record_stream_opened(table_name: &str) {
```
dbatomic: Can we do better naming? `register_stream_opened` and `record_stream_opened` sound the same.
(ideally, let's just keep only churn_threshold = 100 so we don't need to do both.

rust/NEXT_CHANGELOG.md
irinatomic-db: Good point.

Main idea was to check if someone is misusing the SDK (treating it like a file push system, opening 1 stream for each record). But the stream churn warning is enough. I'll keep only that one.

rust/README.md
```diff
@@ -676,6 +657,28 @@ match stream.close().await {
 }
 ```
 
+## Client-side warnings
```
elenagaljak-db, line 680: We should add this entry in table of contents

elenagaljak-db:
Btw, you need to setup signed commits. Since you've already pushed commits, it needs to be done "retroactively". I usually do it like this:
```
> git rebase --exec 'git commit --amend --no-edit -S' origin/main
> git push origin HEAD:<branch_name> --force-with-lease
```

rust/sdk/src/client_warnings.rs
irinatomic-db: Dropped the concurrent monitor entirely

rust/sdk/src/lib.rs
irinatomic-db: Fixed. Moved the call out of `new_stream()` and into the builder, so `recreate_stream` and `recreate_arrow_stream` no longer count as churn.

rust/sdk/src/client_warnings.rs
```diff
@@ -0,0 +1,291 @@
...
+            .map(|&t| now.saturating_sub(t) > CHURN_WINDOW_MS)
+            .unwrap_or(false)
+        {
+            deque.pop_front();
+        }
+        deque.push_back(now);
+        deque.len()
+    };
+
+    // Fire exactly when count reaches the threshold; re-fires after rate recovers.
```
teodordelibasic-db: I see we mention at several places how the warning re-fires after rate recovers, but I don't see we are ever decrementing the number of opened streams. Do we want a `notify_stream_closed` function maybe?

teodordelibasic-db (commented):
I think the PR description is maybe outdated since it mentions two types of warnings.

rust/sdk/src/client_warnings.rs
irinatomic-db: The monitor tracks timestamps of when streams were created, not open/close state - the count is just "how many creation events in the last 60 seconds". Timestamps older than the window are evicted lazily on each new creation, so the count drops naturally over time without any close signal. 

Renamed `notify_stream_opened` → `record_stream_creation` to make this clearer.

teodordelibasic-db (approved):
LGTM

PR #331, opened 2026-05-28
[Rust] Enforce 10 MiB client-side payload limit on ingest calls

### LINKED ISSUE #101: [All] Introduce gRPC message size limit
**Problem Statement**
Introduce message size limit.

**Proposed Solution**
Check the size of the serialized record received from the client.

**Additional Context**
The server has a limit of 10MB per received message. We should introduce this on client side as well.

elenagaljak-db:

Enforce the Zerobus server's 10 MiB payload limit client-side on `ingest_record_offset` and `ingest_records_offset`. Without this, an oversized payload travels through the landing zone and gRPC layer before being rejected by the server. With this change, the SDK returns `ZerobusError::InvalidArgument` immediately, before any network I/O.

The limit applies to the total encoded byte size of a single call (sum of all record bytes in the batch). It does not apply to Arrow Flight (`ingest_batch` / `ingest_ipc_batch`), which has a separate code path.

**Perf impact:** measured with Criterion: the check costs ~3 ns for a single record and ~20 ns for a 100-record batch (reading `.len()` from a `SmallVec`, no allocation). Well within the noise of the mutex + gRPC overhead.

## Changes

- `sdk/src/lib.rs` — module-level `MAX_INGEST_PAYLOAD_BYTES` constant (10 MiB) with a note that it mirrors the server limit; check at the top of `ingest_internal_v2`
- `sdk/src/record_types.rs` — `EncodedBatch::total_byte_size()` (pub) + unit tests
- `tests/src/rust_tests.rs` — integration tests for single-record and batch rejection
- `README.md` — new "Payload Size Limit" subsection under Error Handling
- `NEXT_CHANGELOG.md` — entry under New Features and Improvements

## How is this tested?

`cargo test --workspace`

### DIFF
--- a/rust/sdk/src/builder/stream_builder.rs
+++ b/rust/sdk/src/builder/stream_builder.rs
@@ -267,6 +267,18 @@ impl<'a> StreamBuilder<'a> {
         self
     }
 
+    /// Set the maximum total encoded byte size allowed per ingest call.
+    ///
+    /// This is the sum of all record bytes passed to a single ingest call.
+    /// Calls exceeding this limit fail fast with
+    /// [`ZerobusError::InvalidArgument`] before any network I/O.
+    ///
+    /// Defaults to 10 MiB, which matches the server-side limit.
+    pub fn max_ingest_payload_bytes(mut self, bytes: usize) -> Self {
+        self.grpc_config.max_ingest_payload_bytes = bytes;
+        self
+    }
+
     /// Set the maximum wait time during graceful stream pause (JSON/proto and Arrow streams).
     pub fn stream_paused_max_wait_time_ms(mut self, ms: Option<u64>) -> Self {
         self.grpc_config.stream_paused_max_wait_time_ms = ms;
@@ -550,6 +562,25 @@ mod tests {
         let builder = sdk.stream_builder().table("t").oauth("a", "b").json();
         assert_eq!(builder.grpc_config.max_inflight_requests, 1_000_000);
         assert!(builder.grpc_config.recovery);
+        assert_eq!(
+            builder.grpc_config.max_ingest_payload_bytes,
+            10 * 1024 * 1024
+        );
+    }
+
+    #[test]
+    fn max_ingest_payload_bytes_override() {
+        let sdk = test_sdk();
+        let builder = sdk
+            .stream_builder()
+            .table("t")
+            .oauth("a", "b")
+            .json()
+            .max_ingest_payload_bytes(5 * 1024 * 1024);
+        assert_eq!(
+            builder.grpc_config.max_ingest_payload_bytes,
+            5 * 1024 * 1024
+        );
     }
 
     #[tokio::test]
--- a/rust/sdk/src/builder/stream_builder.rs
+++ b/rust/sdk/src/builder/stream_builder.rs
@@ -267,13 +267,18 @@ impl<'a> StreamBuilder<'a> {
         self
     }
 
-    /// Set the maximum total encoded byte size allowed per ingest call.
+    /// Set the maximum total encoded record byte size allowed per ingest call
+    /// (gRPC JSON/proto streams only).
     ///
     /// This is the sum of all record bytes passed to a single ingest call.
     /// Calls exceeding this limit fail fast with
     /// [`ZerobusError::InvalidArgument`] before any network I/O.
     ///
-    /// Defaults to 10 MiB, which matches the server-side limit.
+    /// Defaults to slightly below the 10 MiB server limit.
+    ///
+    /// Note: this setting only applies to streams built with [`build()`](Self::build).
+    /// Arrow Flight streams (built with `build_arrow()`) do not read this value
+    /// and have no client-side payload-size enforcement.
     pub fn max_ingest_payload_bytes(mut self, bytes: usize) -> Self {
         self.grpc_config.max_ingest_payload_bytes = bytes;
         self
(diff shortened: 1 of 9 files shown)


rust/sdk/src/lib.rs
```diff
@@ -108,6 +108,10 @@ pub mod zeroparser;
 
 const SHUTDOWN_TIMEOUT_SECS: u64 = 1;
 
+/// Maximum encoded byte size allowed per `ingest_record_offset` / `ingest_records_offset` call.
+/// Matches the server limit so oversize payloads fail fast client-side.
+const MAX_INGEST_PAYLOAD_BYTES: usize = 10 * 1024 * 1024; // 10 MiB
```
teodordelibasic-db: I think we should expose this as configurable for the client for forward compatibility. For example if in the future we allow larger messages then 10MB on the server and someone is using the next version of the SDK containing this change, they would have to wait for a new version that also bumps the message size limit in the SDK. Perhaps that version contains breaking changes and they don't want to upgrade. By exposing it as configurable and setting the default to the server side limit, if they increase it over the limit it's on them.

teodordelibasic-db (commented):
LGTM, I just think this should be configurable.

rust/sdk/src/lib.rs
elenagaljak-db: Yep, I agree. Fixed

rust/sdk/src/stream_configuration.rs
```diff
@@ -157,6 +157,17 @@ pub struct StreamConfigurationOptions {
...
     /// Default: `Some(5000)` (wait 5 seconds)
     pub callback_max_wait_time_ms: Option<u64>,
+
+    /// Maximum total encoded byte size allowed per ingest call.
+    ///
+    /// This is the sum of all record bytes passed to a single
+    /// `ingest_record()` / `ingest_records()` (and their `_offset` variants) call.
+    /// Calls exceeding this limit fail fast with
+    /// [`ZerobusError::InvalidArgument`](crate::ZerobusError::InvalidArgument)
+    /// before any network I/O, matching the server-side limit.
```
teodordelibasic-db, line 167: The client-side check matches the application server-side check for 10 MiB, but server-side we also have the transport-layer tonic check on the full serialized `EphemeralStreamRequest`. This means that 10 MiB of raw record bytes that pass the client-side check can potentially fail server-side in the transport layer.

We should probably reword the comments/documentation surrounding this and maybe set the default to slightly less than 10 MiB. Worth a test probably.

rust/sdk/src/builder/stream_builder.rs
```diff
@@ -240,6 +240,18 @@ impl<'a> StreamBuilder<'a> {
...
     }
 
+    /// Set the maximum total encoded byte size allowed per ingest call.
+    ///
+    /// This is the sum of all record bytes passed to a single ingest call.
+    /// Calls exceeding this limit fail fast with
+    /// [`ZerobusError::InvalidArgument`] before any network I/O.
+    ///
+    /// Defaults to 10 MiB, which matches the server-side limit.
+    pub fn max_ingest_payload_bytes(mut self, bytes: usize) -> Self {
```
teodordelibasic-db, line 282: Builder docstring should probably say this is gRPC-only - Arrow Flight streams use `.build_arrow()` and don't read `grpc_config.max_ingest_payload_bytes`. Right now a user calling `.arrow(schema).max_ingest_payload_bytes(X).build_arrow()` would silently have no client-side enforcement.

rust/tests/src/rust_tests.rs
```diff
@@ -984,6 +984,64 @@ mod schema_tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn test_ingest_record_too_large_fails() -> Result<(), Box<dyn std::error::Error>> {
```
teodordelibasic-db, line 988: We can add a test that a payload at exactly `max_ingest_payload_bytes` is accepted.

rust/tests/src/rust_tests.rs
```diff
@@ -984,6 +984,64 @@ mod schema_tests {
...
+        let oversized = vec![0u8; 10 * 1024 * 1024 + 1];
+        let result = stream.ingest_record_offset(oversized).await;
+
+        assert!(matches!(result, Err(ZerobusError::InvalidArgument(_))));
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_ingest_batch_too_large_fails() -> Result<(), Box<dyn std::error::Error>> {
```
teodordelibasic-db: I think the wording here is a bit confusing maybe since it hints at this test being for Arrow `ingest_batch`, but instead it's for `ingest_records_offset`.

rust/sdk/src/stream_configuration.rs
elenagaljak-db: Fixed

rust/sdk/src/builder/stream_builder.rs
elenagaljak-db: Added

rust/tests/src/rust_tests.rs
elenagaljak-db: Done

rust/tests/src/rust_tests.rs
elenagaljak-db: Fixed

teodordelibasic-db (dismissed):
LGTM!

teodordelibasic-db (approved):

PR #332, opened 2026-05-29
builder/stream_builder: validate UC three-part table name format in validate()

Dev-X25874:

**WHAT:** `StreamBuilder::validate()` now checks that the table name follows the
required `catalog.schema.table` format, returning `ZerobusError::InvalidTableName`
immediately if it does not.

**WHY:** `validate()` previously only checked that the table name was non-empty.
A malformed name like `"mytable"` or `"schema.table"` would pass `validate()` and
`build()`, only to fail later inside the async OAuth token fetch — far from the
misconfigured builder. The three-part invariant was already enforced in
`DefaultTokenFactory::parse_table_name`; this PR moves that check to the earliest
possible point so the error fires synchronously, before any I/O occurs.

## How is this tested?

Three unit tests added to the existing `#[cfg(test)]` block in `stream_builder.rs`:
- `validate_rejects_single_part_table_name`
- `validate_rejects_two_part_table_name`
- `validate_rejects_table_name_with_empty_part`

All three fail on the original code and pass after the fix.

### DIFF
--- a/rust/sdk/src/builder/stream_builder.rs
+++ b/rust/sdk/src/builder/stream_builder.rs
@@ -599,6 +599,50 @@ mod tests {
         assert!(debug_str.contains("Json"));
     }
 
+    #[test]
+    fn validate_rejects_single_part_table_name() {
+        let sdk = test_sdk();
+        let builder = sdk.stream_builder().table("mytable").oauth("a", "b").json();
+        match builder.validate() {
+            Err(ZerobusError::InvalidTableName(msg)) => {
+                assert!(msg.contains("exactly 3 parts"), "unexpected: {}", msg);
+            }
+            other => panic!("expected InvalidTableName, got {:?}", other),
+        }
+    }
+
+    #[test]
+    fn validate_rejects_two_part_table_name() {
+        let sdk = test_sdk();
+        let builder = sdk
+            .stream_builder()
+            .table("schema.table")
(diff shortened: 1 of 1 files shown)


rust/sdk/src/builder/stream_builder.rs
```diff
@@ -309,6 +309,30 @@ impl<'a> StreamBuilder<'a> {
                 "table name is required: call .table()".into(),
             ));
         }
+        {
```
teodordelibasic-db, line 313: We should guard this block with `#[cfg(not(feature = "testing"))]` because we use non-three-part table names when testing.

teodordelibasic-db (commented):
LGTM, thanks for contributing, let's only do this when `testing` feature is off.

Dev-X25874:
@teodordelibasic-db Done, guarded the block with #[cfg(not(feature = "testing"))] in the latest commit.

teodordelibasic-db (commented):
LGTM, you just need to rebase to newest `main`.

Dev-X25874:
@teodordelibasic-db Done, rebased onto latest main. Ready for another look!

teodordelibasic-db:
Ah so we are hitting an issue where CI is broken for external contributions from forks because of the need to fetch a JFrog token to access the proxy registry. I have asked internally if there is an established solution for this.

PR #333, opened 2026-05-30
[Rust] Set zerobuswrite operation on token mint

jovanpavl-db:
Add "operations": ["zerobuswrite"] to the TABLE authorization detail when minting OAuth tokens, in both the SDK token factory and the generate_files tool. This is done in order to make all priviliges working with the new sdk.

## How is this tested?
Tested on dev.


teodordelibasic-db (approved):
LGTM, thanks for contributing!

elenagaljak-db (approved):

PR #358, opened 2026-06-11
[Rust][FFI] Expose SDK builder from Rust

teodordelibasic-db:

Expose the Rust `ZerobusSdkBuilder` over the C FFI as a builder API, so
wrapper SDKs can configure SDK construction options that the existing
three-argument `zerobus_sdk_new` constructor cannot reach. Bumps the FFI
crate from `1.2.1` to `1.3.0`. ABI back-compat for `zerobus_sdk_new` is
preserved.

This PR is the FFI half of a two-part change. The matching Go SDK PR
sits on top of this release and exposes the new options as Go-level
`SdkOption` functions (`WithApplicationName`, `WithNoTLS`).

### Motivation

The legacy `zerobus_sdk_new(endpoint, unity_catalog_url, result)` cannot
configure:

- the SDK identifier on the `user-agent` header (so the Go SDK currently
  wears the Rust SDK's default identifier — a bug the matching Go PR
  fixes);
- an application name to append to the `user-agent` header;
- a no-TLS channel option for local development against plain-HTTP
  Shinkansen servers.

Adding a `zerobus_sdk_new_with_options(...)` variant would force a new
function per option set and risk an ABI break the next time an option
landed. A builder API is purely additive: each new option becomes a new
setter function with no impact on any existing entrypoint.

### New API surface

```c
struct CZerobusSdkBuilder;  // opaque

CZerobusSdkBuilder* zerobus_sdk_builder_new(void);

void zerobus_sdk_builder_endpoint(CZerobusSdkBuilder*, const char* value);
void zerobus_sdk_builder_unity_catalog_url(CZerobusSdkBuilder*, const char* value);
void zerobus_sdk_builder_sdk_identifier(CZerobusSdkBuilder*, const char* value);
void zerobus_sdk_builder_application_name(CZerobusSdkBuilder*, const char* value);
void zerobus_sdk_builder_disable_tls(CZerobusSdkBuilder*);

// Consumes the builder. Frees it on both success and failure paths;
// any further use of the pointer is undefined behavior.
CZerobusSdk* zerobus_sdk_builder_build(CZerobusSdkBuilder*, CResult* result);

// Drops a builder without consuming it (use only when abandoning a
// builder between _new and _build). No-op on NULL.
void zerobus_sdk_builder_free(CZerobusSdkBuilder*);
```

`zerobus.h` is regenerated by cbindgen.

### What stays the same

- `zerobus_sdk_new(endpoint, unity_catalog_url, result)` is retained with
  the same signature for ABI back-compat. It now internally drives the
  builder.
- `zerobus_sdk_set_use_tls` remains as a no-op (ABI placeholder, has been
  a no-op since the Rust core 2.0.0 release).

### Behavior change

`zerobus_sdk_new` no longer infers any TLS configuration from the
endpoint scheme — it constructs the SDK with the default
`SecureTlsConfig` regardless of whether the URL is `http://` or
`https://`. Callers that need a plain-HTTP channel must use the builder
API and call `zerobus_sdk_builder_disable_tls` explicitly.

In practice this is observationally a no-op for existing callers: tonic's
gRPC transport already selects plain HTTP/2 for `http://` URLs at the
scheme level, regardless of the configured `TlsConfig`.

### Implementation notes

- The opaque `*CZerobusSdkBuilder` is a reinterpret cast of
  `Box<ZerobusSdkBuilder>`. A `type SdkBuilderAlloc = ZerobusSdkBuilder;`
  alias is shared across the four cast sites (allocate, mutate, consume,
  drop) to keep them in lock-step.
- Setters mutate in place via `std::mem::take` (`ZerobusSdkBuilder`
  implements `Default`), so the slot stays valid even if a setter
  closure panics.
- `_build` reclaims ownership with `Box::from_raw` and drops on every
  outcome — null/empty/build-error all free the builder before
  returning.

### Migration for direct FFI consumers

If you currently call `zerobus_sdk_new("http://...", ...)` and relied on
TLS being disabled implicitly: explicitly call the new builder setter.

```c
CZerobusSdkBuilder* b = zerobus_sdk_builder_new();
zerobus_sdk_builder_endpoint(b, "http://...");
zerobus_sdk_builder_unity_catalog_url(b, "");
zerobus_sdk_builder_disable_tls(b);
CZerobusSdk* sdk = zerobus_sdk_builder_build(b, &result);
```

For most callers this is observationally a no-op (see "Behavior change"
above).

### Version bump

`rust/ffi/Cargo.toml`: `1.2.1` → `1.3.0`. Minor bump — new API,
back-compat preserved. `Cargo.lock` updated accordingly.

`rust/ffi/NEXT_CHANGELOG.md` documents the change. On version-bump
merge, contents move to `CHANGELOG.md` per release convention.

### Follow-up

A separate Go SDK PR (currently on `go/user-agent`) builds on this
release. After this PR merges and an `ffi/v1.3.0` release is cut, the Go
PR will:

- Add `WithApplicationName(name)` and `WithNoTLS()` as `SdkOption`
  functions on `NewZerobusSdk`.
- Set the SDK identifier to `zerobus-sdk-go/<version>` via the new
  builder setter (fixes the wire `user-agent` identification bug).
- Drop the now-dead `zerobus_sdk_set_use_tls` declaration from cgo.

### Reviewer focus

- `rust/ffi/src/lib.rs`: the `with_builder` helper, the `_build` /
  `_free` ownership story, and the `zerobus_sdk_new` delegation.
- `rust/ffi/zerobus.h`: cbindgen-regenerated; review the diff.
- `rust/ffi/src/tests.rs`: the new builder-API tests.
- `rust/ffi/Cargo.toml`: version bump.
- `rust/ffi/NEXT_CHANGELOG.md`: behavior-change wording.

## How is this tested?

10 new builder-API unit tests, all under
`rust/ffi/src/tests.rs`:

- minimal build (endpoint only)
- each setter individually + both user-agent setters together
- empty strings treated as no-ops
- error path: missing endpoint must still consume the builder
- `_free` without `_build` (abandonment)
- `_free` on null
- setters on null builder are no-ops
- `_disable_tls` for plain-http endpoint builds cleanly

Total FFI test count: **30 passing**, `cargo clippy --all -- -D warnings`
clean.

### DIFF
--- a/rust/ffi/src/lib.rs
+++ b/rust/ffi/src/lib.rs
@@ -19,7 +19,7 @@ use bytes::Bytes;
 use databricks_zerobus_ingest_sdk::databricks::zerobus::RecordType;
 use databricks_zerobus_ingest_sdk::{
     EncodedRecord, HeadersProvider, NoTlsConfig, ZerobusError, ZerobusResult, ZerobusSdk,
-    ZerobusStream,
+    ZerobusSdkBuilder, ZerobusStream,
 };
 use databricks_zerobus_ingest_sdk::{RecordBatch, StreamBuilder, ZerobusArrowStream};
 use prost::Message;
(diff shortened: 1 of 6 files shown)


rust/ffi/src/lib.rs
```diff
@@ -974,58 +974,185 @@ pub(crate) fn write_success_result(result: *mut CResult) {
...
+    _private: [u8; 0],
+}
+
+/// Concrete type behind `*mut CZerobusSdkBuilder`. All cast sites in this
+/// module must agree on this alias.
+type SdkBuilderAlloc = ZerobusSdkBuilder;
+
+/// SAFETY: `b` must be a valid pointer from `_new` that hasn't been consumed
+/// or freed. `mem::take` keeps the slot valid even if `f` panics.
+unsafe fn with_builder<F>(b: *mut CZerobusSdkBuilder, f: F)
```
elenagaljak-db, line 1000: Is this comment needed?

elenagaljak-db (approved):

rust/ffi/src/lib.rs
teodordelibasic-db: I'll revisit it in a follow up!

PR #370, opened 2026-06-17
[Rust][FFI] Expose UC-schema → protobuf descriptor + record encoding

zlata-stefanovic-db:
Add a dynamic-protobuf C FFI so a pure-C consumer (e.g. a Fluent Bit output plugin) can follow the same UC-schema → descriptor → record-encode path the Vector sink uses, without a companion Rust crate. The logic already exists in the SDK (`schema::descriptor_from_uc_schema`) and prost-reflect; these are thin wrappers over it.

## New APIs

**Opaque type:** `CZerobusProtoSchema`

**Functions:**
- `zerobus_proto_schema_from_uc_json` — build a handle from UC table-metadata JSON
- `zerobus_proto_schema_descriptor_bytes` — borrow the serialized DescriptorProto to pass to `zerobus_sdk_create_stream` (byte-identical to the descriptor the encoder uses)
- `zerobus_proto_schema_encode_json` — encode one JSON record into protobuf bytes; unknown keys are ignored
- `zerobus_free_proto_bytes` / `zerobus_proto_schema_free` — free a buffer / a handle

## Usage Example

```c
CResult r = {0};

/* init: fetch GET /api/2.1/unity-catalog/tables/{name} and pass its JSON body */
CZerobusProtoSchema *schema = zerobus_proto_schema_from_uc_json(uc_table_json, &r);

uintptr_t dlen;
const uint8_t *desc = zerobus_proto_schema_descriptor_bytes(schema, &dlen);
CZerobusStream *stream = zerobus_sdk_create_stream(sdk, table_name, desc, dlen,
                                                   client_id, client_secret, &opts, &r);

/* per record, at flush time */
uint8_t *buf; uintptr_t len;
if (zerobus_proto_schema_encode_json(schema, record_json, &buf, &len, &r)) {
    /* collect buf/len into a batch, ingest via zerobus_stream_ingest_proto_records(...) */
    zerobus_free_proto_bytes(buf, len);
}

/* shutdown */
zerobus_proto_schema_free(schema);
```

## Encoding Contract

- Record object keys are matched to column names; unknown keys are ignored (upstream records often carry extra non-column metadata)
- `DATE`/`TIMESTAMP`/`TIMESTAMP_NTZ` columns are integers (days / micros since epoch), not strings

## Testing

[...]

PR #84, opened 2026-01-28
Add JNI and C FFI

teodordelibasic-db:
## What changes are proposed in this pull request?

- Adds the JNI bindings needed for wrapping this SDK for Java.
- Adds the C FFI from Go repo so we can create a C# SDK and customers can use it for C/C++.

## How is this tested?

Tested locally for Java, C FFI is from Go.

PR #371, opened 2026-06-17
[Rust][FFI] Expose UC-schema → protobuf descriptor + record encoding

### LINKED ISSUE #377: [Rust][FFI] Expose UC-schema → protobuf descriptor + record encoding
### SDK

Rust

### Description

Add a dynamic-protobuf C FFI so a pure-C consumer (e.g. a Fluent Bit output plugin) can follow the same UC-schema → descriptor → record-encode path the Vector sink uses, without a companion Rust crate. The logic already exists in the SDK (schema::descriptor_from_uc_schema) and prost-reflect; these are thin wrappers over it.

### Proposed Solution

**Add New APIs**
Opaque type: CZerobusProtoSchema

Functions:

zerobus_proto_schema_from_uc_json — build a handle from UC table-metadata JSON
zerobus_proto_schema_descriptor_bytes — borrow the serialized DescriptorProto to pass to zerobus_sdk_create_stream (byte-identical to the descriptor the encoder uses)
zerobus_proto_schema_encode_json — encode one JSON record into protobuf bytes; unknown keys are ignored
zerobus_free_proto_bytes / zerobus_proto_schema_free — free a buffer / a handle

### Additional Context

_No response_

zlata-stefanovic-db:
Add a dynamic-protobuf C FFI so a pure-C consumer (e.g. a Fluent Bit output plugin) can follow the same UC-schema → descriptor → record-encode path the Vector sink uses, without a companion Rust crate. The logic already exists in the SDK (`schema::descriptor_from_uc_schema`) and prost-reflect; these are thin wrappers over it.

> Re-raised from #370 as a same-repo branch so CI can authenticate (fork PRs can't obtain the JFrog OIDC token). Also includes a rustfmt fix in `ffi/src/lib.rs` for the default (CI) rustfmt config.

## New APIs

**Opaque type:** `CZerobusProtoSchema`

**Functions:**
- `zerobus_proto_schema_from_uc_json` — build a handle from UC table-metadata JSON
- `zerobus_proto_schema_descriptor_bytes` — borrow the serialized DescriptorProto to pass to `zerobus_sdk_create_stream` (byte-identical to the descriptor the encoder uses)
- `zerobus_proto_schema_encode_json` — encode one JSON record into protobuf bytes; unknown keys are ignored
- `zerobus_free_proto_bytes` / `zerobus_proto_schema_free` — free a buffer / a handle

## Usage Example

```c
CResult r = {0};

/* init: fetch GET /api/2.1/unity-catalog/tables/{name} and pass its JSON body */
CZerobusProtoSchema *schema = zerobus_proto_schema_from_uc_json(uc_table_json, &r);

uintptr_t dlen;
const uint8_t *desc = zerobus_proto_schema_descriptor_bytes(schema, &dlen);
CZerobusStream *stream = zerobus_sdk_create_stream(sdk, table_name, desc, dlen,
                                                   client_id, client_secret, &opts, &r);

/* per record, at flush time */
uint8_t *buf; uintptr_t len;
if (zerobus_proto_schema_encode_json(schema, record_json, &buf, &len, &r)) {
    /* collect buf/len into a batch, ingest via zerobus_stream_ingest_proto_records(...) */
    zerobus_free_proto_bytes(buf, len);
}

/* shutdown */
zerobus_proto_schema_free(schema);
```

## Encoding Contract

- Record object keys are matched to column names; unknown keys are ignored (upstream records often carry extra non-column metadata)
- `DATE`/`TIMESTAMP`/`TIMESTAMP_NTZ` columns are integers (days / micros since epoch), not strings

## How is this tested?

Adds prost-reflect (serde) + serde_json as direct deps of the ffi crate, exports the opaque type via cbindgen (regenerated `zerobus.h`), and updates `NEXT_CHANGELOG` and `README`. Covered by 8 unit tests including a round-trip that decodes the encoded bytes and asserts field values. CI (`rust` / `cross-sdk-*` jobs) now runs against this same-repo branch.


rust/ffi/zerobus.h
```diff
@@ -497,6 +505,41 @@ void zerobus_free_error_message(char *message);
...
+ * Borrow the serialized descriptor bytes. Valid until `zerobus_proto_schema_free`.
+ * Pass directly to `zerobus_sdk_create_stream`. Returns NULL on null handle.
+ */
+const uint8_t *zerobus_proto_schema_descriptor_bytes(const struct CZerobusProtoSchema *schema,
+                                                     uintptr_t *out_len);
+
+/**
+ * Encode JSON record to protobuf bytes. Unknown keys are ignored.
+ * DATE/TIMESTAMP columns must be integers (days/micros since epoch), not strings.
+ * Returns true on success; caller must free buffer with `zerobus_free_proto_bytes`.
```
irinatomic-db, line 544: Lets' document the encoding contract for all [supported types](https://docs.databricks.com/aws/en/ingestion/zerobus-limits#type-support) and tests for the tricky ones. Specifically:
- Binary columns map to proto bytes, which prost-reflect's serde layer only accepts as a base64-encoded string
- Decimal maps to proto string
- Large BIGINT/TIMESTAMP values above 2^53 lose precision when emitted as JSON numbers by most encoders.

rust/ffi/src/lib.rs
```diff
@@ -1975,3 +1977,205 @@ pub extern "C" fn zerobus_get_default_config() -> CStreamConfigurationOptions {
...
+        Ok(s) => s,
+        Err(e) => {
+            write_error_result(result, e, false);
+            return false;
+        }
+    };
+
+    let mut deserializer = serde_json::Deserializer::from_str(&json);
+    // Records carry extra non-column fields; ignore them rather than erroring.
+    let options = DeserializeOptions::new().deny_unknown_fields(false);
```
irinatomic-db, line 2170: Proto2 allows required columns. So let's add here a required columns presence check, so validate that all required fields are present before encoding.

rust/ffi/src/lib.rs
```diff
@@ -1975,3 +1977,205 @@ pub extern "C" fn zerobus_get_default_config() -> CStreamConfigurationOptions {
...
+
+/// Safe wrapper to validate proto schema pointer
+fn validate_proto_schema_ptr<'a>(
+    schema: *const CZerobusProtoSchema,
+) -> Result<&'a ProtoSchema, &'static str> {
+    if schema.is_null() {
+        return Err("Proto schema pointer is null");
+    }
+    unsafe { Ok(&*(schema as *const ProtoSchema)) }
+}
```
irinatomic-db: From Claude:

This helper should either be marked unsafe fn or have a // SAFETY: comment explaining the invariant. It creates a &'a ProtoSchema with an unbounded lifetime fabricated from a raw pointer - soundness depends entirely on the pointer originating from _from_uc_json and not having been freed or aliased mutably. The rest of the FFI layer signals this kind of unsafe precondition explicitly (e.g. c_str_to_string is unsafe fn). Makes it easy for a future reader to miss the invariant.

rust/ffi/src/lib.rs
```diff
@@ -1975,3 +1977,205 @@ pub extern "C" fn zerobus_get_default_config() -> CStreamConfigurationOptions {
...
+#[no_mangle]
+pub extern "C" fn zerobus_free_proto_bytes(data: *mut u8, len: usize) {
+    if !data.is_null() && len > 0 {
+        unsafe {
+            // data came from Box::into_raw(bytes.into_boxed_slice()), so
+            // capacity == len and this reconstruction is sound.
+            let _ = Box::from_raw(std::ptr::slice_from_raw_parts_mut(data, len));
+        }
+    }
+}
```
irinatomic-db, line 2236: `encode_to_vec()` returns a `Vec<u8>`. If a record has every field at its default value (e.g. id = 0, empty string, zero timestamp), protobuf omits all of them - the encoded output is zero bytes. So bytes is an empty Vec, and len is 0.

`bytes.into_boxed_slice()` converts that empty Vec into a `Box<[u8]>`. In Rust, a zero-length boxed slice is valid = it has a non-null dangling pointer (Rust guarantees this; the pointer just can't be dereferenced). Then `Box::into_raw(...)` gives you that non-null pointer, which gets written into `*out_data`.

So the C caller receives: `out_data = <non-null pointer>, out_len = 0`.

The free side in `zerobus_proto_schema_free()`:
if !data.is_null() && len > 0 {
  data is non-null ✓, but len == 0 - so the condition is false and the function does nothing. The box is never reconstructed, never dropped, and the allocation is leaked.

The fix is to drop the `&& len > 0` condition - just check `!data.is_null()`. Reconstructing a zero-length box via `Box::from_raw(slice_from_raw_parts_mut(ptr, 0))` is okay in Rust.

rust/ffi/src/lib.rs
```diff
@@ -1975,3 +1977,205 @@ pub extern "C" fn zerobus_get_default_config() -> CStreamConfigurationOptions {
...
+// ============================================================================
+// Dynamic Protobuf FFI
+// ============================================================================
+//
+// Pure-C consumers can build a protobuf descriptor from Unity Catalog metadata
+// and encode JSON records to protobuf bytes without a companion Rust crate.
+// Lifecycle: `_from_uc_json` → `_descriptor_bytes` / `_encode_json` → `_free`.
+
+/// Holds a table's serialized `DescriptorProto` plus a prepared protobuf
+/// encoder built from it. Opaque to C; the concrete type is [`ProtoSchema`].
```
irinatomic-db: Given that the primary use case here is a Fluent Bit output plugin, which runs in a multi-threaded pipeline, I think we should make this safe by design rather than relying on the caller to get the threading right.

The current implementation has a UAF (Use-After-Free): a thread calling `_free` while another is mid-encode will drop the ProtoSchema box while `schema_ref.message` is still being accessed. In a Fluent Bit plugin the schema handle would be initialized once at startup and shared across worker threads - this will get hit.

The simplest fix is to wrap the handle in an `Arc: encode_json` and descriptor_bytes clone the Arc on entry (cheap - just an atomic increment), and `_free` decrements it. The memory stays live until the last reference drops, so a concurrent free is safe. The C API surface doesn't change at all - callers still call `_free` exactly once.

rust/ffi/src/tests.rs
```diff
@@ -441,4 +441,244 @@ mod tests {
...
+            decoded.get_field_by_name("ts").unwrap().as_i64(),
+            Some(1700000000000000)
+        );
+
+        zerobus_free_proto_bytes(out_data, out_len);
+        zerobus_proto_schema_free(schema);
+    }
+
+    #[test]
+    fn test_proto_schema_from_uc_json_invalid_json_errors() {
```
irinatomic-db, line 568: Nit: let's also assert on some part of the error message, like an error code

rust/ffi/src/lib.rs
zlata-stefanovic-db: Done, wrapped in Arc

rust/ffi/src/lib.rs
zlata-stefanovic-db: Done

rust/ffi/src/tests.rs
zlata-stefanovic-db: Done

rust/ffi/zerobus.h
zlata-stefanovic-db: Done, documented in detail in the FFI readme and here.

rust/ffi/src/lib.rs
zlata-stefanovic-db: Done below

rust/ffi/src/lib.rs
zlata-stefanovic-db: The helper is now an unsafe fn with a documented # Safety section, matching the c_str_to_string convention you pointed at:

/// Validate a schema handle and clone its owning [`Arc`], keeping the
/// [`ProtoSchema`] alive for the caller even if another thread frees the
/// handle concurrently.
///
/// # Safety
///
/// `schema` must be null or a live handle from
/// [`zerobus_proto_schema_from_uc_json`] not yet passed to
/// [`zerobus_proto_schema_free`].
unsafe fn clone_proto_schema_arc(
    schema: *const CZerobusProtoSchema,
) -> Result<Arc<ProtoSchema>, &'static str> { ... }

It also no longer fabricates the &'a ProtoSchema with an unbounded lifetime - it returns an owned Arc<ProtoSchema>. So soundness no longer rests on a conjured lifetime that nothing actually backs; the returned Arc carries its own validity (count>=1 for as long as it's held). Both call sites (encode_json,
descriptor_bytes) now invoke it inside an unsafe { ... } block with a //SAFETY: note stating they uphold the handle contract, so the precondition is explicit at every use.

rust/ffi/src/lib.rs
zlata-stefanovic-db: Thanks for flagging this — I dug into it and confirmed the same zero-length free pattern exists in pre-existing FFI paths, not just in the new proto-schema code.

Specifically:
- `zerobus_free_record_array` (`if !record.data.is_null() && record.data_len > 0`) was introduced in `7c3e36d4` (`Add JNI and C FFI (#84)`, Jan 29, 2026, Teodor Delibasic).
- Arrow cleanup/free paths (`if !ptr.is_null() && len > 0`) were introduced in `b129c9a2` (`[Go] Add experimental Arrow Flight ingestion support (#134)`, Mar 17, 2026, elenagaljak-db).

In this PR I fixed the same issue for the new dynamic-protobuf path (`zerobus_free_proto_bytes`) by freeing on `!data.is_null()` even when `len == 0`.

I agree we should follow up and align the older free paths to the same behavior so zero-length boxed slices cannot leak there either.

rust/ffi/src/lib.rs
zlata-stefanovic-db: Thanks — good catch. I investigated this pattern and confirmed it also exists in pre-existing stream-handle code (`clone_stream_arc`), so this PR reused an existing approach rather than introducing a brand-new one.

That said, I agree the raw-pointer `Arc::increment_strong_count` / `Arc::from_raw` handle pattern is fragile under concurrent free/use and deserves hardening. The safer direction is to avoid dereferencing raw Arc pointers from C handles (e.g., use an internal handle registry/token + cloned `Arc` lookup), so concurrent `_free` cannot race into UB.

I can open a focused follow-up to harden this pattern consistently for both stream and proto-schema handles.

rust/ffi/src/tests.rs
irinatomic-db: If there is an error code let's assert on that. Error message text can change but code shouldn't

irinatomic-db (dismissed):
Looks good to me.

In tests, let's just assert on error codes. They shouldn't change, but error messages can change.

irinatomic-db (dismissed):

rust/ffi/NEXT_CHANGELOG.md
```diff
@@ -7,6 +7,11 @@
 ### New Features and Improvements
 
 - **C-builder API for SDK construction**: `zerobus_sdk_builder_new`, per-option setters (`_endpoint`, `_unity_catalog_url`, `_sdk_identifier`, `_application_name`, `_disable_tls`), and `_build` / `_free`. Mirrors the Rust `ZerobusSdkBuilder`; new options are added as setters without ABI breaks. Legacy `zerobus_sdk_new` is retained and delegates to the builder.
+- **Dynamic protobuf from a Unity Catalog schema**: a pure-C consumer can now follow the same UC-schema → protobuf-descriptor → record-encoding path the Vector sink uses, without a companion Rust crate. New opaque type `CZerobusProtoSchema` and functions:
```
davidtosovic-db: We don't need to say that we use the same path Vector sync uses. Please check the change for any similar comments.

rust/ffi/NEXT_CHANGELOG.md
```diff
@@ -7,6 +7,11 @@
...
+  - `zerobus_proto_schema_from_uc_json` — build a schema handle from UC table-metadata JSON (the body of `GET /api/2.1/unity-catalog/tables/{name}`).
+  - `zerobus_proto_schema_descriptor_bytes` — borrow the serialized `DescriptorProto` to pass straight to `zerobus_sdk_create_stream` (byte-identical to the descriptor the encoder uses).
+  - `zerobus_proto_schema_encode_json` — encode one JSON record into protobuf bytes; unknown keys are ignored. `DATE`/`TIMESTAMP`/`TIMESTAMP_NTZ` columns are integers (days / micros since epoch), `BINARY` is a base64 string, `DECIMAL` is a string, and 64-bit integers above 2^53 should be passed as strings to avoid JSON precision loss. Non-nullable columns are proto2 `required`; a record missing one is rejected.
```
davidtosovic-db: Regarding the 2^53 and precision loss - Do SDKs behave in the same way today, e.g., using strings?

zlata-stefanovic-db: yes, other SDKs run into the same large-number issue when using JSON inputs.

This PR is not changing SDK behavior; it only documents it clearly for this FFI path:
- If values are sent as typed protobuf, big integers are fine.
- If values are sent through JSON, very large integers can be unsafe unless passed as strings (and the newly exposed functions have JSON input and protobuf output).

So this guidance is consistent with how JSON ingestion generally works across SDKs.

irinatomic-db (dismissed):

rust/ffi/README.md
```diff
@@ -58,6 +58,80 @@ private static extern IntPtr zerobus_sdk_new(string endpoint, string ucUrl, ref
...
+  past year 2255) lose precision when emitted as a JSON number by most encoders.
+  Pass them as JSON **strings** (the canonical protobuf-JSON form for 64-bit
+  ints), which round-trip exactly.
+- **`DATE`/`TIMESTAMP*` unit mismatches** are silent: writing milliseconds where
+  microseconds are expected shifts every row by 10³.
+
+Non-nullable columns become proto2 `required` fields; a record missing one is
+rejected with an error rather than encoded.
+
+The schema handle is reference-counted internally, so it is safe to share one
```
teodordelibasic-db: The "concurrent free is safe" claim may not be true. `clone_proto_schema_arc` calls `Arc::increment_strong_count` on the raw handle, whose only strong reference is the handle itself. If `zerobus_proto_schema_free` runs on another thread first, it can drop the last reference and free the allocation before the increment runs - use-after-free. The Arc only makes multiple concurrent readers safe; it cannot make `free` race-safe.

We can reword to the real contract: the handle may be shared by concurrent readers (`encode_json`, `descriptor_bytes`), but `free` must happen after all in-flight calls have returned, and be called exactly once.

rust/ffi/src/lib.rs
```diff
@@ -1975,3 +1979,266 @@ pub extern "C" fn zerobus_get_default_config() -> CStreamConfigurationOptions {
...
+        Err(e) => {
+            write_error_result(result, e, false);
+            return ptr::null_mut();
+        }
+    };
+
+    match build_proto_schema(&json) {
+        Ok(schema) => {
+            write_success_result(result);
+            // Reference-counted so a concurrent `_free` can't drop the schema
```
teodordelibasic-db: This comment overstates a bit what the Arc gives you. A `_free` racing a call that hasn't yet hit `increment_strong_count` still frees the allocation. We can reword to something like "concurrent readers are fine, `free` must not race any other use".

rust/ffi/src/lib.rs
```diff
@@ -1975,3 +1979,266 @@ pub extern "C" fn zerobus_get_default_config() -> CStreamConfigurationOptions {
...
+
+/// Concrete type behind `*mut CZerobusProtoSchema`.
+struct ProtoSchema {
+    /// Serialized `DescriptorProto` bytes (passed to `zerobus_sdk_create_stream`).
+    descriptor_bytes: Vec<u8>,
+    /// Message descriptor for encoding JSON records to protobuf.
+    message: MessageDescriptor,
+}
+
+/// Validate a schema handle and clone its owning [`Arc`], keeping the
```
teodordelibasic-db: The doc says it keeps the schema alive "even if another thread frees the handle concurrently." The real invariant is that the caller must not have freed the handle (which the Safety section already states).

rust/ffi/src/lib.rs
```diff
@@ -1975,3 +1979,266 @@ pub extern "C" fn zerobus_get_default_config() -> CStreamConfigurationOptions {
...
+        }
+    }
+}
+
+/// Free a handle from `zerobus_proto_schema_from_uc_json`.
+#[no_mangle]
+pub extern "C" fn zerobus_proto_schema_free(schema: *mut CZerobusProtoSchema) {
+    if !schema.is_null() {
+        unsafe {
+            // Release the handle's strong reference; the schema drops only when
```
teodordelibasic-db: "a free racing an in-flight call is safe" is another incorrect claim. We can reword to state that `free` must be ordered after all other use of the handle.

rust/ffi/NEXT_CHANGELOG.md
```diff
@@ -7,6 +7,11 @@
...
+- **Dynamic protobuf from a Unity Catalog schema**: a pure-C consumer can now build a protobuf descriptor from UC table metadata and encode records without a companion Rust crate. New opaque type `CZerobusProtoSchema` and functions:
+  - `zerobus_proto_schema_from_uc_json` — build a schema handle from UC table-metadata JSON (the body of `GET /api/2.1/unity-catalog/tables/{name}`).
+  - `zerobus_proto_schema_descriptor_bytes` — borrow the serialized `DescriptorProto` to pass straight to `zerobus_sdk_create_stream` (byte-identical to the descriptor the encoder uses).
+  - `zerobus_proto_schema_encode_json` — encode one JSON record into protobuf bytes; unknown keys are ignored. `DATE`/`TIMESTAMP`/`TIMESTAMP_NTZ` columns are integers (days / micros since epoch), `BINARY` is a base64 string, `DECIMAL` is a string, and large 64-bit integers are accepted as JSON strings (the protobuf-JSON canonical form) to avoid precision loss in producers that emit numbers as IEEE-754 doubles. Non-nullable columns are proto2 `required`; a record missing one is rejected.
```
teodordelibasic-db: "a record missing a non-nullable column is rejected" is only true for scalar/struct columns. Non-nullable ARRAY and MAP columns map to `repeated` (schema.rs:225), which has no presence, so an omitted one encodes as empty and is not rejected. Nested required fields inside a STRUCT aren't checked either. We should either narrow the wording to top-level scalar columns, or extend the check.

rust/ffi/src/lib.rs
```diff
@@ -1975,3 +1979,266 @@ pub extern "C" fn zerobus_get_default_config() -> CStreamConfigurationOptions {
...
+}
+
+/// Borrow the serialized descriptor bytes. Valid until `zerobus_proto_schema_free`.
+/// Pass directly to `zerobus_sdk_create_stream`. Returns NULL on null handle.
+#[no_mangle]
+pub extern "C" fn zerobus_proto_schema_descriptor_bytes(
+    schema: *const CZerobusProtoSchema,
+    out_len: *mut usize,
+) -> *const u8 {
+    // SAFETY: caller upholds the handle contract (valid, unfreed handle).
```
teodordelibasic-db, line 2096: We clone the Arc to read a pointer into it here, then drop the clone - two atomic ops per call for no benefit, since the returned pointer's validity depends on the handle's own reference anyway. We can borrow directly with `&*(schema as *const ProtoSchema)` after the null check.

rust/ffi/src/lib.rs
```diff
@@ -1975,3 +1979,266 @@ pub extern "C" fn zerobus_get_default_config() -> CStreamConfigurationOptions {
...
+
+// ============================================================================
+// Dynamic Protobuf FFI
+// ============================================================================
+//
+// Pure-C consumers can build a protobuf descriptor from Unity Catalog metadata
+// and encode JSON records to protobuf bytes without a companion Rust crate.
+// Lifecycle: `_from_uc_json` → `_descriptor_bytes` / `_encode_json` → `_free`.
+
+/// Holds a table's serialized `DescriptorProto` plus a prepared protobuf
```
teodordelibasic-db: The doc comment leaks Rust syntax into the generated header: `the concrete type is [\`ProtoSchema\`]`. C consumers can't follow that link. We can reword without the backtick-bracket form before regenerating.

rust/ffi/src/tests.rs
```diff
@@ -441,4 +441,504 @@ mod tests {
         assert_eq!(headers.len(), 1);
         assert!(headers.contains_key("Authorization"));
     }
+
+    // ========================================================================
+    // Dynamic protobuf schema tests
```
teodordelibasic-db, line 446: Can we add encode/decode round-trip tests for MAP, STRUCT, ARRAY<STRUCT>, DATE, and TIMESTAMP_NTZ, etc.

rust/ffi/src/lib.rs
```diff
@@ -1975,3 +1979,266 @@ pub extern "C" fn zerobus_get_default_config() -> CStreamConfigurationOptions {
...
+            write_error_result(result, &err, false);
+            ptr::null_mut()
+        }
+    }
+}
+
+/// Borrow the serialized descriptor bytes. Valid until `zerobus_proto_schema_free`.
+/// Pass directly to `zerobus_sdk_create_stream`. Returns NULL on null handle.
+#[no_mangle]
+pub extern "C" fn zerobus_proto_schema_descriptor_bytes(
```
teodordelibasic-db, line 2087: If `out_len` is null, we still return a non-null pointer with no way for the caller to learn the length. We can consider returning null (or requiring out_len) instead.

teodordelibasic-db (commented):
Mostly comments regarding comments and docs.

Also, preexisting, but none of the FFI functions wrap their bodies in `catch_unwind`. Worth a follow-up potentially to add panic guards crate-wide.

rust/ffi/src/lib.rs
zlata-stefanovic-db: You're right, the clone/drop is two atomics per call for no benefit, since validity rests on the handle's own reference anyway. (The Arc was originally added for a concurrency concern Irina raised, but as you noted in the sibling threads it never actually made free race-safe.) Removed the Arc machinery entirely and went back to a plain Box + a borrow helper (&*(schema as *const ProtoSchema) after the null check). Reworded the free/helper docs and the README to the real contract: concurrent readers are fine; free must be called exactly once after all in-flight calls return and must not race them.

rust/ffi/README.md
zlata-stefanovic-db: Addressed - removed the Arc and reworded to the real contract

rust/ffi/src/lib.rs
zlata-stefanovic-db: Addressed - removed the Arc and reworded to the real contract

rust/ffi/src/lib.rs
zlata-stefanovic-db: Addressed - removed the Arc and reworded to the real contract

rust/ffi/src/lib.rs
zlata-stefanovic-db: Addressed - removed the Arc and reworded to the real contract

rust/ffi/src/lib.rs
zlata-stefanovic-db: Both fixed - removed the [ProtoSchema] rustdoc link before regenerating zerobus.h, and descriptor_bytes now returns NULL when out_len is null rather than handing back an unsizable pointer.

rust/ffi/src/tests.rs
zlata-stefanovic-db: Added

rust/ffi/NEXT_CHANGELOG.md
zlata-stefanovic-db: Agreed the claim was wrong - fixed the docs (changelog/README/header) to say top-level non-nullable scalar and struct columns, and called out that ARRAY/MAP are repeated and nested struct fields aren't checked. I went with narrowing rather than extending the check, because enforcing presence on a non-nullable ARRAY/MAP isn't really well-defined: repeated has no presence on the wire, and an empty array is itself a valid non-null value - distinct from a null array - so "reject when omitted" would mean reading JSON key-presence and treating [] differently from absent, which is a semantics call (and a behavior change) I don't think belongs in this PR. Recursing into structs for nested required fields is doable but is similarly a scope/behavior expansion.

rust/ffi/src/lib.rs
zlata-stefanovic-db: Done, requiring out_len now

zlata-stefanovic-db:
> Mostly comments regarding comments and docs.
> 
> Also, preexisting, but none of the FFI functions wrap their bodies in `catch_unwind`. Worth a follow-up potentially to add panic guards crate-wide.

Agreed. Since it's preexisting and applies to every FFI entry point (not just the new proto functions), I'll open a follow-up (new ticket) to wrap all extern "C" bodies in a catch_unwind panic guard crate-wide so a panic converts to a CResult error instead of unwinding into C. Tracking separately from this PR. @teodordelibasic-db

rust/ffi/src/lib.rs
```diff
@@ -1975,3 +1979,272 @@ pub extern "C" fn zerobus_get_default_config() -> CStreamConfigurationOptions {
...
+            false,
+        );
+        return false;
+    }
+
+    // Top-level non-nullable scalar/struct columns are proto2 `required`, but
+    // prost-reflect doesn't enforce presence on encode — reject a missing one
+    // here rather than emit wire bytes the server rejects. (ARRAY/MAP are
+    // `repeated`, which has no presence, and nested struct fields aren't walked.)
+    let missing: Vec<String> = schema_ref
```
teodordelibasic-db, line 2195: The required-field check only walks top-level fields, so a record that omits a non-nullable field nested inside a STRUCT still encodes successfully here and is only rejected later by the server.

Since we have documented this, we can leave it as a follow up PR as well to add a recursive presence walk over nested messages, repeated message elements, and map values. Could you file an issue please 🙏

teodordelibasic-db (approved):
LGTM, like discussed, we can leave a few things as follow ups.

rust/ffi/src/lib.rs
zlata-stefanovic-db: https://github.com/databricks/zerobus-sdk/issues/393

PR #378, opened 2026-06-17
[Rust] Introduce per-table minted OAuth token cache

teodordelibasic-db:

This adds a per-table OAuth token cache to the Rust SDK so the built-in OAuth path stops minting a fresh Unity Catalog token on every stream creation.

What changes:

- New internal `token_cache` module. `TokenCache` lives on the `ZerobusSdk` instance and caches one token per `(client_id, sha256(client_secret), table_name)`, served until it nears expiry and refreshed lazily on access. Two builder knobs: `token_cache_enabled` (default on) and `token_refresh_buffer` (default 5 min).
- `DefaultTokenFactory` now reads the OAuth `expires_in` (previously discarded) via an internal `fetch_token`; the public `get_token` is unchanged and delegates to it.
- `OAuthHeadersProvider` is wired to the shared cache via an internal `with_cache(...)`; the public `new(...)` is unchanged (it now caches for its own lifetime).
- Cache invalidation on server-side auth rejection: a new `HeadersProvider::invalidate` (default no-op) is called when the server returns `Unauthenticated`/`PermissionDenied`, on both the gRPC and Arrow Flight paths, at initial creation, recovery, and mid-stream. The OAuth provider drops its cached token so the next attempt re-mints.
- Observability: an info `minted UC OAuth token` log (with `reason`, `expires_in_secs`, `elapsed_ms`), debug for cache hits, and warn for the degraded paths. Only the table name is logged, never the token or secret.
- Two bug fixes surfaced in the OAuth/stream-setup path (both pre-existing): the raw authorization token is no longer written to an error log or error message; and Arrow `FlightError → tonic::Status` now uses `.into()` (which preserves the inner gRPC code) instead of `Status::from_error` (which flattened it to `Unknown`, breaking retry and auth-rejection classification on the Arrow path).

Edge cases handled:

- Concurrent stream creations to the same table mint a single token (per-key single-flight lock), not one per caller.
- Refresh failure: a retryable error falls back to a still-valid cached token; a non-retryable error (bad/revoked credentials) propagates immediately rather than being masked.
- A response with no usable `expires_in` is never cached.
- A token that cannot be encoded as an `authorization` header value is rejected at mint, so it never enters the cache and can't poison it until expiry.
- An absurd `expires_in` or a caller-supplied `Duration::MAX` refresh buffer cannot panic (checked time arithmetic; overflow is treated as non-cacheable / always-refresh).
- A server auth rejection at any stream lifecycle point invalidates the cached token so the next attempt re-mints.

Why:

Service principal credentials are exchanged for a Unity Catalog token valid for ~1 hour, but a single stream lives at most ~15 minutes, and the SDK minted a new token on every stream connect and recovery. Clients that churn through many short-lived streams therefore hammered the UC token endpoint. Caching collapses that to roughly one mint per table per token lifetime.

Decisions not visible from the code:

- Keyed by table because the minted token is downscoped to a single `catalog.schema.table`; a token for one table is not valid for another.
- Scoped per `ZerobusSdk` instance (not process-global) to avoid global state; callers must reuse one SDK to benefit.
- The secret is keyed by its SHA-256 digest, not plaintext, so the raw secret never lives in the cache map while staying collision-free.
- The cache reuses a token until it nears expiry, so a permission change is re-checked at most once per token lifetime rather than per stream (the old per-stream minting re-checked every time as a side effect). A server-side rejection still invalidates immediately; UC also enforces grants at mint time, so a re-mint after a revocation is denied.

## How is this tested?

- Unit tests in `token_cache.rs`, `errors.rs`, `default_token_factory.rs`, and `builder/sdk_builder.rs`: cache reuse, per-table separation, secret rotation, no-TTL no-cache, disabled mode, single-flight under concurrency, retryable vs non-retryable refresh failure, no-TTL non-eviction, invalidate-forces-remint, header-validity check, `expires_in` parsing, `is_auth_rejection` classification, and the `FlightError → Status` code-preservation invariant. `make check` and the full suite pass under both the default and `arrow-flight` feature sets.
- Manual end-to-end on staging (e2-dogfood, `shinkansen.default.air_quality_managed`): five streams to one table from one SDK mint one token with the cache on (four cache hits) versus five with it off; the first stream is ~2.1 s (pays the ~0.8 s token round-trip) and cached ones ~0.6 s. Logs were inspected to confirm no token/secret appears at any level.
- [WIP] Testing edge cases around granting then revoking permissions to a service principal, as well as dropping and creating a table with the same name.

### DIFF
--- a/rust/sdk/src/lib.rs
+++ b/rust/sdk/src/lib.rs
@@ -55,6 +55,7 @@ pub mod schema;
 mod stream_configuration;
 pub mod stream_options;
 mod tls_config;
+mod token_cache;
 
 use std::collections::HashMap;
 use std::fmt::Debug;
@@ -275,6 +276,9 @@ pub struct ZerobusSdk {
     /// Final value sent as the HTTP `user-agent` header on every request.
     /// Either `"zerobus-sdk-rs/<version>"` or `"zerobus-sdk-rs/<version> <application_name>"`.
     pub(crate) sdk_identifier: Arc<str>,
+    /// Shared cache of OAuth tokens, keyed per table, reused across all streams
+    /// created from this SDK instance via the default OAuth path.
+    pub(crate) token_cache: Arc<token_cache::TokenCache>,
 }
 
 impl ZerobusSdk {
@@ -336,6 +340,7 @@ impl ZerobusSdk {
     /// fully-resolved value sent as the HTTP `user-agent` header; the builder
     /// is responsible for composing it from the default prefix and any
     /// caller-supplied `application_name` or override.
+    #[allow(clippy::too_many_arguments)]
     pub(crate) fn new_with_config(
         zerobus_endpoint: String,
         unity_catalog_url: String,
@@ -343,6 +348,8 @@ impl ZerobusSdk {
         tls_config: Arc<dyn TlsConfig>,
         connector_factory: Option<ConnectorFactory>,
         sdk_identifier: Arc<str>,
+        token_cache_enabled: bool,
+        token_refresh_buffer: Duration,
     ) -> Self {
         ZerobusSdk {
             zerobus_endpoint,
@@ -352,6 +359,10 @@ impl ZerobusSdk {
             tls_config,
             connector_factory,
             sdk_identifier,
+            token_cache: Arc::new(token_cache::TokenCache::new(
+                token_cache_enabled,
+                token_refresh_buffer,
+            )),
         }
     }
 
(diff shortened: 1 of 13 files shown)


rust/sdk/src/lib.rs
```diff
@@ -822,6 +833,11 @@ impl ZerobusStream {
                 let _ = server_error_tx.send(Some(error.clone()));
                 if !error.is_retryable() || !options.recovery {
                     is_closed.store(true, Ordering::Relaxed);
+                    // A mid-stream auth rejection means the cached token is no
+                    // longer accepted; drop it so the next stream re-mints.
+                    if error.is_auth_rejection() {
```
flaviofcruz, line 838: Are auth rejections non retryable errors? If they are retryable we could still hammer the endpoint.

flaviofcruz (approved):

nikolaobradovic-db (approved):

rust/sdk/src/lib.rs
teodordelibasic-db: Checked and they are non-retriable, but your question uncovered a potential day 0 Arrow issue where we mask reconnect errors as unavailable which are retriable. I will create a follow up PR.

PR #387, opened 2026-06-18
Make SDK ingestion docs agent-friendly

elenagaljak-db:
## What

Reworks the ingestion documentation across **all five SDKs** (Rust, Python, TypeScript, Java, Go) plus the root docs so that clients — including AI coding assistants — write performant code by default.

## Why

Users (and agents) were producing slow clients that call the wait-for-acknowledgment method (`wait_for_offset` / `waitForOffset` / `WaitForOffset`, or `.join()` on a per-record future) after **every** ingest. Because ingestion is asynchronous and pipelined, waiting per record forces a full server round-trip before the next record is sent, limiting throughput to ~one record per round-trip. The docs were partly to blame: every SDK led with the per-record-wait example, and the high-throughput pattern was buried far below.

## Changes

- **Lead with the idiomatic flow** — ingest in a loop, then `flush()` once — as the first example in every README. Per-record waiting is presented as a legitimate tool for low-volume "confirm this specific record" cases, not the default.
- **Positive framing**: a new "Acknowledgments and throughput" section explains how acks work, positioned *after* the API descriptions (JSON / Protobuf / Arrow) rather than wedged before them.
- **API doc comments** updated where consumers actually see them: Rust `///`, Python docstrings (sync + async), TypeScript JSDoc, Java Javadoc, Go godoc — on `ingest_*`, `wait_for_offset`/`flush` and equivalents.
- **Examples**: single-record examples now ingest-then-`flush()` once.
- **Java QuickStart** rewritten off the deprecated `ingestRecord().join()`-in-a-loop onto `ingestRecordOffset()` + wait-once.
- **`CLAUDE.md`** (root + per-SDK) gain a "client code patterns" / performance rule for in-repo agents.
- **`NEXT_CHANGELOG.md`** updated in each SDK under Documentation.

## Notes

- The "wait on the last offset confirms all prior" guidance is verified against the Rust core: the ack watermark is monotonic (`wait_for_offset_internal` waits for `last_received_offset >= target`).
- Ack callbacks exist only in Rust/Python/Java; the TS and Go docs use the `flush()`/last-offset approach (no invented APIs).
- Verified: Rust doctests pass, Rust examples compile, Go is gofmt-clean, Python compiles. Docs/prose only — no library logic changed.


go/README.md
```diff
@@ -898,16 +929,17 @@ The test suite includes:
...
-4. **Tune Inflight Limits** - Adjust `MaxInflightRequests` based on memory and throughput needs
-5. **Enable Recovery** - Always set `Recovery: true` in production environments
-6. **Use Batch Ingestion** - For high throughput, ingest many records before calling `Flush()`
-7. **Monitor Errors** - Log and alert on non-retryable errors
-8. **Use Protocol Buffers for Production** - More efficient than JSON for high-volume scenarios
+5. **Tune Inflight Limits** - Adjust `MaxInflightRequests` based on memory and throughput needs
+6. **Enable Recovery** - Always set `Recovery: true` in production environments
+7. **Use Batch Ingestion** - For high throughput, ingest many records before calling `Flush()`
+8. **Monitor Errors** - Log and alert on non-retryable errors
+9. **Use Protocol Buffers for Production** - More efficient than JSON for high-volume scenarios
```
teodordelibasic-db, line 944: Duplicate number 9 in list below, this line should be followed with 10/11/12.

go/README.md
```diff
@@ -502,28 +530,26 @@ if err != nil {
     log.Fatal(err)
 }
 log.Printf("Batch queued with offset: %d", batchOffset)
+// ... ingest more batches ...
+stream.Flush() // wait for everything at the end
```
teodordelibasic-db: Other `Flush()` and `WaitForOffset()` calls in this README check for errors; this lone one does not. We can change to something like:

```go
if err := stream.Flush(); err != nil {
    log.Fatal(err)
}
```

python/README.md
```diff
@@ -352,12 +368,36 @@ for batch in unacked_batches:
...
 |--------|------------|----------|
 | `ingest_record_nowait()` | **Highest** | Fire-and-forget: no offset returned; maximum throughput when you do not need per-record ack tracking in the hot path |
-| `ingest_record_offset()` | Medium | Recommended for most apps: returns an offset after queueing; call `wait_for_offset()` when you need durability confirmation |
+| `ingest_record_offset()` | Medium | Recommended for most apps: returns an offset after queueing. Ingest in a loop, then `flush()` once |
 | `ingest_record()` | Low | **Deprecated** — prefer offset-based APIs |
 
+**Idiomatic flow:**
+
+```python
+for record in records:
```
teodordelibasic-db, line 388: The Quick Start above this is async (`await ... ; asyncio.run(main())`), so maybe we should change this section to use async as well.

```python
for record in records:
    await stream.ingest_record_offset(record)
await stream.flush()
```

python/README.md
```diff
@@ -352,12 +368,36 @@ for batch in unacked_batches:
...
+    stream.ingest_record_offset(record)   # queues immediately, no round-trip
+stream.flush()                            # one wait for everything
+```
+
+**Confirming a specific record** (waiting on the last offset confirms all prior records):
+
+```python
+for record in records:
```
teodordelibasic-db, line 396: ditto:

```python
for record in records:
    offset = await stream.ingest_record_offset(record)
await stream.wait_for_offset(offset)            # confirm the run before continuing
```

typescript/src/lib.rs
```diff
@@ -412,11 +419,14 @@ impl ZerobusStream {
     /// # Example
     ///
     /// ```typescript
-    /// // Promise resolves immediately with offset (before server ack)
-    /// const offset1 = await stream.ingestRecordOffset(record1);
-    /// const offset2 = await stream.ingestRecordOffset(record2);
-    /// // Wait for both to be acknowledged
-    /// await stream.waitForOffset(offset2);
+    /// // High-throughput pattern: ingest in a loop, wait once at the end.
+    /// let lastOffset;
```
teodordelibasic-db: If `records` is empty, `lastOffset` is `undefined` and `waitForOffset(undefined)` runs instead of being a
no-op. Inconsistent with the neighboring `ingestRecordsOffset()` example (line 486), which correctly
guards with a `null` sentinel. Either prefer `await stream.flush();` here, or guard:

```typescript
let lastOffset: bigint | null = null;
for (const record of records) lastOffset = await stream.ingestRecordOffset(record);
if (lastOffset !== null) await stream.waitForOffset(lastOffset);
```

typescript/src/lib.rs
```diff
@@ -527,12 +549,12 @@ impl ZerobusStream {
     /// # Example
     ///
     /// ```typescript
-    /// const offsets = [];
+    /// let lastOffset;
```
teodordelibasic-db: Same as [typescript/src/lib.rs:423](https://github.com/databricks/zerobus-sdk/pull/387/changes#r3451090360).

rust/README.md
```diff
@@ -530,27 +538,21 @@ let records: Vec<ProtoMessage<YourMessage>> = vec![
     ProtoMessage(YourMessage { id: Some(3), /* ... */ }),
 ];
 
-// Returns Some(offset) for non-empty batches, None for empty batches
-if let Some(offset) = stream.ingest_records_offset(records).await? {
-    stream.wait_for_offset(offset).await?;
-}
+// Returns Some(offset) for non-empty batches, None for empty batches.
+// Queue many batches this way; flush() once when done.
```
teodordelibasic-db, line 544: Since the comment says "flush() once when done" we can add a trailing `stream.flush().await?;` so the fragment is self-contained.

typescript/README.md
```diff
@@ -875,13 +877,14 @@ Represents an active ingestion stream.
 async ingestRecordOffset(payload: Buffer | string | object): Promise<bigint>
 ```
 
-**(Recommended)** Ingests a single record. The Promise resolves immediately after the record is queued (before server acknowledgment). Use `waitForOffset()` to wait for acknowledgment when needed.
+**(Recommended)** Ingests a single record. The Promise resolves immediately after the record is queued (before server acknowledgment); the round-trip happens in the background. The idiomatic flow is to ingest in a loop and then `flush()` once to confirm everything queued so far. The returned offset, together with `waitForOffset()`, lets you confirm a specific record when needed — prefer that for bulk over waiting after each record, since per-record waiting limits throughput to one round-trip per record.
 
 ```typescript
-// High-throughput pattern: send many, wait once
+// Idiomatic flow: ingest in a loop, then flush once
 const offset1 = await stream.ingestRecordOffset(record1);  // Resolves immediately
```
teodordelibasic-db: `offset1` is dead. We can use a small loop with a single `lastOffset` (matching the JSDoc in `src/lib.rs`) or
drop the first binding.

go/CLAUDE.md
```diff
@@ -2,6 +2,24 @@
 
 Go wrapper around the Rust core via cgo and the C FFI library.
 
+## Client code patterns (performance)
+
+When writing or reviewing client/example code, follow the idiomatic async flow.
+`IngestRecordOffset()` (and `IngestRecordsOffset()` / `IngestBatch()`) return as
```
teodordelibasic-db: `IngestBatch` is a method on `ZerobusArrowStream` (`go/arrow_stream.go:168`), not on `ZerobusStream` where `IngestRecordOffset`/`IngestRecordsOffset` live. This reads as if all three are on the same type. We should maybe drop `IngestBatch()` from the list or note it belongs to the Arrow stream. Flagged by LLM so probably useful.

typescript/README.md
```diff
@@ -1115,6 +1118,7 @@ enum RecordType {
 5. **Use Protocol Buffers for production**: Protocol Buffers (the default) provides better performance and schema validation. Use JSON only when you need schema flexibility or for quick prototyping.
 6. **Store credentials securely**: Use environment variables, never hardcode credentials
 7. **Use batch ingestion**: For high-throughput scenarios, use `ingestRecordsOffset()` instead of individual `ingestRecordOffset()` calls
+8. **Ingest in a loop, then `flush()`**: `ingestRecordOffset()` / `ingestRecordsOffset()` resolve as soon as the record is queued; the SDK sends and tracks acknowledgment in the background. Confirm durability with a single `flush()` (once for a bounded batch, or periodically for a long-running stream). Each ingest returns an offset, and `waitForOffset(offset)` confirms a specific record when you need it (acks are ordered, so the last offset confirms the whole run). Just avoid calling `waitForOffset()` after every record in a tight loop, since that limits throughput to one record per round-trip.
```
teodordelibasic-db: This repeats the same explanation as the "Acknowledgments and throughput" blockquote at line 185. We can shorten #8 to a cross-reference ("See Acknowledgments and throughput above"). Separately, other SDKs give this its own `### Acknowledgments and throughput` heading (e.g. python/README.md), whereas TS uses a long `>` blockquote - promoting it to a heading would match the rest of the set.

teodordelibasic-db (commented):
Public doc examples in `record_types.rs` and `arrow_stream.rs` still ends an `ingest_*` call with an immediate `stream.wait_for_offset(offset).await?;`.

Per location, the line to change and the fix:

- `rust/sdk/src/record_types.rs:59` (`ProtoBytes` example) — we can drop the trailing
`stream.wait_for_offset(offset).await?;` in favor of `stream.flush().await?;`, or add a one-line note that per-record waiting is for low-volume confirmation only.
- `rust/sdk/src/record_types.rs:83` (`JsonString` example) — same change.
- `rust/sdk/src/record_types.rs:108` (`ProtoMessage` example) — same change.
- `rust/sdk/src/record_types.rs:141` (`JsonValue` example) — same change.
- `rust/sdk/src/arrow_stream.rs:190` (`ZerobusArrowStream` type-level example) — the example ingests one batch then waits on its offset; lead with the loop-then `flush()` pattern instead, matching `lib.rs`, and keep the single-batch wait only as the labeled low-volume case.
- `rust/sdk/src/arrow_stream.rs:1221` (`ingest_batch` method example) — same: replace the immediate
`wait_for_offset(offset)` with `flush()` once at the end, or label it as the confirm-a specific-batch case.

danilonajkov-db:
A thing that I have noticed in READMEs is that a JSON example is always given first. We should put proto first and for json add some disclamer about throughput.

elenagaljak-db:
I don't agree that it should be first since JSON is easier to set up to just test and its always labeled as quick start. We can emphasize more that its better to use proto/arrow for production.

teodordelibasic-db:
> I don't agree that it should be first since JSON is easier to set up to just test and its always labeled as quick start. We can emphasize more that its better to use proto/arrow for production.

Ideally once we add support for dynamic protobuf etc. it should be as easy to setup as JSON.

go/README.md
elenagaljak-db: Fixed, thanks

python/README.md
elenagaljak-db: Fixed

python/README.md
elenagaljak-db: Fixed

rust/README.md
elenagaljak-db: Added

python/README.md
```diff
@@ -12,8 +12,9 @@ A high-performance Python client for streaming data ingestion into Databricks De
 - [Features](#features)
 - [Installation](#installation)
 - [Quick Start](#quick-start)
-  - [JSON (Simplest)](#option-1-json-simplest)
-  - [Protocol Buffers](#option-2-protocol-buffers)
+  - [Protocol Buffers (Recommended)](#option-1-protocol-buffers-recommended)
```
teodordelibasic-db: I think this breaks the TOC links because headings were not reordered/renamed:

- TOC: `[Protocol Buffers (Recommended)](#option-1-protocol-buffers-recommended)` and `[JSON (Simplest)](#option-2-json-simplest)`.
- Headings: `### Option 1: JSON (Simplest)` (line 100, anchor `#option-1-json-simplest`) and `### Option 2: Protocol Buffers` (line 173, anchor `#option-2-protocol-buffers`).

Both TOC links point to anchors that don't exist.

teodordelibasic-db (commented):
LGTM, could you please just check whether TOC is broken.

python/README.md
elenagaljak-db: Yep, you were right. I also fixed it for java, go and rust because they had some stale lines too

teodordelibasic-db (approved):

PR #390, opened 2026-06-20
fix: add exponential recovery backoff

### LINKED ISSUE #192: [Rust (affects all language bindings)] Use exponential backoff with jitter for stream recovery retry strategy
### SDK

Rust (affects all language bindings)

### Description

The current retry strategy for stream creation and recovery uses `tokio_retry::strategy::FixedInterval` (`lib.rs:990`, `arrow_stream.rs:509`). This means all retry attempts use the same fixed delay (default: 2 seconds), with no jitter.

Fixed-interval retries are problematic in distributed systems:

- **Thundering herd**: When a server failure disconnects many clients simultaneously, all clients retry at the same interval, concentrating load on the recovering server.
- **No backoff under sustained failures**: A fixed 2-second interval gives the server no additional recovery time as failures persist.
- Exponential backoff with jitter is the industry standard recommended by AWS, GCP, and gRPC best practices.

### Proposed Solution

1. Add a `RetryStrategy` enum to `StreamConfiguration`:
   ```rust
   pub enum RetryStrategy {
       Fixed,
       ExponentialBackoffWithJitter,
   }
   ```
2. Add `max_recovery_backoff_ms` field (cap for exponential growth, default: 30,000 ms).
3. Change the default strategy from `Fixed` to `ExponentialBackoffWithJitter`.
4. `recovery_backoff_ms` serves as the initial backoff for exponential, or the interval for fixed.
5. No new dependencies required — `tokio-retry` already provides `ExponentialBackoff` and `jitter`.

Changes are needed in 2 locations in the Rust core (`lib.rs` and `arrow_stream.rs`), plus configuration structs. All language bindings inherit the behavior through their respective FFI/binding layers.

### Additional Context

References:
- [Exponential Backoff And Jitter (AWS Architecture Blog)](https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/)
- [gRPC Retry Design](https://github.com/grpc/proposal/blob/master/A6-client-retries.md)
- `tokio-retry` supports `ExponentialBackoff` and `jitter` out of the box

Current code (`lib.rs:990-991`):
```rust
let strategy = FixedInterval::from_millis(options.recovery_backoff_ms)
    .take(options.recovery_retries as usize);
```

Proposed equivalent:
```rust
use tokio_retry::strategy::{ExponentialBackoff, jitter};
let strategy = ExponentialBackoff::from_millis(options.recovery_backoff_ms)
    .max_delay(Duration::from_millis(options.max_recovery_backoff_ms))
    .map(jitter)
    .take(options.recovery_retries as usize);
```

nanookclaw:
## Summary

Closes #192.

This changes Rust stream creation and recovery retries from a fixed interval default to exponential backoff with full jitter. `recovery_backoff_ms` remains the fixed interval when `RetryStrategy::Fixed` is selected, and becomes the base delay for `RetryStrategy::ExponentialBackoffWithJitter`; `max_recovery_backoff_ms` caps the computed exponential delay before jitter.

The same retry helper is used by gRPC stream creation, Arrow Flight stream creation, and Arrow supervisor recovery sleeps so the behavior stays consistent across both stream implementations. The default strategy is now exponential backoff with jitter, so wrapper SDKs that rely on the Rust core inherit the safer retry behavior without new dependencies.

## Verification

- `cargo fmt --manifest-path rust/sdk/Cargo.toml`
- `cargo check -p databricks-zerobus-ingest-sdk`
- `cargo check -p databricks-zerobus-ingest-sdk --features arrow-flight`
- `cargo test -p databricks-zerobus-ingest-sdk --lib`
- `cargo test -p databricks-zerobus-ingest-sdk --features arrow-flight --lib`
- `git diff --check`

### DIFF
--- a/rust/sdk/src/arrow_configuration.rs
+++ b/rust/sdk/src/arrow_configuration.rs
@@ -3,7 +3,7 @@
 //! **Beta**: Arrow Flight ingestion is in Beta. The API is stabilising but may
 //! still change before reaching GA.
 
-use crate::stream_options::defaults;
+use crate::stream_options::{defaults, RetryStrategy};
 use arrow_ipc::CompressionType;
 
 /// Configuration options for Arrow Flight stream creation and operation.
@@ -53,10 +53,24 @@ pub struct ArrowStreamConfigurationOptions {
 
     /// Backoff time in milliseconds between stream recovery retry attempts.
     ///
-    /// The SDK will wait this duration before attempting another recovery after a failure.
+    /// For fixed retry strategy, the SDK will wait this duration before each retry.
+    /// For exponential retry strategy, this is the base delay before jitter is applied;
+    /// retries use full jitter in the range `0..=computed_delay`.
     ///
     /// Default: 2,000 (2 seconds)
     pub recovery_backoff_ms: u64,
+
+    /// Maximum recovery backoff time in milliseconds.
+    ///
+    /// This caps the exponential retry strategy. It is ignored by the fixed strategy.
+    ///
+    /// Default: 30,000 (30 seconds)
+    pub max_recovery_backoff_ms: u64,
+
+    /// Retry strategy for stream creation and recovery.
+    ///
+    /// Default: `RetryStrategy::ExponentialBackoffWithJitter`
+    pub retry_strategy: RetryStrategy,
 
     /// Maximum number of recovery retry attempts before giving up.
     ///
@@ -122,6 +136,8 @@ impl Default for ArrowStreamConfigurationOptions {
             recovery: defaults::RECOVERY,
             recovery_timeout_ms: defaults::RECOVERY_TIMEOUT_MS,
             recovery_backoff_ms: defaults::RECOVERY_BACKOFF_MS,
+            max_recovery_backoff_ms: defaults::MAX_RECOVERY_BACKOFF_MS,
+            retry_strategy: RetryStrategy::default(),
             recovery_retries: defaults::RECOVERY_RETRIES,
             server_lack_of_ack_timeout_ms: defaults::SERVER_LACK_OF_ACK_TIMEOUT_MS,
             flush_timeout_ms: defaults::FLUSH_TIMEOUT_MS,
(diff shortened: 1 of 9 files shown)


rust/sdk/src/arrow_configuration.rs
```diff
@@ -131,3 +147,22 @@ impl Default for ArrowStreamConfigurationOptions {
...
+        assert_eq!(
+            options.retry_strategy,
+            RetryStrategy::ExponentialBackoffWithJitter
+        );
+        assert_eq!(
+            options.max_recovery_backoff_ms,
+            defaults::MAX_RECOVERY_BACKOFF_MS
+        );
+    }
+}
```
elenagaljak-db, line 168: I'm not sure we need these types of tests, we don't have them for the rest of the configs.

rust/sdk/src/stream_configuration.rs
```diff
@@ -176,3 +192,22 @@ impl Default for StreamConfigurationOptions {
...
+        assert_eq!(
+            options.retry_strategy,
+            RetryStrategy::ExponentialBackoffWithJitter
+        );
+        assert_eq!(
+            options.max_recovery_backoff_ms,
+            defaults::MAX_RECOVERY_BACKOFF_MS
+        );
+    }
+}
```
elenagaljak-db, line 213: Same as in arrow_configuration file.

rust/sdk/src/stream_options.rs
```diff
@@ -2,15 +2,65 @@
...
+/// Retry strategy for stream creation and recovery.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+#[non_exhaustive]
+pub enum RetryStrategy {
+    /// Retry after the configured `recovery_backoff_ms` interval.
+    Fixed,
+    /// Retry with exponential backoff and jitter, starting from `recovery_backoff_ms`.
+    #[default]
+    ExponentialBackoffWithJitter,
+}
```
elenagaljak-db, line 18: I think fixed should be kept as a default one until we add support for this in all SDKs because if we upgrade Rust version in other SDKs before exposing retry_strategy/max_recovery_backoff_ms in them, there would be no way for users to opt back to Fixed and it would be a silent behavioral change. 
Otherwise we would have to commit to add support for this when we upgrade other SDKs. What do you think @teodordelibasic-db?

rust/sdk/src/stream_options.rs
```diff
@@ -23,3 +73,51 @@ pub mod defaults {
...
+    #[test]
+    fn exponential_retry_strategy_applies_jitter_under_cap() {
+        let mut strategy =
+            recovery_retry_strategy(RetryStrategy::ExponentialBackoffWithJitter, 100, 250);
+
+        for _ in 0..10 {
+            assert!(strategy.next().unwrap() <= Duration::from_millis(250));
+        }
+    }
+}
```
elenagaljak-db, line 123: I would change these tests for tests against the mock server (in rust/test), for consistency.

PR #392, opened 2026-06-22
[Python] Expose application_name to customize user-agent

### LINKED ISSUE #382: [Python] Application name option for customizable user-agent
Part of #379.

Add an application name option on SDK construction so users can customize the identifier the SDK sends in the request header (user-agent), matching the Rust behavior (PR #268).

- [ ] Application name option on SDK construction
- [ ] Appended to the identifier sent in the request header (user-agent)
- [ ] Example, tests, changelog entry, docs

andrijast-db:

Accept an optional `application_name` argument on SDK initialization.  When set, it is appended to the HTTP `user-agent` header sent on every request, so callers can be identified in server-side telemetry. The SDK prefix is preserved, so the wire value becomes `"zerobus-sdk-py/<version> <application_name>"` (e.g. `"my-app/1.0"`).

Closes #382

## How is this tested?

Tested with existing tests. New tests are added as well.

### DIFF
--- a/python/NEXT_CHANGELOG.md
+++ b/python/NEXT_CHANGELOG.md
@@ -8,10 +8,13 @@
 
 - **`ZerobusSdk(application_name=...)`**: Both the sync and async `ZerobusSdk`
   constructors accept an optional `application_name` argument. When set, it is
-  appended to the HTTP `user-agent` header sent on every request, so callers
-  can be identified in server-side telemetry. The SDK prefix is preserved, so
-  the wire value becomes `zerobus-sdk-py/<version> <application_name>`. By
-  convention use `<product>/<version>` (e.g. `"my-app/1.0"`).
+  appended to the HTTP `user-agent` header on gRPC requests to the Zerobus
+  server (it is not sent on the requests to the login service that mint the
+  OAuth token), so callers can be identified in server-side telemetry. The SDK
+  prefix is preserved, so the wire value becomes
+  `zerobus-sdk-py/<version> <application_name>`. By convention use
+  `<product>/<version>` (e.g. `"my-app/1.0"`). The value is trimmed; blank
+  values are ignored and control characters are rejected.
 
 ### Bug Fixes
 
--- a/python/NEXT_CHANGELOG.md
+++ b/python/NEXT_CHANGELOG.md
@@ -6,9 +6,19 @@
 
 ### New Features and Improvements
 
+- **`ZerobusSdk(application_name=...)`**: Both the sync and async `ZerobusSdk`
+  constructors accept an optional `application_name` argument. When set, it is
+  appended to the HTTP `user-agent` header sent on every request, so callers
+  can be identified in server-side telemetry. The SDK prefix is preserved, so
+  the wire value becomes `zerobus-sdk-py/<version> <application_name>`. By
+  convention use `<product>/<version>` (e.g. `"my-app/1.0"`).
+
 ### Bug Fixes
 
 ### Documentation
+
+- README: documented the `application_name` constructor argument.
+- Examples: all examples under `examples/` now demonstrate `application_name`.
 
 ### Internal Changes
 
@@ -17,3 +27,7 @@
 ### Deprecations
 
 ### API Changes
+
+- `ZerobusSdk.__init__` gains an optional `application_name: Optional[str] = None`
+  parameter (sync and async). Strictly additive; existing two-argument callers
+  are unaffected.
--- a/python/NEXT_CHANGELOG.md
+++ b/python/NEXT_CHANGELOG.md
@@ -13,8 +13,7 @@
   OAuth token), so callers can be identified in server-side telemetry. The SDK
   prefix is preserved, so the wire value becomes
   `zerobus-sdk-py/<version> <application_name>`. By convention use
-  `<product>/<version>` (e.g. `"my-app/1.0"`). The value is trimmed; blank
-  values are ignored and control characters are rejected.
+  `<product>/<version>` (e.g. `"my-app/1.0"`).
 
 ### Bug Fixes
 
--- a/python/README.md
+++ b/python/README.md
@@ -368,7 +368,7 @@ Main entry point. Sync: `from zerobus.sdk.sync import ZerobusSdk` / Async: `from
 sdk = ZerobusSdk(server_endpoint: str, unity_catalog_endpoint: str, application_name: Optional[str] = None)
 ```
 
-`application_name` is optional; when set it is appended to the `user-agent` header on gRPC requests to the Zerobus server (not on the OAuth token requests to the login service). It follows the `"<product>/<version>"` convention (e.g. `my-app/1.0`). The value is trimmed; blank is ignored and control characters are rejected.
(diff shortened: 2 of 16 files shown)


python/tests/test_smoke.py
```diff
@@ -192,6 +192,42 @@ def test_record_type_equality(self):
         self.assertNotEqual(proto1, json_type)
 
 
+class TestApplicationName(unittest.TestCase):
```
teodordelibasic-db, line 195: This test asserts `assertIsNotNone(sdk)` - i.e. that the constructor doesn't raise.

Can we try to also verify that `application_name="my-app/1.0"` actually produces the wire value `zerobus-sdk-py/<version> my-app/1.0`?

Perhaps something like:
- Expose a read-only `sdk_identifier` accessor on the PyO3 binding (gated under a `testing` feature) for assertion in Python.
- Add a Rust-level test in `python/rust/` that constructs via PyO3 and inspects the inner SDK.

python/NEXT_CHANGELOG.md
```diff
@@ -6,14 +6,28 @@
 
 ### New Features and Improvements
 
+- **`ZerobusSdk(application_name=...)`**: Both the sync and async `ZerobusSdk`
```
teodordelibasic-db, line 9: Can we just slightly reword to say that this is appended on gRPC requests towards the Zerobus server, excluding the requests towards login service to mint the OAuth token.

Also applies to `python/README.md:371`.

python/rust/src/sync_wrapper.rs
```diff
@@ -299,12 +303,15 @@ impl ZerobusSdk {
...
 
-        let inner = RustSdk::builder()
+        let builder = RustSdk::builder()
             .endpoint(host)
             .unity_catalog_url(unity_catalog_url)
-            .sdk_identifier(sdk_identifier)
-            .build()
-            .map_err(map_error)?;
+            .sdk_identifier(sdk_identifier);
+        let builder = match application_name {
```
teodordelibasic-db, line 311: It would maybe be beneficial to validate the value that the client passed for some basic cases:

1. Trim and treat blank-after-trim as `None` (consistent with empty-string handling).
2. Reject `\r`/`\n`/control characters at the wrapper layer with a `ZerobusError::InvalidArgument` that names the parameter.
3. Document the expected format constraint in the docstring or README.

Same goes for async.

python/rust/src/sync_wrapper.rs
```diff
@@ -283,7 +283,11 @@ pub struct ZerobusSdk {
 #[pymethods]
 impl ZerobusSdk {
     #[new]
-    fn new(host: String, unity_catalog_url: String) -> PyResult<Self> {
+    fn new(
```
teodordelibasic-db, line 287: Every other PyO3 method/constructor with an `Option` parameter in this crate uses an explicit `#[pyo3(signature = ...)]`:
- `common.rs:76` — `TableProperties::new`
- `auth.rs:31` — `HeadersProvider`
- `arrow.rs:186` — Arrow options
- All `create_*_stream` methods in `sync_wrapper.rs` (lines 397, 420) and `async_wrapper.rs` (lines 303, 330, 357, 379)

The new `#[new]` for `ZerobusSdk` is the only exception. It works today because PyO3 0.20 implicitly gives trailing `Option<T>` a `None` default, but:
- It breaks the crate's own convention.
- PyO3 0.22+ removes this implicit-default behavior, so a future PyO3 bump turns this into a hard breakage here while annotated methods elsewhere keep working.

Suggested change in both files:
```rust
#[new]
#[pyo3(signature = (host, unity_catalog_url, application_name = None))]
fn new(
    host: String,
    unity_catalog_url: String,
    application_name: Option<String>,
) -> PyResult<Self> {
```

python/zerobus/sdk/sync/zerobus_sdk.py
```diff
@@ -197,8 +197,8 @@ def get_unacked_batches(self) -> list:
 class ZerobusSdk:
     """Python wrapper around Rust ZerobusSdk that provides unified create_stream API."""
 
-    def __init__(self, host: str, unity_catalog_url: str):
-        self._inner = _RustZerobusSdk(host, unity_catalog_url)
+    def __init__(self, host: str, unity_catalog_url: str, application_name: Optional[str] = None):
```
teodordelibasic-db, line 201: Preceding this change, but let's add a docstring for this method, something along the lines of:

```python
def __init__(self, host: str, unity_catalog_url: str, application_name: Optional[str] = None):
    """
    Args:
        host: Zerobus endpoint URL (e.g., "https://<workspace>.zerobus.<region>.cloud.databricks.com").
        unity_catalog_url: Unity Catalog URL for OAuth.
        application_name: Optional caller identifier (conventionally "<product>/<version>")
            appended to the HTTP user-agent on Zerobus gRPC requests. The SDK prefix
            "zerobus-sdk-py/<version>" is always preserved; the wire value becomes
            "zerobus-sdk-py/<version> <application_name>".
    """
    self._inner = _RustZerobusSdk(host, unity_catalog_url, application_name)
```

(Adjust the inner type for the async wrapper.)

python/README.md
teodordelibasic-db, line 110: All four Quick Start examples (lines 110, 138, 212, 238) still construct with the bare two-arg form `ZerobusSdk(server_endpoint, workspace_url)`.

We can add a commented `application_name="my-app/1.0"` line to at least one sync and one async Quick Start so new users can discover the feature without scrolling to the API Reference.

python/zerobus/sdk/sync/zerobus_sdk.py
teodordelibasic-db, line 11: We can also add the argument here, the async file, and in python/zerobus/__init__.py

teodordelibasic-db (commented):
Looks great, mostly nits.

python/README.md
andrijast-db: Which ones of those four to add comment to tho?

I'd leave it out, it's not really relevant to Quick Start.

python/rust/src/common.rs
```diff
@@ -496,6 +496,31 @@ pub(crate) fn apply_grpc_options<'a>(
     Ok(b)
 }
 
+/// Validate and normalize a caller-supplied `application_name`.
+///
+/// Trims surrounding whitespace and treats a blank result as unset (`None`),
+/// consistent with how the core builder ignores an empty identifier. Rejects
+/// values containing control characters (e.g. `\r`, `\n`) with an
+/// `InvalidArgument` error naming the parameter.
+pub(crate) fn normalize_application_name(
```
teodordelibasic-db: Like discussed offline, we should actually do this in Rust so the logic is centralized and we don't duplicate it in wrappers.

python/zerobus/__init__.py
teodordelibasic-db: Let's update this one as well.

teodordelibasic-db (commented):
LGTM

python/rust/src/common.rs
andrijast-db: Until Rust core is updated, invalid app names will get rejected instead of normalized

teodordelibasic-db (approved):

PR #394, opened 2026-06-22
[Rust] Add stream recovery lifecycle logging

elenagaljak-db:
## What

Adds informative logging around the stream recovery lifecycle in the Rust SDK, which was previously largely silent between a stream failure and a successful recovery.

New logs:
- **`info`** — when recovery starts, including the number of pending records (`pending_records`).
- **`info`** — when a recovered stream re-sends unacknowledged records, including how many (`resent_records`). `LandingZone::reset_observe()` now returns the count moved back from observed to unobserved (additive change to a private module).
- **`warn`** — each failed stream-creation attempt, with attempt number, max attempts, retryability, and backoff.
- **`error`** — on a non-retryable failure, how many records were left unacknowledged. These records are **retained** (not dropped) and remain retrievable via `get_unacked_records` / `get_unacked_batches`.

## Why

Recovery storms and silent re-sends (which can produce duplicate rows downstream) were hard to diagnose from logs alone. These additions make the recovery path observable without adding noise to the hot path — they only fire on rare lifecycle events.

## Notes

- No public API changes: `landing_zone` is a private module.
- `NEXT_CHANGELOG.md` updated under New Features and Improvements.
- `cargo build`, `cargo clippy`, and the `landing_zone` tests all pass.


rust/sdk/src/lib.rs
```diff
@@ -676,6 +676,13 @@ impl ZerobusStream {
                 return Ok(());
             }
 
+            if !initial_stream_creation {
+                info!(
+                    pending_records = landing_zone.len(),
```
teodordelibasic-db: `pending_records = landing_zone.len()` and `len()` is `queue.len() + observed_items.len()`,
where each item is `Box<IngestRequest>` wrapping an `EncodedBatch`. So one
`ingest_records_offset()` call carrying 10,000 rows logs as `1`. The field name says "records".

We can either rename to `pending_batches`, or sum `record.payload.get_record_count()` for a true
row count (`get_record_count()` is already used at lib.rs:1157/1204).

rust/sdk/src/lib.rs
```diff
@@ -744,7 +766,14 @@ impl ZerobusStream {
             }
 
             // 2. Reset landing zone.
-            landing_zone_recovery.reset_observe();
+            let resent_records = landing_zone_recovery.reset_observe();
```
teodordelibasic-db: Start logs `landing_zone.len()` = queue + observed. Here for resend we only log
`reset_observe()` = observed items moved back. A trace of `pending_records=10` then
`resent_records=2` reads as "8 records vanished" when 8 may simply still be queued.

We can: log both counts on the resend line, or rename so the relationship is clear
(e.g. `observed_resent` vs `total_pending`). Same batch-vs-record naming issue as above applies.

rust/sdk/src/lib.rs
teodordelibasic-db, line 751: On this path the code emits only per-attempt warns, then (for non-initial) calls
`fail_all_pending_records`. There is no single definitive "recovery failed" error like Arrow's
"Supervisor: Max recovery retries exceeded" (arrow_stream.rs:660). If the landing zone is empty,
`fail_all_pending_records` logs nothing, so a terminal recovery failure can produce zero
error-level lines. We can maybe add one `error!` before `return Err(e)` with `attempt`, `max_attempts`,
`table_name`, and the final error.

rust/sdk/src/landing_zone.rs
```diff
@@ -158,11 +158,18 @@ impl<T: Clone> LandingZone<T> {
     ///
     /// This is used during stream recovery to re-send items that were observed but
     /// not yet acknowledged by the server.
-    pub fn reset_observe(&self) {
+    ///
+    /// # Returns
+    ///
+    /// The number of items moved back to the queue.
+    pub fn reset_observe(&self) -> usize {
```
teodordelibasic-db, line 165: Can we add a small unit test (observe N; `reset_observe()` returns N; queue-only state returns 0). We can also pair it with a multi-record batch so the batch-vs-record semantics get pinned at the same time.

elenagaljak-db: Yep

rust/sdk/src/lib.rs
elenagaljak-db: I agree with this, but I can't seem to find a clean solution for this records-per-batch sum calculation. 
One option is to expose an iterator in landing zone, but that gets tricky with holding locks. 
Another option is too have a method that would calculate the number in the landing zone but it breaks the responsibilities of the landing zone if it becomes aware of the type it holds. 
Third option is to keep a counter inside the stream, which I also dislike. 
What do you think, any suggestions?

rust/sdk/src/lib.rs
elenagaljak-db: Added

rust/sdk/src/lib.rs
elenagaljak-db: Added

teodordelibasic-db (approved):
LGTM

PR #398, opened 2026-06-22
[Rust][FFI] Split lib.rs into per-surface modules

### LINKED ISSUE #391: [Rust][FFI] Split lib.rs into per-functionality modules, leaving lib.rs as declarations + re-exports
### SDK

Rust

### Description

rust/ffi/src/lib.rs has grown to ~2,250 lines and now mixes several unrelated FFI surfaces in one file: Arrow Flight, the SDK + builder, stream ingestion, the shared result/header/error plumbing, and the new dynamic-protobuf API. The companion tests.rs is another ~1,070 lines.

A single file this size makes the crate hard to navigate, produces noisy diffs and review collisions when unrelated areas change in the same PR (e.g. an Arrow change touching the same file as a proto-schema change), and obscures the natural boundaries between the FFI surfaces. The file already self-documents those boundaries with // === banner sections (Arrow Flight FFI, SDK Functions, ZerobusSdkBuilder FFI, Dynamic Protobuf FFI), which makes it a good candidate for a mechanical, behavior-preserving split.

This is a pure refactor: no public API, ABI, or behavior change. The hard requirement is that the cbindgen-generated zerobus.h is byte-identical before and after, because Go and Java link against that header as the ABI contract.

### Proposed Solution

Break lib.rs into one module per FFI surface, plus a shared module for cross-cutting plumbing. lib.rs keeps only: crate-level docs (//!), shared uses, the mod/pub mod declarations, pub use re-exports of the public types, and mod tests;.

### Additional Context

_No response_

zlata-stefanovic-db:

Fixes #391. Splits the ~2,250-line `rust/ffi/src/lib.rs` into one module per FFI surface, leaving `lib.rs` with only module declarations and re-exports:

- `common.rs` — shared `#[repr(C)]` types (`CResult`, `CRecord(Array)`, `CHeader(s)`, `CStreamConfigurationOptions`, opaque `CZerobusSdk`/`CZerobusStream`, `HeadersProviderCallback`, `CallbackHeadersProvider`) and the `pub(crate)` helpers (`validate_*`, `write_*`, `c_str_to_string`, `intern_header_key`, `c_record_type`, `init_logging`, `RUNTIME`).
- `arrow.rs`, `builder.rs`, `sdk.rs`, `stream.rs`, `proto_schema.rs` — the types and `extern "C"` functions for each surface.

**This is a pure, behavior-preserving refactor:** no API, ABI, or behavior change. The cbindgen-generated `zerobus.h` is **byte-identical** before and after (verified against a pre-refactor baseline). Visibility is unchanged — only the C types and `#[no_mangle]` functions are `pub`; all helpers stay `pub(crate)`.

### Reviewer notes (intentional quirks, to preserve the byte-identical header)
cbindgen emits functions in source-walk order, so the **`mod` declaration order in `lib.rs` is load-bearing**, and a comment between each `mod` line prevents rustfmt's `reorder_modules` from alphabetizing (and thus reordering the header). For the same ordering reason, two cross-cutting functions sit in non-obvious modules: `zerobus_free_headers` is in `arrow.rs` and `zerobus_free_error_message` is in `stream.rs`. `build.rs` now watches the whole `src/` tree so the header regenerates on any module change (previously only `lib.rs` was watched).

## How is this tested?

Pure refactor verified by: `cargo build -p zerobus-ffi` succeeds, `cargo test -p zerobus-ffi` passes all 52 tests (unchanged — `tests.rs` is byte-identical), `cargo clippy -p zerobus-ffi --tests` clean (only the pre-existing `module_inception` on `mod tests`), and `diff` of the regenerated `zerobus.h` against the pre-refactor baseline is empty (byte-identical ABI).


teodordelibasic-db (approved):
Nice change!

irinatomic-db (approved):

PR #413, opened 2026-06-23
[Rust] Dynamic protobuf (proto from UC)

### LINKED ISSUE #372: [Rust] Dynamic protobuf: runtime descriptor fetch + dynamic message ingestion
Part of #99.

Descriptor generation from a Unity Catalog schema already shipped in Rust (PR #200; UC to Arrow in PR #270). This ticket adds the dynamic message path so a user can build and ingest records at runtime against a descriptor obtained without a compiled `.proto` file — whether that descriptor comes from Unity Catalog or is constructed in code.

This is the keystone ticket: it sets the cross-SDK API contract, so the four wrapper tickets are blocked by it.

## Decide the API contract first

Pick the canonical way a user supplies records dynamically, and document the choice so wrappers can follow it:

- native dynamic-message objects, or
- JSON records encoded against the descriptor (a C/FFI encode path is being added in PR #371).

A given SDK may expose both, but the core path should be consistent across languages.

## Scope

- [ ] Fetch a table's descriptor from Unity Catalog at stream creation (no local `.proto`)
- [ ] Provide an ergonomic way to construct a descriptor programmatically in code (no UC, no `.proto`) as an alternative descriptor source
- [ ] Build and ingest records dynamically against the descriptor, without statically compiled proto types
- [ ] Example, unit tests, changelog entry, doc comments

andrijast-db:

Closes #372

Ingest on the protobuf path with no compiled `.proto` and no generated types.
The descriptor is obtained at runtime; records are supplied as JSON and encoded to protobuf bytes client-side. Canonical contract = JSON-against-descriptor.

- `StreamBuilder::proto_from_uc()` — fetch the descriptor from Unity Catalog at `build()`
- `schema::TableDescriptorBuilder` — build a descriptor in code (no UC)
- `DynamicProtoEncoder` via `ZerobusStream::encoder()` — JSON → proto bytes, then ingest through the existing `ingest_record_offset`

Adds tests, a `proto_dynamic` example, README + changelog entries; `prost-reflect` becomes a regular SDK dependency.

## How is this tested?

Manually tested with custom Rust client

### DIFF
--- a/rust/examples/README.md
+++ b/rust/examples/README.md
@@ -40,6 +40,7 @@ The SDK supports two serialization formats and two ingestion methods:
 | [JSON Batch](json/README.md#batch-example) | JSON | Batch | `cargo run -p rust-examples-json --example json_batch` |
 | [Proto Single](proto/README.md#single-record-example) | Protocol Buffers | Single-record | `cargo run -p rust-examples-proto --example proto_single` |
 | [Proto Batch](proto/README.md#batch-example) | Protocol Buffers | Batch | `cargo run -p rust-examples-proto --example proto_batch` |
+| [Proto Dynamic](proto/README.md) | Protocol Buffers (dynamic, no `.proto`) | Single-record | `cargo run -p rust-examples-proto --example proto_dynamic` |
 | [Arrow](arrow/README.md) | Arrow Flight (Beta) | `RecordBatch` | `cargo run -p example_arrow` |
 
 ## Prerequisites
--- a/rust/examples/proto/dynamic.rs
+++ b/rust/examples/proto/dynamic.rs
@@ -1,0 +1,143 @@
+//! Dynamic protobuf ingestion: no compiled `.proto`, no generated Rust types.
+//!
+//! Two descriptor sources are shown:
+//!  1. `.proto_from_uc()` — the SDK fetches the table's schema from Unity Catalog
+//!     at stream creation and derives the protobuf descriptor (the default,
+//!     used in `main`).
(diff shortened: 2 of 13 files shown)


rust/examples/proto/dynamic.rs
teodordelibasic-db, line 1: Just adding to the offline discussion, not 100% confident, but I think it's more clear to have another folder under `examples/proto/`, i.e. `dynamic/`, which then also branches into `single.rs` and `batch.rs` or something like that. We can also rename current `proto/` to `compiled/`

rust/examples/README.md
```diff
@@ -40,6 +40,7 @@ The SDK supports two serialization formats and two ingestion methods:
 | [JSON Batch](json/README.md#batch-example) | JSON | Batch | `cargo run -p rust-examples-json --example json_batch` |
 | [Proto Single](proto/README.md#single-record-example) | Protocol Buffers | Single-record | `cargo run -p rust-examples-proto --example proto_single` |
 | [Proto Batch](proto/README.md#batch-example) | Protocol Buffers | Batch | `cargo run -p rust-examples-proto --example proto_batch` |
+| [Proto Dynamic](proto/README.md) | Protocol Buffers (dynamic, no `.proto`) | Single-record | `cargo run -p rust-examples-proto --example proto_dynamic` |
```
teodordelibasic-db, line 43: Adding to comment in `examples/proto/dynamic.rs`, it feels off to have `Proto Dynamic` as a dimension here. IMO Proto should branch into dynamic and compiled.

PR #415, opened 2026-06-24
[C++] Add C++ SDK (RAII wrapper over the C FFI)

zlata-stefanovic-db:
## Summary

Adds a C++17 SDK under `cpp/`, a thin RAII wrapper over the Zerobus C FFI
(`rust/ffi`). It exposes the same gRPC streaming / OAuth / recovery / ingestion
engine as the other SDKs, with a C++-idiomatic surface:

- **Proto and JSON ingestion** — single, batched, and fire-and-forget (`*_nowait`).
- **Arrow Flight ingestion** (Beta).
- **Dynamic protobuf from Unity Catalog metadata** (`ProtoSchema`) — build a
  descriptor and encode records with no `.proto` file or `protoc`.
- **Custom auth** via a `HeadersProvider` interface.
- Failures surface as `zerobus::ZerobusException` (`is_retryable()`); resources
  are RAII and move-only (handles freed exactly once).

## Distribution

CMake + GitHub Releases only — no package manager (no Conan/vcpkg), mirroring the
C FFI:

- Build from source via `add_subdirectory` / `FetchContent`, or
- `find_package(zerobus)` against an install tree or a prebuilt per-platform
  archive attached to a `cpp/v*` GitHub Release. `cmake --install` produces that
  archive: headers, `libzerobus_cpp`, the bundled `libzerobus_ffi`, and the
  `find_package` package config.

## Testing

- Hermetic, dependency-free unit suite (tiny in-repo harness — no framework,
  package manager, or network), run via `make test`. Covers marshalling, RAII /
  move, `CResult` → exception mapping, the headers trampoline, config
  conversion, and the proto-schema round trip.
- `ZEROBUS_SANITIZE` CMake option + a `sanitize` CI job that runs the suite under
  AddressSanitizer (with `ASAN_OPTIONS=detect_leaks=0`, since the core's global
  tokio runtime is intentionally never freed) to catch the memory bugs a thin
  FFI wrapper is prone to.

Rationale for not adding a simulated-server/integration test: the C++ layer is a
thin forwarder; the server behavior is covered end-to-end by the Rust mock tests,
and the same C FFI binding path is exercised end-to-end by the Go integration
tests. ASan targets the only genuinely C++-specific risk (memory/lifetime in the
glue) without adding dependencies.

## Docs and examples

README, per-API doc comments on the public headers, file- and implementation-
level comments in `src/`, runnable `examples/`, and `cpp/CLAUDE.md`.

## Test plan

- `make test` — hermetic unit suite.
[...]

PR #420, opened 2026-06-24
[C++] Core SDK implementation and build (split 3/6 from #415)

zlata-stefanovic-db:
## Summary

The core of the C++ SDK: a C++17, header + static-library wrapper over the
Zerobus **C FFI** (`rust/ffi`), which in turn wraps the Rust core. It gives C++
callers the same gRPC streaming / OAuth / recovery / ingestion engine as the
other SDKs, behind an idiomatic, exception-based, RAII API. This PR contains the
public API, its implementation, the build, and CI - but **not** the tests or
examples (those are stacked on top; see Merge order). Arrow Flight ingestion is
**not** part of this PR; it is peeled into its own PR in the split. The
fire-and-forget `*_nowait` ingestion APIs are **not** part of the initial C++
version — they are omitted here to avoid a silent data-loss footgun (see Design
notes and #473).

## What's in this PR

**Public API -`cpp/include/zerobus/`** (the surface; opaque, forward-declared
FFI handles only, so `zerobus.h` never leaks into consumers):
- `Sdk` / `SdkBuilder` + `TableProperties` - connection factory and stream creation.
- `Stream` - proto/JSON ingestion: single and batched (blocking, offset-returning),
  plus `flush`, `wait_for_offset`, `get_unacked_records`, `close`.
- `ProtoSchema` - build a descriptor + encode records straight from Unity Catalog
  table metadata (no `.proto`/`protoc`).
- `HeadersProvider` - custom auth headers.
- `ZerobusException` (`is_retryable()`), `StreamOptions`, `UnackedRecord`,
  `version()`, and the `zerobus.hpp` umbrella header.

**Implementation -`cpp/src/`** (the only place that includes `zerobus.h`):
- `sdk.cpp`, `stream.cpp`, `proto_schema.cpp` forward to the C FFI;
  `headers_callback.cpp` is the `extern "C"` trampoline.
- `src/detail/` internals: `ResultGuard` (`CResult` - `ZerobusException`, always
  freeing the C error string), the `StreamOptions` - C config conversion, and the
  trampoline declaration.

**Build & CI:**
- `CMakeLists.txt` - library target (`zerobus::zerobus`), the `install`/
  `find_package(zerobus)` export with the FFI archive bundled, and the
  `ZEROBUS_SANITIZE` option (off by default).
- `cmake/BuildRustFfi.cmake` builds `libzerobus_ffi` from local Rust source by
  default, or links a prebuilt lib via `-DZEROBUS_FFI_LIBRARY=`.
- `Makefile` (`build`/`test`/`lint`/`fmt`, `SANITIZE=` pass-through), `.clang-format`.
- `ci-cpp.yml` (fmt + test + Addresspush.yml`'s path filter.

## Design notes

- **Memory ownership**: Rust owns every handle; wrapper classes are move-only and
  free their handle exactly once (the source is nulled on move). Errors are thrown,
  never returned.
- **No gRPC/Protobuf C++ dependency**: all marshalling crosses the C FFI as byte
  buffers / pointer arrays; the batch helpers build only the small pointer/length
  arrays the C entry points need.
- **No fire-and-forget (`*_nowait`) ingestion in this release**: the blocking
  `ingest_*` APIs are already high-throughput with the loop-then-`flush()` pattern.
  The FFI `*_nowait` entry points return after merely spawning a background enqueue
  task, so `ingest_*_nowait(rec); flush(); close();` can race the task past
  `close()` and silently drop `rec` (the task's error is discarded). Rather than
  ship that footgun, the C++ surface exposes only the blocking variants. Tracked in
  #473 for a possible future version with proper drain-on-close semantics.
- **Distribution** (separate PRs): CMake + GitHub Releases, no package manager.

## Review follow-ups addressed

- `wait_for_offset()` now rejects negative offsets (e.g. the `-1` an empty batch
  returns) with a clear `ZerobusException` instead of forwarding the sentinel to
  the FFI.
- The raw `ingest_proto_record(data, len)` overload now treats `{nullptr, 0}` as a
  valid empty record (matching the vector overload) instead of a null-pointer error.
- `HeadersProvider`'s lifetime docs no longer claim a `shared_ptr` is "sufficient";
  they warn that a `get_headers()` outrunning the ~1s close budget can be invoked on
  a freed provider (see #436).

## Merge order

Off `main`. **Tests and examples are stacked on this PR** and merge after it. The
`add_subdirectory(tests)`/`(examples)` wiring is intentionally omitted here and
arrives with those PRs, so this branch configures and builds the library cleanly
on its own.

Part of the #415 split.

## Test plan

- `make build` - configures and builds the library (FFI from source).
- `make lint` - clang-format check + `-Wall -Wextra`.
- Verified a separate project can `find_package(zerobus)` and link
  `zerobus::zerobus` from a `cmake --install` tree.
- Unit tests and the AddressSanitizer run land with the stacked tests PR.


cpp/include/zerobus/stream.hpp
```diff
@@ -0,0 +1,127 @@
...
+
+  /// Fire-and-forget batch ingestion. Copies the payloads before returning, so
+  /// the caller's buffers may be released immediately. An empty batch is a
+  /// no-op.
+  void ingest_proto_records_nowait(
+      const std::vector<std::vector<std::uint8_t>>& records);
+  void ingest_json_records_nowait(const std::vector<std::string>& records);
+
+  /// Block until the record at `offset` has been acknowledged by the server.
+  void wait_for_offset(std::int64_t offset);
```
irinatomic-db, line 90: **Not a blocker for this PR, just something to think about and see if it makes sense and is worth doing as a follow-up.**

The C++ SDK has the pull-based ack model (ingest → offset → wait_for_offset / flush) but no push-based AckCallback like Python and Java have.

For Go this gap is fine - **goroutines** are userspace-scheduled by the Go runtime, multiplexed onto a small number of OS threads. Blocking a goroutine on wait_for_offset just suspends it and yields its OS thread to another goroutine; no kernel involvement. You can have thousands of goroutines blocked on ack tracking with negligible cost.

C++ std::thread maps 1:1 to an OS thread: kernel-scheduled. Blocking one on `wait_for_offset parks` that OS thread doing nothing until the server ack arrives. Scaling this across many concurrent streams might get expensive fast.

The idiomatic C++ solution for "notify me when something happens" is a callback, not a blocked thread. With `AckCallback`, the Rust tokio runtime fires the callback from its own thread pool when the ack arrives - the application thread never blocks waiting for acks and no extra OS threads are needed.

For now the pull model is functional. I suggest looking into this as a follow up to see if it makes sense to implement it (adding AckCallback to the C FFI). Some issues I see in implementation to think about:
- Language boundaries
- Callback object lifetime

cpp/include/zerobus/stream.hpp
```diff
@@ -0,0 +1,127 @@
...
+  /// Ingest a batch of protobuf records. Returns the offset of the last record
+  /// in the batch. Prefer batch APIs over per-record calls in hot paths — each
+  /// FFI crossing has a fixed cost. Throws `ZerobusException` if `records` is
+  /// empty (an empty batch has no offset to return).
+  std::int64_t ingest_proto_records(
+      const std::vector<std::vector<std::uint8_t>>& records);
+
+  /// Ingest a batch of JSON records. Returns the offset of the last record.
+  /// Throws `ZerobusException` if `records` is empty.
+  std::int64_t ingest_json_records(const std::vector<std::string>& records);
```
irinatomic-db: What do other SDKs do in case that `records` is empty, do they also throw an exception?

zlata-stefanovic-db: Most of them don't, this is a good idea to change it for consistency

zlata-stefanovic-db: Just made this change, thank you! @irinatomic-db

cpp/include/zerobus/stream.hpp
```diff
@@ -0,0 +1,126 @@
...
+  ///
+  /// WARNING — do not combine the `_nowait` APIs with a custom
+  /// `HeadersProvider`. A fire-and-forget task is detached: neither `close()`
+  /// nor the destructor drains it, and a task that still needs fresh headers
+  /// may call back into the provider after the `Stream` (and the `shared_ptr`
+  /// keeping the provider alive) is destroyed — a use-after-free. The FFI
+  /// exposes no way to drain these tasks, so there is no safe ordering. With a
+  /// `HeadersProvider`, use only the blocking ingest variants, which complete
+  /// before they return.
+  void ingest_proto_record_nowait(const std::uint8_t* data, std::size_t len);
```
irinatomic-db: The doc warning is correct but insufficient - it relies on the caller reading and following the warning. The Rust background task is detached; neither `close()` nor the destructor drains it, so a task that needs fresh headers can call back through a raw pointer after `provider_` is destroyed. This should be enforced in code, not just documented. 

Maybe smth like:
```cpp
  void Stream::ingest_proto_record_nowait(...) {
      if (provider_ != nullptr)
          throw ZerobusException("_nowait APIs cannot be used with a custom HeadersProvider", false);
      ...
  }
```

cpp/src/stream.cpp
```diff
@@ -0,0 +1,260 @@
...
+// The ingest methods below share one shape: route a CResult through a
+// ResultGuard, call the matching zerobus_stream_* entry point, then
+// throw_if_error(). The blocking variants return the server-assigned offset;
+// the _nowait variants return void and only report argument-validation errors.
+
+std::int64_t Stream::ingest_proto_record(const std::uint8_t* data,
+                                         std::size_t len) {
+  detail::ResultGuard guard;
+  std::int64_t offset =
+      zerobus_stream_ingest_proto_record(handle_, data, len, guard.ptr());
```
irinatomic-db: The FFI returns -2 for an empty batch and -1 for errors, alongside CResult.success. C++ relies entirely on throw_if_error() and hands the raw offset straight to the caller. If the FFI ever returned a negative sentinel with success == true, the caller would get -2 as a real offset and pass it to wait_for_offset(-2). Go defends against this explicitly (`if offset == -2 { return -1, nil } then if offset < 0 { ... error  }`). Worth adding a post-call guard:
```cpp
guard.throw_if_error();
  if (offset < 0)
      throw ZerobusException("unexpected negative offset from FFI", false);
  return offset;
```

cpp/include/zerobus/stream.hpp
zlata-stefanovic-db: Agree with this. I discussed it with Danilo already — it’s a real issue, will do this C++ mitigation now.
The reason I didn’t do the full fix in this PR is that the root issue is at the C FFI contract, not just in C++ wrapper logic. For this beta PR we kept the warning-only mitigation to avoid cross-SDK FFI changes right now, but we should enforce it in code.
I’ll file a follow-up for _nowait + custom HeadersProvider enforcement (starting with C++ guard, then proper FFI-safe lifecycle fix).

cpp/src/stream.cpp
zlata-stefanovic-db: Addressing this as well, thank you!

cpp/include/zerobus/stream.hpp
zlata-stefanovic-db: Made the mitigation now

irinatomic-db (dismissed):
Approving but let's wait for others to review as well.

cpp/include/zerobus/stream.hpp
zlata-stefanovic-db: Opened the follow up issue: https://github.com/databricks/zerobus-sdk/issues/436

cpp/include/zerobus/stream.hpp
zlata-stefanovic-db: Added a follow up issue: https://github.com/databricks/zerobus-sdk/issues/437

cpp/include/zerobus/stream.hpp
zlata-stefanovic-db: Following up here - after tracing this end to end, I ended up removing the _nowait guard and the related docs, because the _nowait-specific rationale turned out to be a wrong conclusion. Wanted to lay out why, since it's the opposite of what I said earlier.

The _nowait task doesn't actually reach the provider. It calls ingest_record_offset - ingest_internal_v2 (rust/sdk/.../ingest.rs), which only enqueues: it checks is_closed (returns StreamClosedError if the stream is closing), takes the lock, assigns an offset, and appends to the landing zone. It never connects, never sends, never touches get_headers. The only caller of get_headers is the supervisor, during connect/recovery (connection.rs:74), and both close() and Drop cancel the CancellationToken and abort() the supervisor. So a detached _nowait task running after teardown just appends to a landing zone whose supervisor is already gone - no callback, no UAF.

That said, your underlying concern is real - there is a provider use-after-free, just via a different (and not _nowait-specific) path:

- get_headers for the FFI provider is a synchronous call into the raw user_data pointer (rust/ffi/src/common.rs:211), so tokio's abort() - being cooperative - can't preempt a worker thread that's inside it.
- close() waits only SHUTDOWN_TIMEOUT_SECS = 1 for the supervisor, then abort()s without re-awaiting (close.rs).
- The provider is then freed immediately (~Stream drops the shared_ptr), with no handshake.

So if a server-driven recovery is in flight and its get_headers (e.g. a token fetch) runs longer than 1s while the stream is being torn down, the worker can still be inside the C++ callback when the provider is freed. This applies to blocking ingest too, not just _nowait.

I've reframed #436 around that real race (fix belongs at the FFI level - take ownership of the provider so it's freed only once no task can call it), and removed the _nowait guard + the "incompatible with _nowait" docs since they targeted the wrong mechanism. Thanks for pushing on this - your instinct that the lifetime issue was real and that a doc warning wasn't enough was right; it just isn't _nowait that triggers it.

.github/workflows/push.yml
```diff
@@ -113,6 +126,17 @@ jobs:
     with:
       use_local_sdk: true
 
+  # The C++ build always compiles the Rust core from local source, so this job
+  # inherently tests against local Rust when rust/** changes — no use_local_sdk
+  # toggle needed (ci-cpp.yml has no such input).
```
elenagaljak-db: We don't need this comment

cpp/cmake/BuildRustFfi.cmake
```diff
@@ -0,0 +1,84 @@
+# Builds the Zerobus C FFI static library from the local Rust source and exposes
+# it as an imported target `zerobus::ffi`.
+#
+# Mirrors go/build_rust.sh: invokes `cargo build --release` in `rust/ffi`, which
+# produces `rust/target/release/libzerobus_ffi.a` and regenerates `zerobus.h`
+# (via the crate's cbindgen build script). The header directory is added to the
```
elenagaljak-db: We don't need to mention go here.

.github/workflows/push.yml
zlata-stefanovic-db: removed

cpp/cmake/BuildRustFfi.cmake
zlata-stefanovic-db: removed

cpp/src/headers_callback.cpp
```diff
@@ -0,0 +1,125 @@
...
+  if (headers.empty()) {
+    return result;
+  }
+
+  auto* arr =
+      static_cast<CHeader*>(std::calloc(headers.size(), sizeof(CHeader)));
+  if (arr == nullptr) {
+    return make_error("out of memory marshalling headers");
+  }
+
```
elenagaljak-db: Here header strings are allocated with std::malloc and the array with std::calloc. Rust frees the array with libc::free and each string with CString::from_raw. On Linux/macOS there's one libc, so malloc and free match, but on windows BuildRustFfi.cmake statically links the Rust .lib into the C++ binary and if the C++ translation unit and the Rust staticlib resolve to different CRT heaps (e.g., static vs dynamic CRT, or /MT vs /MD), freeing a C++-malloc pointer through Rust's allocator is heap corruption. Could we check this? Ideally we should have Rust free what Rust allocates

cpp/include/zerobus/stream.hpp
```diff
@@ -0,0 +1,117 @@
...
+class Stream {
+ public:
+  ~Stream();
+
+  Stream(Stream&& other) noexcept;
+  Stream& operator=(Stream&& other) noexcept;
+  Stream(const Stream&) = delete;
+  Stream& operator=(const Stream&) = delete;
+
+  /// Ingest a single protobuf-encoded record. Returns the logical offset
```
elenagaljak-db: Just a nit, maybe these comments should be more descriptive or doc looking? Kind of like these:
```
def ingest_batch(self, batch) -> int:
Ingest a pyarrow.RecordBatch or pyarrow.Table.

Args:
    batch: A pyarrow.RecordBatch or pyarrow.Table to ingest.

Returns:
    The offset ID assigned to this batch.
```

cpp/src/headers_callback.cpp
zlata-stefanovic-db: Fixed now (rust frees what rust allocates):  rather than touch zerobus_free_headers (Go relies on it), I added two additive FFI allocators - zerobus_alloc_header_array (libc::calloc) and zerobus_alloc_cstring (CString::into_raw) - that allocate with exactly the allocators it frees with, and routed the C++ trampoline through them.

cpp/include/zerobus/stream.hpp
zlata-stefanovic-db: Done

rust/ffi/src/arrow.rs
```diff
@@ -651,6 +651,57 @@ pub extern "C" fn zerobus_arrow_get_default_config() -> CArrowStreamConfiguratio
...
+/// resolve to different CRT heaps; allocating here keeps the alloc/free pair on
+/// one heap and avoids the cross-heap free that would otherwise corrupt memory.
+///
+/// Zero-initialised so a partially populated array is safe to pass to
+/// `zerobus_free_headers` (unset key/value pointers are null and skipped).
+/// Returns null if `count` is 0 or the allocation fails.
+// Not wrapped in `ffi_guard`: a pure `calloc` that returns null on failure,
+// with no panic-capable operation.
+#[no_mangle]
+pub extern "C" fn zerobus_alloc_header_array(count: usize) -> *mut CHeader {
```
davidtosovic-db, line 670: Nit: This is not important and should not be done in this PR.
All of these header functions don't really have to do anything with arrow. We should probably move them to the common file.

zlata-stefanovic-db: Noted, filling an issue for this

davidtosovic-db (dismissed):

cpp/include/zerobus/stream.hpp
```diff
@@ -0,0 +1,174 @@
...
+
+  /// Ingest a batch of JSON records, blocking until they are queued.
+  ///
+  /// @param records The records, each a UTF-8 JSON string.
+  /// @return The logical offset of the last record in the batch, or -1 if
+  ///         @p records is empty (a no-op).
+  /// @throws ZerobusException if the stream is closed or ingestion fails.
+  std::int64_t ingest_json_records(const std::vector<std::string>& records);
+
+  /// Fire-and-forget single-record ingestion: queues the record on a background
```
teodordelibasic-db: The bug we discussed offline seems real. `_nowait` returns success after the FFI spawns the enqueue task, before the record is queued (`rust/ffi/src/stream.rs:504-509`). `ingest_*_nowait(rec); flush(); close();` can lose `rec`: `flush()` runs before the task enqueues (nothing pending), `close()` flips `is_closed`, then the task enqueues into a closed stream and drops the error (`ingest.rs:189`, `let _ = ...`). No C++ primitive observes enqueue completion, and `flush()` looks like the natural drain.

There seem to be a few long term solutions which we can pick up as a follow up if we decide the `nowait` API is useful in C++.

cpp/src/stream.cpp
```diff
@@ -0,0 +1,317 @@
...
+    return;
+  }
+  JsonBatchView v = make_json_batch(records);
+  detail::ResultGuard guard;
+  zerobus_stream_ingest_json_records_nowait(handle_, v.ptrs.data(),
+                                            v.ptrs.size(), guard.ptr());
+  guard.throw_if_error();
+}
+
+void Stream::wait_for_offset(std::int64_t offset) {
```
teodordelibasic-db, line 208: `wait_for_offset()` forwards any `std::int64_t` to the FFI, including the `-1` an empty batch returns from `ingest_*_records`. Ingest returns are guarded by `checked_offset` (`stream.cpp:42-48`) but the input to `wait_for_offset` is not, so `wait_for_offset(ingest_json_records({}))` passes `-1` across the boundary.

We can reject negative offsets in `wait_for_offset()` with a clear `ZerobusException`, mirroring `checked_offset`.

cpp/include/zerobus/headers_provider.hpp
```diff
@@ -0,0 +1,38 @@
...
+/// invokes `get_headers()` whenever it needs fresh headers, possibly from an
+/// internal worker thread.
+///
+/// Thread safety: the core guards against *concurrent* invocations and will
+/// surface an error if two calls overlap, but `get_headers()` may still be
+/// called from a different thread than the one that created the stream.
+/// Implementations should therefore be thread-safe with respect to their own
+/// state.
+///
+/// Lifetime: a provider passed to `Sdk::create_stream` must remain alive for as
```
teodordelibasic-db: Small addition to prior discussion regarding this.

The comment says holding a `shared_ptr` for the `Stream`'s lifetime is "sufficient", but that is the unsafe condition issue #436 describes. `get_headers()` is a synchronous, non-preemptible callback reached during recovery; `close()` waits only 1s then `abort()`s and `~Stream()` drops the provider, so a slow `get_headers()` can run through a freed object. Removing the old `_nowait` guard left this header reassuring where it should warn.

We can replace the "shared_ptr is sufficient" line with the real caveat, a `get_headers()` that can exceed the ~1s shutdown budget may be invoked after the `Stream`/provider is destroyed; keep it well under that budget or keep the provider alive past the `Stream`.

cpp/src/stream.cpp
```diff
@@ -0,0 +1,317 @@
...
+  }
+  return *this;
+}
+
+// The ingest methods below share one shape: route a CResult through a
+// ResultGuard, call the matching zerobus_stream_* entry point, then
+// throw_if_error(). The blocking variants return the server-assigned offset;
+// the _nowait variants return void and only report argument-validation errors.
+
+std::int64_t Stream::ingest_proto_record(const std::uint8_t* data,
```
teodordelibasic-db, line 148: This overload forwards `(nullptr, 0)` to the FFI, which rejects a null data pointer even at `len == 0` (`rust/ffi/src/stream.rs:231`). The vector overload maps empty input to a non-null sentinel (`stream.cpp:154-157`), so `ingest_proto_record(std::vector<std::uint8_t>{})` succeeds while `ingest_proto_record(nullptr, 0)` throws, which is inconsistent handling of the same empty payload. `{nullptr, 0}` is a standard empty span.

We can treat `data == nullptr` as an error only when `len > 0`; for `len == 0` substitute the same sentinel the vector overload uses.

cpp/include/zerobus/stream.hpp
zlata-stefanovic-db: Filled an issue for this: https://github.com/databricks/zerobus-sdk/issues/473, might be good to omit the ingest_nowait() from the first C++ release - there seem to be a couple of issues surrounding it

cpp/src/stream.cpp
zlata-stefanovic-db: Done now, rejecting negative offsets in wait_for_offset().

cpp/include/zerobus/headers_provider.hpp
zlata-stefanovic-db: Added this part

cpp/src/stream.cpp
zlata-stefanovic-db: Done as suggested

cpp/include/zerobus/stream.hpp
teodordelibasic-db: Oh yes, I actually left the comment confusingly worded due to rewriting. Let's remove it for start and track to add it later, although I wouldn't say it's high priority.

zlata-stefanovic-db: Removed it from this PR. @teodordelibasic-db

teodordelibasic-db (approved):
LGTM to merge, let's just also update the PR description to latest state, i.e. `nowait` API being removed from the initial version. 🙂

PR #425, opened 2026-06-25
[Java] Add streamBuilder() fluent API and deprecate create*Stream methods

danilonajkov-db:
## What

Adds `ZerobusSdk.streamBuilder()`, a fluent builder for creating Zerobus streams, mirroring the Rust SDK's `stream_builder()`. JSON, Protocol Buffer, and Arrow Flight streams are all created through one chainable API:

```java
ZerobusJsonStream stream = sdk.streamBuilder()
    .table("catalog.schema.table")
    .oauth(clientId, clientSecret)
    .recovery(true)
    .maxInflightRecords(100)
    .json()
    .build()
    .join();
```

The existing `createJsonStream` / `createProtoStream` / `createArrowStream` methods are now `@Deprecated` in favour of `streamBuilder()` and scheduled for removal in the next major release.

## Approach

The builder accumulates configuration in Java and crosses the FFI once at `build()` via package-private `create{Json,Proto,Arrow}StreamInternal` cores (the existing `nativeCreateStream` / `nativeCreateArrowStream` path, which already drives the Rust `StreamBuilder`). The deprecated public factory methods delegate to the same cores, so the builder never depends on deprecated API.

## Verification

- StreamBuilderTest`: 11/11 pass (validation, gRPC defaults, Arrow default preservation, shared + Arrow-specific setters).
- Manually tested vs staging all 3 formats


java/src/main/java/com/databricks/zerobus/StreamBuilder.java
```diff
@@ -0,0 +1,412 @@
...
+  }
+
+  // ==================== Package-private helpers ====================
+
+  /** Validates that the required table name and authentication have been configured. */
+  void validateRequired() {
+    if (tableName == null || tableName.isEmpty()) {
+      throw new IllegalStateException("table name is required: call table()");
+    }
+    if (clientId == null) {
```
teodordelibasic-db: We should probably validate `clientSecret` as well.

java/src/main/java/com/databricks/zerobus/StreamBuilder.java
```diff
@@ -0,0 +1,412 @@
...
+  /**
+   * Sets the maximum number of in-flight records.
+   *
+   * <p>Applies to JSON and Protocol Buffer (gRPC) streams. It is ignored for Arrow streams, which
+   * use {@link ArrowStreamBuilder#maxInflightBatches(int)} instead.
+   *
+   * @param maxInflightRecords the maximum number of in-flight records
+   * @return this builder for method chaining
+   */
+  public StreamBuilder maxInflightRecords(int maxInflightRecords) {
```
teodordelibasic-db, line 161: The numeric setters (`maxInflightRecords`, `recoveryTimeoutMs`, `recoveryRetries`, etc.) accept negatives and store them unchecked. The JNI extractor casts them straight to unsigned (`.i()? as usize` / `as u64` / `as u32`), so `maxInflightRecords(-1)` becomes a huge limit rather than an error. (`streamPausedMaxWaitTimeMs(-1)` is the one intentional negative, handled at options.rs:148.) We should probably reject non-positive limits/timeouts and negative retry counts before storing.

java/src/main/java/com/databricks/zerobus/ZerobusSdk.java
teodordelibasic-db: This comment was confusing in and of itself even before IMO, but let's remove this section which says `recommended`. Same goes for JSON probably.

java/README.md
teodordelibasic-db: I think there are probably a few stale `createArrowStream` / `createProtoStream` / `createJsonStream` like this one.

java/src/test/java/com/databricks/zerobus/StreamBuilderTest.java
```diff
@@ -0,0 +1,167 @@
...
+ * Unit tests for {@link StreamBuilder}.
+ *
+ * <p>These tests verify required-field validation and configuration accumulation without requiring
+ * the native library. They exercise the builder up to (but not including) {@code build()}, which is
+ * the only operation that touches native code, so a {@code null} SDK reference is sufficient.
+ */
+public class StreamBuilderTest {
+
+  private static StreamBuilder builder() {
+    return new StreamBuilder(null);
```
teodordelibasic-db, line 33: Not sure how plausible it is, but LLM says we should improve E2E testing for this a bit:

> Tests use `new StreamBuilder(null)` and stop before `build()`, so nothing verifies that `json()`/`compiledProto()`/`arrow()` route to the correct `*Internal` method with the right table/credentials/options. The recommended API could mis-wire a terminal builder and still pass all 11 tests. Consider a Mockito spy on `ZerobusSdk` asserting each sub-builder calls the expected internal method, or one builder-path integration test per format.

java/src/main/java/com/databricks/zerobus/ZerobusSdk.java
```diff
@@ -130,13 +151,19 @@ public ZerobusSdk(String serverEndpoint, String unityCatalogEndpoint) {
    * @param clientSecret The OAuth client secret for authentication.
    * @return A CompletableFuture that completes with the ZerobusProtoStream when the stream is
    *     ready.
+   * @deprecated Use {@link #streamBuilder()} instead, for example {@code
+   *     sdk.streamBuilder().table(tableName).oauth(clientId,
+   *     clientSecret).compiledProto(descriptorProto).build()}. This method will be removed in the
+   *     next major release.
    */
+  @Deprecated
```
teodordelibasic-db, line 156: Let's add `@Deprecated(since = "<current>", forRemoval = true)` to all deprecations.

teodordelibasic-db (commented):
Looks good, mostly minor stuff I'd say.

LLM generated list of reference/usage sites that should be migrated to `streamBuilder()` or removed. For tests we should probably maintain both the new and old API until we remove the old one.

### Javadoc pointers in main source (point users at the deprecated factories)
- `ZerobusSdk.java:197` — "the deprecated `createProtoStream` overloads"
- `ZerobusSdk.java:296` — "the deprecated `createJsonStream` overloads"
- `ZerobusSdk.java:437` — "the deprecated `createArrowStream` overloads"
- `ArrowStreamConfigurationOptions.java:23` — `@see ZerobusSdk#createArrowStream`
- `BaseZerobusStream.java:24` — `sdk.createJsonStream(...)` in class example
- `ZerobusJsonStream.java:13, 16, 37, 110` — "Create instances using `createJsonStream`" / example / `@see` / ctor doc
- `ZerobusProtoStream.java:16, 19, 40, 50` — same pattern for `createProtoStream`
- `ZerobusArrowStream.java:27, 35, 65, 92, 271` — same pattern for `createArrowStream`

### README.md
- `:652` `createArrowStream` (Arrow quick-start)
- `:683` `createProtoStream` (offset-based example)
- `:759, :763, :799` `createJsonStream` (JSON section presented as recommended)
- `:1006, :1015` `createJsonStream` (API reference)
- `:1035, :1045` `createArrowStream` (API reference)

### examples/ (docs)
- `examples/README.md:93` `createArrowStream`
- `examples/arrow/README.md:92, :121` `createArrowStream`
- `examples/json/README.md:44` `createJsonStream`
- `examples/proto/README.md:45` `createProtoStream`
- `examples/legacy/README.md:77` `createProtoStream`

### Tests (still on the deprecated API; no builder coverage)
- `IntegrationTest.java`: `createProtoStream` at `:107, :184, :479, :533, :603, :658, :781, :863, :1096, :1100`; `createJsonStream` at `:278, :341, :723, :939`
- `ArrowIntegrationTest.java`: `createArrowStream` at `:102, :145, :203, :248, :296, :368, :425, :463, :504, :506, :542, :567, :590, :618, :652`

java/src/main/java/com/databricks/zerobus/StreamBuilder.java
danilonajkov-db: done

java/src/main/java/com/databricks/zerobus/StreamBuilder.java
danilonajkov-db: added requirePositive and requireNotNegative helpers for validation.

java/README.md
danilonajkov-db: replaced all comments and mentions of the old methods i could find

java/src/test/java/com/databricks/zerobus/StreamBuilderTest.java
danilonajkov-db: added some routing tests with a mocked zerobus sdk, to verify that the stream builder routes to the correct native method with a sdk mock

java/src/main/java/com/databricks/zerobus/ZerobusSdk.java
danilonajkov-db: removed, thanks

java/src/main/java/com/databricks/zerobus/ZerobusSdk.java
danilonajkov-db: added, thanks

teodordelibasic-db: Okay my bad, it seems I may have tricked you here.

`since` / `forRemoval` are Java 9+ members of `java.lang.Deprecated`. `pom.xml` declares `source/target 1.8` and the README/CLAUDE.md advertise Java 8+ source compatibility. CI is green only because it runs JDK 11/17/21 with the legacy `-source 8 -target 8` (no `--release 8`, no `<bootclasspath>`, no animal-sniffer), so javac resolves the annotation against its own 9+ `java.lang.Deprecated`. A genuine JDK 8 build — or switching the pom to the correct `--release 8` enforcement — fails immediately, and CI cannot catch it as configured. 

We should probably revert back to bare `@Deprecated` and put the since/removal note in the `@deprecated` Javadoc (consistent with the legacy deprecations in this file).

teodordelibasic-db (commented):
I think the Java 8 breakage due to deprecation annotations is the only blocker, sorry for that. Also here is a list of nits below compiled by LLM, feel free to resolve none/some/all.

- `StreamBuilder.java:220` — blank OAuth/table strings still pass: `oauth("", "")` / `table("")`... (only null is rejected). Reject empty/blank `tableName`/`clientId`/`clientSecret` and add a test.
- `StreamBuilder.java:12` — class Javadoc "Setters may be called in any order" is only true before format selection. Narrow to "shared setters may be called in any order, before selecting a record format."
- `StreamBuilder.java:61, 73, 174` — `table()` / `oauth()` / `ackCallback()` accept null silently, unlike `compiledProto()` / `arrow()` (which `requireNonNull`). Add `Objects.requireNonNull` for consistent fail-fast.
- `ZerobusSdk.java:133, 239` — deprecated `createProtoStream` / `createJsonStream` `<pre>` examples still call the deprecated method. Repoint to a `streamBuilder()` chain or drop.
- `README.md:659, 790`, `examples/README.md:102`, `examples/arrow/README.md:108` — `offset.ifPresent(stream::waitForOffset)` doesn't compile: `waitForOffset` throws checked `ZerobusException`. Use `if (offset.isPresent()) stream.waitForOffset(offset.get());`.
- `README.md:781` — `Map.of(...)` is Java 9+. Use `new HashMap<>()` for a Java 8 example.
- `ZerobusArrowStream.java:43` — Javadoc shows `long offset = stream.ingestBatch(batch)` but `ingestBatch` returns `Optional<Long>`.

java/src/main/java/com/databricks/zerobus/ZerobusSdk.java
danilonajkov-db: ah okay, will revert. We should add a java 8 build in the CI as well.

danilonajkov-db:
> I think the Java 8 breakage due to deprecation annotations is the only blocker, sorry for that. Also here is a list of nits below compiled by LLM, feel free to resolve none/some/all.
> 
> * `StreamBuilder.java:220` — blank OAuth/table strings still pass: `oauth("", "")` / `table("")`... (only null is rejected). Reject empty/blank `tableName`/`clientId`/`clientSecret` and add a test.
> * `StreamBuilder.java:12` — class Javadoc "Setters may be called in any order" is only true before format selection. Narrow to "shared setters may be called in any order, before selecting a record format."
> * `StreamBuilder.java:61, 73, 174` — `table()` / `oauth()` / `ackCallback()` accept null silently, unlike `compiledProto()` / `arrow()` (which `requireNonNull`). Add `Objects.requireNonNull` for consistent fail-fast.
> * `ZerobusSdk.java:133, 239` — deprecated `createProtoStream` / `createJsonStream` `<pre>` examples still call the deprecated method. Repoint to a `streamBuilder()` chain or drop.
> * `README.md:659, 790`, `examples/README.md:102`, `examples/arrow/README.md:108` — `offset.ifPresent(stream::waitForOffset)` doesn't compile: `waitForOffset` throws checked `ZerobusException`. Use `if (offset.isPresent()) stream.waitForOffset(offset.get());`.
> * `README.md:781` — `Map.of(...)` is Java 9+. Use `new HashMap<>()` for a Java 8 example.
> * `ZerobusArrowStream.java:43` — Javadoc shows `long offset = stream.ingestBatch(batch)` but `ingestBatch` returns `Optional<Long>`.

some of these were preexisting, but small changes so I fixed most of them here in this PR as well. Thanks

teodordelibasic-db (dismissed):
LGTM

teodordelibasic-db (approved):

PR #426, opened 2026-06-25
[Rust] Multiplexed stream ingest followups

### LINKED ISSUE #423: [Rust] MultiplexedStream follow-ups
Follow-ups from review of #188 (MultiplexedStream core). Both are agreed in unresolved review threads and should be addressed before the `testing`-gated feature is un-gated.

## 1. Poison/ack race — reserved-capacity ingest (must-do before un-gating)

There is a timing window where `ingest_*` returns `Ok(MessageId)` for a record that can never be acked:

- Between the second `is_closed` check and the actual enqueue, another task can poison the mux and flush past the record.
- `wait_for_capacity` only checks `landing_zone.len() < max_inflight_requests` without reserving the slot, so several concurrent callers can all pass that check and then queue past the limit / block inside `LandingZone::add`, which doesn't observe poisoning.

**Agreed plan:** add `pub(crate)` reserve-capacity ingest APIs on `ZerobusStream` so the mux can reserve a slot → take the ingest/stream-order lock → re-check closed → `enqueue_reserved` with no new blocking point. Sketch:

```rust
let slot = lane.stream.reserve_capacity().await?;
let _stream_order = lane.stream.ingest_lock().await;

let _admit = lane.admission.read().await;
self.check_closed()?;
lane.stream.check_open()?;

let off = lane.stream.enqueue_reserved(encoded_batch, slot);
Ok(MessageId::new(i, off))
```

Ref: https://github.com/databricks/zerobus-sdk/pull/188#discussion_r3436740227

## 2. Eager `is_closed()` detection of async sub-stream death

If an inner sub-stream dies asynchronously and no caller subsequently touches it (no `ingest_*`, `flush`, or `wait_for_message_id` to that stream), the outer `is_closed()` keeps returning `false` and round-robin writes that hit healthy sub-streams are still accepted.

**Agreed approach:** split into two methods —
- public `is_closed()` → `self.is_closed.load(...) || self.streams.iter().any(|s| s.is_closed())`
- private fast `is_closed` retaining the current lazy hot-path behavior.

Ref: https://github.com/databricks/zerobus-sdk/pull/188#discussion_r3242334378

---

Tracking issue requested on #188. cc @teodordelibasic-db

danilonajkov-db:

This fixes the `MultiplexedStream` poison/admission race and makes capacity waits bounded.

The new ingest flow is:

1. Reserve real landing-zone capacity on the selected sub-stream.
2. Enter the sub-stream's private ingest-ordering section.
3. Acquire the mux admission read lock and re-check mux state.
4. Re-check sub-stream state.
5. Allocate the offset and enqueue using the reserved capacity.

The poison path acquires the mux admission write lock before flushing sub-streams, so no record can be admitted between the final closed check and the poison flush.

`ZerobusStream` retains ownership of its internal ordering lock through `enqueue_reserved_admitted`; the mux supplies only the admission check. Capacity reservations are opaque and cannot expose the underlying semaphore implementation.

Capacity waits are bounded by the existing `flush_timeout_ms` option. If acknowledgments never free capacity, `ingest_record` / `ingest_records` return `ZerobusError::ConnectionTimeout` instead of waiting indefinitely. A capacity timeout does not poison an otherwise healthy mux.

## How is this tested?

- Regression test for concurrent ingests waiting for capacity while a sub-stream failure poisons the mux.
- Regression test for a permanently full sub-stream returning `ConnectionTimeout` after the configured timeout.
- Full multiplexed stream integration suite: 23 tests passed.
- Rust SDK unit suite: 141 tests passed.
- Doc tests passed.
- `cargo fmt --all --check`


rust/NEXT_CHANGELOG.md
```diff
@@ -21,6 +21,8 @@
 ### Bug Fixes
 - Redacted the OAuth authorization token from an error log and error message on the gRPC stream-setup path; a malformed token value is no longer written to logs.
 - A UC token that cannot be encoded as an HTTP `authorization` header value is now rejected at mint time rather than cached, so it cannot poison the cache and fail every stream creation until its refresh window.
+- Fixed `MultiplexedStream` ingest admission so concurrent callers reserve actual sub-stream capacity before taking the stream-order lock. If a sub-stream failure poisons the mux while other ingests are waiting for capacity, those ingests now fail instead of returning `Ok(MessageId)` for records that were never included in the poison-path flush.
```
elenagaljak-db: Nit: why is this added in the middle of bug fixes

rust/sdk/src/landing_zone.rs
```diff
@@ -4,6 +4,8 @@ use std::sync::Arc;
 use thiserror::Error;
 use tokio::sync::{Notify, OwnedSemaphorePermit, Semaphore};
 
+pub(crate) type CapacityReservation = OwnedSemaphorePermit;
+
```
elenagaljak-db: Any special reason we introduced a new type?

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -178,11 +227,17 @@ impl MultiplexedStream {
...
+            match tokio::time::timeout(
+                std::time::Duration::from_millis(backoff_ms),
+                stream.reserve_capacity(),
+            )
+            .await
+            {
+                Ok(Ok(reservation)) => return Ok(reservation),
+                Ok(Err(e)) => return Err(self.handle_ingest_error(e, stream, idx).await),
+                Err(_elapsed) => {}
             }
```
elenagaljak-db, line 217: Maybe add a comment to explain what happens if we reach the Err branch. Something like 
```
// Timed out waiting for a permit: the sub-stream is still at
// capacity. Loop to re-check liveness, back off, and keep waiting.
```

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -89,9 +110,31 @@ pub struct MultiplexedStream {
...
+            stream_index < (1 << STREAM_BITS),
+            "MultiplexedStream supports at most {} sub-streams",
+            1 << STREAM_BITS
+        );
+        Arc::new(MultiplexedAckCallbackAdapter {
+            stream_index,
+            callback,
+        })
+    }
+
```
elenagaljak-db: We should check if there is a cleaner way to do this, other than exposing this method. Something that would check if both ack_callback and multiplexing are set.

rust/sdk/src/callbacks.rs
```diff
@@ -81,6 +83,25 @@ pub trait AckCallback: Send + Sync {
...
+    /// The message ID identifies both the sub-stream and the offset within
+    /// that sub-stream.
+    fn on_ack(&self, message_id: MessageId);
+
+    /// Called when an error occurs for a specific record/batch.
+    ///
+    /// The message ID identifies both the sub-stream and the offset within
+    /// that sub-stream.
+    fn on_error(&self, message_id: MessageId, error_message: &str);
+}
```
elenagaljak-db: Do we need both this and the Adapter class (from multiplex_Stream.rs)?

rust/sdk/src/lib.rs
```diff
@@ -1990,8 +2032,39 @@ impl ZerobusStream {
...
+        &self,
+    ) -> ZerobusResult<landing_zone::CapacityReservation> {
+        self.check_open()?;
+        Ok(self.landing_zone.reserve_capacity().await)
+    }
+
+    #[cfg(feature = "testing")]
+    pub(crate) async fn ingest_lock(&self) -> tokio::sync::MutexGuard<'_, ()> {
+        self.sync_mutex.lock().await
+    }
```
elenagaljak-db: I don't think it's good practice to expose internal lock like this (even though its only between the files) as the multiplexer now holds `ZerobusStream`'s internal ordering lock across its own logic. We could do something like this:
```
#[cfg(feature = "testing")]
pub(crate) async fn ingest_admitted<F, Fut>(
    &self,
    encoded_batch: EncodedBatch,
    admit: F,                       // mux supplies the admission check
) -> ZerobusResult<OffsetId>
where
    F: FnOnce() -> Fut,
    Fut: Future<Output = ZerobusResult<()>>,
{
    self.check_open()?;
    let reservation = self.landing_zone.reserve_capacity().await;
    let _guard = self.sync_mutex.lock().await;     // never leaves this scope
    admit().await?;                                 // == admission.read() + check_closed()
    self.check_open()?;
    Ok(self.enqueue_reserved(encoded_batch, reservation))
}
```
and then the multiplexed stream side comes down to
```
// ZerobusStream — lock stays private; mux injects the middle step
#[cfg(feature = "testing")]
pub(crate) async fn enqueue_admitted<F, Fut>(
    &self,
    encoded_batch: EncodedBatch,
    reservation: CapacityReservation,
    admit: F,
) -> ZerobusResult<OffsetId>
where
    F: FnOnce() -> Fut,
    Fut: Future<Output = ZerobusResult<()>>,
{
    let _guard = self.sync_mutex.lock().await;   // never escapes this scope
    admit().await?;                               // mux: admission.read() + check_closed()
    self.check_open()?;
    Ok(self.enqueue_reserved(encoded_batch, reservation))  // now private
}
async fn enqueue_reserved(
        &self,
        stream: &ZerobusStream,
        idx: usize,
        encoded_batch: EncodedBatch,
    ) -> ZerobusResult<MessageId> {
        let reservation = self.reserve_capacity(stream, idx).await?;
        let result = stream
            .enqueue_admitted(encoded_batch, reservation, || async {
                let _admission = self.admission.read().await;
                self.check_closed()
            })
            .await;
        match result {
            Ok(off) => Ok(MessageId::new(idx, off)),
            Err(e) => Err(self.handle_ingest_error(e, stream, idx).await),
        }
    }
```
What do you think?

danilonajkov-db:
@elenagaljak-db thanks for the review, i've decided to move the ack callback to another PR to not mix contexts here. I've pasted your comments regarding the ack callback on that PR and resolved the ones here

rust/NEXT_CHANGELOG.md
danilonajkov-db: Moved the entry to the end of the Bug Fixes section in 6974603.

rust/sdk/src/landing_zone.rs
danilonajkov-db: Updated in 6974603. `CapacityReservation` is now an opaque newtype rather than a type alias, so callers cannot depend on the semaphore implementation.

rust/sdk/src/lib.rs
danilonajkov-db: Updated in 6974603. The mux no longer receives or holds `ZerobusStream`’s ordering lock. `enqueue_reserved_admitted` keeps the reservation, ingest lock, admission check, stream-open recheck, offset allocation, and enqueue inside `ZerobusStream`.

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: Added the timeout explanation in 6974603: the sub-stream is still at capacity, so the loop re-checks liveness and continues waiting.

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -140,6 +142,7 @@ impl MultiplexedStream {
             "MultiplexedStream poisoned due to sub-stream failure"
         );
 
+        let _admission = self.admission.write().await;
```
teodordelibasic-db: Not sure if we discussed this before.

Since we hold this lock across `join_all(...flush())`, an ingest already parked at `admission.read().await` in `enqueue_reserved_admitted` cannot acquire the read lock until the whole flush finishes, so it stays blocked for that window before it sees `is_closed` and errors, and because it also holds `sync_mutex`, other ingests on that sub-stream queue behind it.

The write lock maybe only needs to be held long enough to drain readers that are mid-enqueue, not for the whole flush. Could we take and immediately drop it as a barrier after setting `is_closed`, then run the flush and `signal_shutdown()` unlocked? Any reader that acquires the read lock afterward fails `check_closed()` anyway, so nothing new is admitted.

rust/sdk/src/stream/grpc/ingest.rs
```diff
@@ -211,7 +236,47 @@ impl ZerobusStream {
...
+    pub(crate) async fn reserve_capacity(
+        &self,
+    ) -> ZerobusResult<crate::landing_zone::CapacityReservation> {
+        self.check_open()?;
+        Ok(self.landing_zone.reserve_capacity().await)
+    }
+
+    #[cfg(feature = "testing")]
+    pub(crate) fn capacity_wait_timeout(&self) -> std::time::Duration {
+        std::time::Duration::from_millis(self.options.flush_timeout_ms)
```
teodordelibasic-db: Do you think it would make sense to have a separate knob for this timeout other than the flush one?

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: make sense to not block ingests if stream is mid close/failure and return error. Fixed

rust/sdk/src/stream/grpc/ingest.rs
danilonajkov-db: I added a seperate timeout variable, but didn't make it configureable for the clients. My thinking is that in the future we want the mux stream automatically try the next stream if the first one is full, so it will become an implementation detail the customer doesn't need to worry about. WDYT?

danilonajkov-db:
ill wait for https://github.com/databricks/zerobus-sdk/pull/523 to be merged so I can fix the tests

rust/sdk/src/stream/grpc/ingest.rs
```diff
@@ -100,6 +94,38 @@ impl ZerobusStream {
...
+            Ok(None)
+        } else {
+            self.enqueue_prepared_batch(encoded_batch)
+                .await
+                .map(Option::Some)
+        }
+    }
+
+    #[allow(clippy::result_large_err)]
+    pub(crate) fn prepare_record_batch(
```
teodordelibasic-db: nit: What do you think about renaming to `prepare_encoded_batch` since that's the actual name of the returned structure and also `RecordBatch` is an Arrow concept?

rust/sdk/src/stream/grpc/ingest.rs
```diff
@@ -100,6 +94,38 @@ impl ZerobusStream {
...
+                ZerobusError::InvalidArgument(
+                    "Record type does not match stream configuration".to_string(),
+                )
+            })?;
+        self.validate_ingest_payload(&encoded_batch)?;
+        Ok(encoded_batch)
+    }
+
+    #[allow(clippy::result_large_err)]
+    pub(crate) fn prepare_records_batch<I, T>(&self, payload: I) -> ZerobusResult<EncodedBatch>
```
teodordelibasic-db: Ah the comment above kind of gets debunked by this method. Still, maybe some alternative could be applied to both methods, not sure...

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -200,10 +213,15 @@ impl MultiplexedStream {
...
+    async fn reserve_capacity(
+        &self,
+        stream: &ZerobusStream,
+        idx: usize,
+    ) -> ZerobusResult<crate::landing_zone::CapacityReservation> {
         let mut backoff_ms = 1u64;
-        let mut total_wait_ms = 0u64;
         let mut logged_backpressure = false;
+        let started_at = tokio::time::Instant::now();
+        let deadline = started_at + CAPACITY_WAIT_TIMEOUT;
```
teodordelibasic-db: The capacity deadline is always `30` seconds, so a stream configured with `.flush_timeout_ms(100)` still blocks for `30` seconds, while one using the default five-minute timeout starts failing much earlier. This also differs from the intended contract that the capacity wait uses the existing `flush_timeout_ms` option. Could we derive the deadline from the selected stream instead?

```rust
let capacity_wait_timeout =
    std::time::Duration::from_millis(stream.options.flush_timeout_ms);
let deadline = started_at + capacity_wait_timeout;
```

Could we also set `flush_timeout_ms: Some(100)` in `test_ingest_times_out_when_capacity_never_recovers` and assert the timeout after advancing past `100` ms? That would cover the configured contract rather than the fixed constant.

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -217,14 +235,28 @@ impl MultiplexedStream {
...
+                return Err(ZerobusError::ConnectionTimeout(format!(
+                    "Timed out waiting for capacity on multiplexed sub-stream {}",
+                    idx
+                )));
+            }
+
+            let wait_duration = std::time::Duration::from_millis(backoff_ms)
+                .min(deadline.saturating_duration_since(now));
+
+            match tokio::time::timeout(wait_duration, stream.reserve_capacity()).await {
```
teodordelibasic-db: `tokio::time::timeout` cancels `stream.reserve_capacity()` on every `1` to `50` ms poll. Tokio's semaphore is FIFO and documents that cancelling `acquire_owned` loses the waiter's place, so concurrent ingests repeatedly move to the back of the queue and can reach the capacity timeout even while permits are being released. Could we keep one reservation future alive for the whole wait and use separate timer branches only for liveness checks and the final deadline?

```rust
let reservation = stream.reserve_capacity();
tokio::pin!(reservation);
let mut health_check = tokio::time::interval(std::time::Duration::from_millis(50));

loop {
    tokio::select! {
        result = &mut reservation => {
            return match result {
                Ok(reservation) => Ok(reservation),
                Err(error) => Err(self.handle_ingest_error(error, stream, idx).await),
            };
        }
        _ = tokio::time::sleep_until(deadline) => {
            return Err(ZerobusError::ConnectionTimeout(format!(
                "Timed out waiting for capacity on multiplexed sub-stream {idx}",
            )));
        }
        _ = health_check.tick() => {
            self.check_closed()?;
            // Keep the existing sub-stream liveness check here.
        }
    }
}
```

rust/sdk/src/stream/grpc/ingest.rs
danilonajkov-db: yep makes sense

rust/sdk/src/stream/grpc/ingest.rs
danilonajkov-db: Still I can rename to something like prepare_encoded_batch

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: I fixed it by pinning one reservation future for the entire capacity wait. The diagnostic timeout only borrows that future and does not recreate the semaphore acquisition, so its FIFO position is preserved. Stream and mux cancellation tokens now handle liveness without periodic polling.

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: flush() still behaves correctly. 
If capacity is full because record A is unacknowledged:
  1. Record B waits for capacity.
  2. flush() snapshots record A.
  3. If A remains unacknowledged past flush_timeout_ms, flush() fails.
  4. If A is acknowledged, flush() may succeed and B may enqueue afterward.

  That last case is intentional: records ingested concurrently with flush() are explicitly excluded. A capacity-waiting ingest has no offset yet, so it has not been admitted and cannot be part of the flush snapshot.

teodordelibasic-db (dismissed):
LGTM. I think the change to `rust/NEXT_CHANGELOG.md` in this PR was reverted in one of the later commits, just checking if it is intentional.

danilonajkov-db:
> LGTM. I think the change to `rust/NEXT_CHANGELOG.md` in this PR was reverted in one of the later commits, just checking if it is intentional.

Yeah I thought it wasn't needed since mux streams is not public yet, but yeah you are right still we should have it. Added it in the latest commit

teodordelibasic-db (approved):

PR #429, opened 2026-06-26
[Rust][FFI] Add darwin and windows-gnu library cross-builds via cargo-zigbuild

teodordelibasic-db:
Same as title.


elenagaljak-db (approved):
thank you 🙏

PR #431, opened 2026-06-26
[Rust][FFI] Recursively enforce required fields

### LINKED ISSUE #393: [Rust][FFI] Recursively enforce required-field presence in proto encode (nested STRUCT/ARRAY/MAP)
### SDK

Rust

### Description

The dynamic-protobuf FFI (`zerobus_proto_schema_encode_json`) validates that non-nullable (proto2 `required`) columns are present before encoding, because prost-reflect does not enforce `required` presence on encode. However, the check only walks **top-level** fields:

```rust
let missing: Vec<String> = schema_ref
    .message
    .fields()                          // top-level fields only
    .filter(|f| matches!(f.cardinality(), Cardinality::Required) && !message.has_field(f))
    ...
```

A record that omits a non-nullable field **nested inside a STRUCT** therefore encodes successfully and is only rejected later by the server (after a network round-trip), instead of failing fast and locally at encode time.

The same blind spot applies to every nested-message position the walk doesn't enter:
- Nested messages — `STRUCT` columns (required fields inside the struct).
- Repeated message elements — `ARRAY<STRUCT>` (required fields in each element).
- Map values — `MAP<K, STRUCT>` (message-typed map values).

This limitation is already documented (code comment in `proto_schema.rs`, the `encode_json` doc comment, `README.md`, and `NEXT_CHANGELOG.md`), so it is acceptable to ship the current top-level-only check and address the deeper fix as a follow-up.

Surfaced in review of the dynamic-protobuf PR (#371).

### Proposed Solution

Replace the top-level-only presence check with a **recursive presence walk** that descends into nested messages: for every `Required` field that is a message, recurse into its sub-message; for `repeated` message fields, check each element; for map fields with message values, check each value. Accumulate the full set of missing required field paths (e.g. `addr.zip`, `items[2].id`) and reject with a clear error before encoding. Add round-trip/rejection tests covering a missing required field in a STRUCT, in an `ARRAY<STRUCT>` element, and in a `MAP` value.

### Additional Context

_No response_

zlata-stefanovic-db:

**WHAT:** Make `zerobus_proto_schema_encode_json` (the dynamic-protobuf FFI) enforce proto2 `required` presence **recursively** instead of only on top-level columns.

- Replaced the top-level-only presence filter with a recursive walk (`collect_missing_required_fields` / `descend_into_messages`) that descends into present message-typed values: nested `STRUCT`s, each `ARRAY<STRUCT>` element, and `MAP` values.
- Missing required fields are reported with their full path (e.g. `addr.zip`, `items[2].id`, `props[home].zip`) and the record is rejected **before** encoding.
- The walk reuses a single path buffer (append on descent, truncate on return), so a valid record allocates no path strings; a heap `String` is produced only when an actual missing field is recorded.
- Updated the `encode_json` doc comment (which regenerates `zerobus.h` via cbindgen), `rust/ffi/README.md`, and `rust/ffi/NEXT_CHANGELOG.md`.

**WHY:** prost-reflect does not enforce `required` presence on encode, and the previous check walked only top-level fields. A record omitting a non-nullable field nested inside a `STRUCT`, an `ARRAY<STRUCT>` element, or a `MAP` value therefore encoded successfully and was only rejected by the server after a network round-trip. Validating recursively fails fast and locally with a precise field path. Recursion depth is bounded by the schema's build-time `MAX_NESTING_DEPTH`, so it cannot overflow the stack, and there is no FFI signature change (only the doc comment in `zerobus.h`).

Closes #393.

## How is this tested?

Added unit tests in `rust/ffi/src/tests.rs` covering a missing required field in:
- a `STRUCT` (asserts the error names `addr.zip`),
- an `ARRAY<STRUCT>` element (asserts `items[1].id`),
- a `MAP` value (asserts `lookup[work].v`),
- a deeply nested / multi-level case (asserts `addr.geo.lat` three message-levels deep, and `items[0].inner.id` through an array element's nested struct),

plus a positive round-trip test with all nested required fields present. The full `zerobus-ffi` suite passes (`cargo test -p zerobus-ffi`, 57 tests), with `cargo clippy` and `rustfmt` clean.


rust/ffi/src/proto_schema.rs
```diff
@@ -139,6 +141,110 @@ pub extern "C" fn zerobus_proto_schema_descriptor_bytes(
     schema_ref.descriptor_bytes.as_ptr()
 }
 
+/// Recursively collect the paths of `Required` (proto2) fields absent from
```
irinatomic-db: Comment for the whole PR: Can we simplify these really long comments? There's no need to explain all the implementation details for a private function.  This one specifically could be simplified like:
> Recursively collect full paths of absent proto2 `required` fields in `message`, descending into nested structs, list elements, and map values. `path` is a scratch buffer; each frame appends its segment and truncates on return.

rust/ffi/src/proto_schema.rs
```diff
@@ -139,6 +141,110 @@ pub extern "C" fn zerobus_proto_schema_descriptor_bytes(
...
+        // Append `.<field>` (or just `<field>` at the root), remembering where to
+        // cut back to so the buffer is restored for the next sibling.
+        let base = path.len();
+        if !path.is_empty() {
+            path.push('.');
+        }
+        path.push_str(field.name());
+
+        if matches!(field.cardinality(), Cardinality::Required) && !message.has_field(&field) {
+            // Absent required field: record it; there is nothing to descend into.
```
irinatomic-db: Not needed comment

rust/ffi/src/proto_schema.rs
```diff
@@ -139,6 +141,110 @@ pub extern "C" fn zerobus_proto_schema_descriptor_bytes(
...
+            missing.push(path.clone());
+        } else if matches!(field.kind(), Kind::Message(_)) && message.has_field(&field) {
+            // Descend into present message-typed values — singular structs, array
+            // elements, and map values — to validate their required fields too.
+            // Scalar fields (including arrays/maps of scalars) have no nested
+            // requirements, and an absent message field imposes none.
```
irinatomic-db: Same here. Not needed

rust/ffi/src/proto_schema.rs
```diff
@@ -139,6 +141,110 @@ pub extern "C" fn zerobus_proto_schema_descriptor_bytes(
...
+/// Recursion depth is bounded by the descriptor depth, itself capped at
+/// schema-build time (`MAX_NESTING_DEPTH` in `type_json` parsing), so a
+/// pathological record cannot blow the stack here.
+fn collect_missing_required_fields(
+    message: &DynamicMessage,
+    path: &mut String,
+    missing: &mut Vec<String>,
+) {
+    for field in message.descriptor().fields() {
+        // Append `.<field>` (or just `<field>` at the root), remembering where to
```
irinatomic-db: Comment worth keeping, but simplified, like "save position so we can restore the buffer after this field"

rust/ffi/src/proto_schema.rs
```diff
@@ -139,6 +141,110 @@ pub extern "C" fn zerobus_proto_schema_descriptor_bytes(
...
+        // cut back to so the buffer is restored for the next sibling.
+        let base = path.len();
```
irinatomic-db: Also, can we rename this to "base_len"? Base sounds too vague,

irinatomic-db (commented):
Can you go through all the changes and simplify the comments/docs? I know Claude likes to write long descriptions but there's no need here, the logic/change is very clear by itself.

rust/ffi/src/proto_schema.rs
zlata-stefanovic-db: Done

rust/ffi/src/proto_schema.rs
zlata-stefanovic-db: Deleted

rust/ffi/src/proto_schema.rs
zlata-stefanovic-db: Deleted

rust/ffi/src/proto_schema.rs
zlata-stefanovic-db: Done

rust/ffi/src/proto_schema.rs
zlata-stefanovic-db: Done

elenagaljak-db (approved):

PR #432, opened 2026-06-26
[Rust] Guard FFI entry points with catch_unwind

### LINKED ISSUE #397: [Rust][FFI] Wrap all extern "C" bodies in catch_unwind panic guards (crate-wide)
### SDK

Rust

### Description

None of the FFI `extern "C"` functions in `rust/ffi/src` wrap their bodies in `std::panic::catch_unwind`. If any of them panics (e.g. an `unwrap`/`expect` deep in a dependency, an allocation failure, or a slicing/bounds error), the panic unwinds **across the FFI boundary into the C caller** (Go via cgo, Java via JNI). Unwinding out of an `extern "C"` frame into non-Rust code is undefined behavior — at best an abort, at worst memory corruption — and it gives the caller no recoverable error.

This is **preexisting** and applies to **every** FFI entry point, not just the new dynamic-protobuf functions. Surfaced in review of #371 by @teodordelibasic-db:

> none of the FFI functions wrap their bodies in `catch_unwind`. Worth a follow-up potentially to add panic guards crate-wide.

### Proposed Solution

Wrap every `#[no_mangle] pub extern "C"` body in a `catch_unwind` guard so a panic is caught at the boundary and converted into the function's normal error channel instead of unwinding into C:

- For functions that take a `*mut CResult`, write a non-retryable error into it (`success = false`, `error_message` set, `is_retryable = false`) and return the function's failure sentinel.
- Use the correct per-signature sentinel on the caught-panic path: `ptr::null_mut()` / `ptr::null()` for pointer returns, `false` for `bool`, `-1` for the `i64` offset functions, etc.
- Prefer a single shared helper (a small wrapper fn or macro) applied uniformly across the crate so the pattern is consistent and hard to forget on new functions, rather than hand-writing `catch_unwind` in each function.
- Note that `catch_unwind` requires the closure to be `UnwindSafe`; raw pointers are, but any captured `&mut` or interior-mutable state may need `AssertUnwindSafe` with a brief justification.

Add tests that force a panic in a guarded function (e.g. via a deliberately panicking `HeadersProvider` callback or malformed input) and assert it returns the failure sentinel + a populated `CResult` rather than aborting.

### Additional Context

Follow-up from #371; tracked separately from that PR since it is preexisting and crate-wide.

zlata-stefanovic-db:
## Summary

Wraps every panic-capable `#[no_mangle] pub extern "C"` body in the `rust/ffi` crate in a single shared `ffi_guard` helper that runs the body under `catch_unwind`. Previously a panic anywhere in an FFI function body (a dependency `unwrap`/`expect`, an allocation failure, a slicing/bounds error) would escape across the `extern "C"` boundary: on current Rust toolchains (1.81+) this aborts the host process, and on pre-1.81 toolchains it was undefined behavior. Either way the C/Go/Java caller loses the process and gets no recoverable error.

A caught panic is now converted into the function's normal failure channel:
- A non-retryable error is written to the `CResult` out-parameter (when present).
- The per-signature failure sentinel is returned: `NULL` for pointer returns, `false` for `bool` (except `zerobus_arrow_stream_is_closed`, which returns `true` to mean "treat as closed"), `-1` for the `i64` offset functions, an empty array struct for the `get_unacked_*` functions, `()` for the `void` finalizers.

The guard relies on the `panic = "unwind"` strategy, pinned in the workspace release profile (`[profile.release]`) so a release build can't silently turn the `catch_unwind` guards into aborts. **No signatures change and `zerobus.h` is byte-identical**, so this is ABI-compatible for existing Go/Java consumers.

Resolves #397 (follow-up from the #371 review).

## Details

- **Shared helper** (`common.rs`): `ffi_guard(result, sentinel, body)` runs `body` inside `catch_unwind(AssertUnwindSafe(..))`; on a caught panic it populates the `CResult` (no-op if null) and returns the sentinel. `AssertUnwindSafe` is required because `ffi_guard` is generic over an arbitrary `FnOnce` with no `UnwindSafe` bound, so the compiler can't prove unwind-safety at the helper layer. The assertion is sound: `UnwindSafe` only guards against *observing* logically-inconsistent state after a catch (never memory unsafety), and each guarded call abandons its operation and returns the sentinel rather than resuming partially-updated state.
- **Mutex poison recovery** (`intern_header_key`): because panics are now caught instead of aborting, a panic taken while the header-key cache lock is held would otherwise poison it for the rest of the process and permanently break header interning for all custom-headers-provider streams. The lock now recovers via `lock().unwrap_or_else(|e| e.into_inner())`; the interned `&'static str` keys stay valid across a panic, so this is sound. (This `Mutex<HashSet>` is the only `std::sync::Mutex` the crate owns.)
- **Intentionally unguarded**: the two `*_get_default_config` getters and the no-op `zerobus_sdk_set_use_tls` — their bodies construct a `#[repr(C)]` struct from compile-time constants (or do nothing) and have no panic-capable operation. Noted inline.

## Test plan

- Added tests covering each sentinel shape (pointer, `bool`, `i64`, struct-return, the no-`CResult` path), `&str`/`String` panic-payload messages, the success passthrough, and poisoned-lock recovery for `intern_header_key`.
- `cargo test -p zerobus-ffi` → 67 passed.
- `cargo fmt --check` and `cargo clippy` clean.
- Confirmed `rust/ffi/zerobus.h` regenerates byte-identical (no ABI change).
- `rust/ffi/NEXT_CHANGELOG.md` updated under Bug Fixes.

> Reviewer tip: the line-level diff looks large because wrapping each body in a closure re-indents it; view with whitespace ignored (`?w=1`) to see the ~370-line real change.


rust/ffi/NEXT_CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 
 ### Bug Fixes
 
+- Every `#[no_mangle] extern "C"` entry point now runs its body inside a `catch_unwind` panic guard (`ffi_guard`). Previously a panic anywhere in an FFI function body (a dependency `unwrap`/`expect`, an allocation failure, a slicing/bounds error, or a panicking callback) would escape across the `extern "C"` boundary, aborting the host process (on current Rust toolchains; it was undefined behavior on pre-1.81 ones) with no recoverable error returned to the C/Go/Java caller. A caught panic is now converted into the function's normal failure channel: a non-retryable error is written to the `CResult` out-parameter (when present) and the per-signature failure sentinel is returned (`NULL` for pointer returns, `false` for `bool`, `-1` for the `i64` offset functions, an empty array struct for the `get_unacked_*` functions). The guard relies on the default `panic = "unwind"` strategy. No signatures changed and `zerobus.h` is byte-identical, so this is ABI-compatible for existing Go/Java consumers.
```
teodordelibasic-db: There may be a few accuracy issues in this entry:
- It lists "a panicking callback" as something now converted to a `CResult`, but the headers provider callback is an `extern "C" fn` called directly from `CallbackHeadersProvider::get_headers`. A Rust panic out of that callback aborts at the callback's own boundary before `ffi_guard` can catch it, so callback panics are not recoverable. We can drop "or a panicking callback", or state explicitly that caller callbacks must not unwind across the ABI.
- "Every `#[no_mangle] extern "C"` entry point" is slightly over-broad, `zerobus_sdk_set_use_tls` and the two `_get_default_config` getters are intentionally unguarded. Worth narrowing to "every panic-capable entry point" with the three exceptions noted.
- The sentinel list says `false` for bool but omits that `zerobus_arrow_stream_is_closed` returns `true` (treat-as-closed) and that void functions return `()`. The PR description has these right.

rust/ffi/src/builder.rs
teodordelibasic-db, line 48: `mem::take` moves the builder out before `f` runs, so a panic inside the setter closure leaves the slot holding a default builder and drops the in-flight config during unwind. The doc comment is right that the slot stays valid, but since these setters return `void` with no `CResult`, the caller has no way to know configuration was lost and keeps calling setters on a silently-reset builder.

We can consider restoring `taken` on the panic path (a small scope guard, or build the result before swapping it in), or document that a caught panic from a setter invalidates the builder and it should be discarded.

rust/ffi/src/common.rs
```diff
@@ -303,3 +311,71 @@ pub(crate) fn write_success_result(result: *mut CResult) {
...
+/// a non-retryable error to `result` (a no-op if `result` is null) and returns
+/// `sentinel`, the function's usual failure value (null pointer, `false`, `-1`,
+/// an empty array struct, or `()`); pass a null `result` for functions with no
+/// `CResult` out-parameter. Relies on the default `panic = "unwind"` strategy.
+///
+/// `AssertUnwindSafe` is needed because the body is an arbitrary `FnOnce` with
+/// no `UnwindSafe` bound; it is sound because a caught panic abandons the call
+/// rather than resuming partially-updated state. The one shared-state caveat, a
+/// poisoned `Mutex`, is handled at the only lock the crate owns (see
+/// `intern_header_key`). See PR #432 for the full rationale.
```
irinatomic-db: For the whole PR: let's revise these doc comments. Not as long and implementation heavy. And let's not reference PRs, they don't mean anything to the reader.

A doc comment should say what the function does and what the caller should pass; not how it works internally

rust/ffi/src/common.rs
```diff
@@ -49,7 +51,13 @@ static HEADER_KEY_CACHE: Lazy<Mutex<HashSet<&'static str>>> =
 /// Intern a header key string to prevent memory leaks
 /// Only leaks memory for unique keys, not on every call
 pub(crate) fn intern_header_key(key: String) -> &'static str {
-    let mut cache = HEADER_KEY_CACHE.lock().unwrap();
+    // Recover from a poisoned lock rather than re-panicking: with `ffi_guard`
+    // catching panics instead of aborting, a panic under this lock would
+    // otherwise poison it and break interning for the rest of the process. The
+    // interned `&'static str` keys stay valid across a panic, so this is sound.
```
irinatomic-db: Let's also shorten this comment

rust/ffi/src/tests.rs
```diff
@@ -1065,4 +1065,142 @@ mod tests {
...
+    //
+    // Every `#[no_mangle] extern "C"` entry point runs its body through
+    // `ffi_guard`, so these tests exercise the shared mechanism that protects
+    // the whole crate: a panic inside the body must be caught at the boundary
+    // and turned into (failure sentinel + populated CResult) rather than
+    // escaping across the `extern "C"` boundary (which aborts the process on
+    // current Rust toolchains, and was undefined behavior on pre-1.81 ones).
+    // They cover each return-signature shape and the no-CResult path.
+
+    // Read a CResult, free its message, and return (success, is_retryable, msg).
```
irinatomic-db, line 1080: The comment says these tests exercise "the shared mechanism that protects the whole crate," but they only call `ffi_guard` directly with synthetic closures. No test drives an actual `#[no_mangle]` entry point into a panic. If a function is added later without the guard, nothing catches it. Consider one integration-style test that panics through a real entry point and asserts the sentinel + clean `CResult`.

irinatomic-db (commented):
Not sure where to put this comment:

 panic = "unwind" is the load-bearing invariant for every guard in this PR, but nothing enforces it. The FFI crate ships as a `staticlib/cdylib`, so the active panic strategy is set by whoever links it - a `panic = "abort"` in any release profile silently turns every `catch_unwind` into a no-op. Let's add to Cargo.toml:
```
[profile.release]
panic = "unwind"
```

rust/ffi/src/tests.rs
zlata-stefanovic-db: Done

rust/ffi/src/common.rs
teodordelibasic-db, line 290: If a guarded body populates `result` via `write_error_result` and then panics before returning, the guard's panic arm overwrites `result.error_message` here without freeing the previous `CString`, a one-time small leak on the panic path. Cheap to close: free a non-null `(*result).error_message` before overwriting.

rust/ffi/src/arrow.rs
```diff
@@ -478,124 +496,145 @@ pub extern "C" fn zerobus_arrow_stream_get_unacked_batches(
...
-                                // so capacity == len.
-                                unsafe {
-                                    let _ =
-                                        Box::from_raw(std::ptr::slice_from_raw_parts_mut(ptr, len));
+
+                    let count = batches.len();
+                    let mut batch_ptrs: Vec<*mut u8> = Vec::with_capacity(count);
+                    let mut batch_lens: Vec<usize> = Vec::with_capacity(count);
+
+                    for batch in &batches {
```
teodordelibasic-db, line 538: The explicit `Err` arm frees the already-converted batches, but the unwind path doesn't: if a panic (not an `Err`) is taken in this loop, `ffi_guard` returns the empty sentinel and the raw pointers already pushed into `batch_ptrs` leak (dropping the `Vec<*mut u8>` doesn't free the slices they point at).

We can either collect owning `Box<[u8]>` and only convert to raw pointers after the loop can no longer panic.

teodordelibasic-db (commented):
Looks good, only a few small fixes flagged by LLM left.

rust/ffi/NEXT_CHANGELOG.md
zlata-stefanovic-db: All done

rust/ffi/src/builder.rs
zlata-stefanovic-db: You're right. I went with the "document it" option rather than restoring, because taken is moved into f and consumed during the unwind, so there's genuinely nothing left to put back. Updated the with_builder doc with more clarity now -> to state that a caught setter panic keeps the slot valid (no UAF) but drops the in-flight config, so a builder that survives a caught panic must be discarded via _free, not reused. Let me know if you'd prefer we make the setters fallible/CResult-returning instead - that'd be the only way to actually signal the loss, but it's an ABI change.

rust/ffi/src/common.rs
zlata-stefanovic-db: Fixed

rust/ffi/src/arrow.rs
zlata-stefanovic-db: Done, took the owning-Box<[u8]> approach.

teodordelibasic-db (approved):
LGTM

PR #434, opened 2026-06-29
[Go] Add generated zerobuspb gRPC/proto bindings

zlata-stefanovic-db:

**WHAT:** Adds the internal `go/internal/zerobuspb` package containing the
generated protobuf message types and gRPC `ZerobusClient` for the Zerobus
service, generated from `zerobus_service.proto`. The service is a single
bidirectional RPC, `EphemeralStream(stream Request) returns (stream Response)`.

This is purely foundational scaffolding — **no public API changes**, no behavior
changes. It is the first step toward a **pure-Go SDK** that talks to the Zerobus
service directly over gRPC, replacing the current cgo + Rust FFI implementation.

Breakdown of the diff (1,275 insertions, but only ~102 are hand-authored):

| Category | Lines | Notes |
|---|---:|---|
| Generated `*.pb.go` | 995 | `go generate ./...`; marked `linguist-generated` |
| Vendored `duration.proto` | 117 | verbatim well-known type; marked `linguist-vendored` |
| `go.mod` / `go.sum` | 51 | add `google.golang.org/grpc` + `protobuf` |
| Hand-written (`.proto` + `gen.go`) | 102 | **the real review surface** |

`.gitattributes` marks the generated and vendored files so GitHub collapses them
in the diff — please focus review on `zerobus_service.proto` and `gen.go`.

**WHY:** The whole SDK is a gRPC client, so every layer above (transport, OAuth,
the ingest/ack state machine, recovery, and the public API) is built on these
generated types — nothing can be written without them. Landing the bindings
first as a small, self-contained PR keeps the subsequent layer-by-layer PRs
small and reviewable. `duration.proto` is vendored under `third_party/` so
regeneration (`go generate ./...`) doesn't depend on a system-wide protobuf
include; its import is remapped to the canonical `durationpb`.

> Note: this is the start of an architectural shift (pure-Go vs. the shared Rust
> core for Go). Flagging for visibility before more layers stack on top.

## How is this tested?

No unit tests yet — this PR only adds generated bindings and module setup.
Validated locally:

- `go build ./internal/...` — passes
- `go vet ./internal/...` — clean
- `go generate ./internal/zerobuspb/` reproduces the committed `*.pb.go` (the
  directive is reproducible)
- `isaac review` on the branch — 0 findings

Functional tests (transport round-trip, ingest, recovery) come with the layers
that consume these bindings in follow-up PRs.


go/go.mod
```diff
@@ -1,5 +1,17 @@
 module github.com/databricks/zerobus-sdk/go
 
-go 1.21
+go 1.25.0
```
teodordelibasic-db: This bumps the module's Go floor from 1.21 to 1.25.0. I think this can break client code. cc: @elenagaljak-db 

Apparently this comes from the `grpc` library so we can either use an older version of that which doesn't require bumping the Go version here, or see if bumping here is okay.

go/NEXT_CHANGELOG.md
```diff
@@ -12,4 +12,8 @@
 
 ### Internal Changes
 
+- Add internal `zerobuspb` package with generated gRPC/protobuf bindings for the
```
teodordelibasic-db: The `no public API changes yet` part is debatable due to the Go version bump.

go/internal/zerobuspb/gen.go
```diff
@@ -0,0 +1,20 @@
...
+//
+// It is internal to the pure-Go SDK: the wire types and the generated
+// ZerobusClient (a bidirectional EphemeralStream RPC) are implementation
+// details, not part of the SDK's public API.
+//
+// To regenerate after editing zerobus_service.proto, run `go generate ./...`
+// from the go/ module root. This requires protoc on PATH plus the
+// protoc-gen-go and protoc-gen-go-grpc plugins:
+//
+//	go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
```
teodordelibasic-db: Since we commit the generated output, using `@latest` means `go generate`
only matches the committed files until the next plugin release. Maybe we should pin the versions
that produced the current code: `protoc-gen-go@v1.36.11`, `protoc-gen-go-grpc@v1.6.2`
(and mention the expected `protoc v6.33.0`).

go/internal/zerobuspb/zerobus_service.proto
```diff
@@ -0,0 +1,82 @@
+syntax = "proto2";
+
+import "google/protobuf/duration.proto";
+
+package databricks.zerobus;
+
+// Internal to the pure-Go SDK. The generated Go bindings live in
+// github.com/databricks/zerobus-sdk/go/internal/zerobuspb and are not part of
+// the SDK's public API surface (enforced by Go's internal/ package rule).
+option go_package = "github.com/databricks/zerobus-sdk/go/internal/zerobuspb";
```
teodordelibasic-db: Not sure if this Go specific option maybe blocks that, but since we are introducing one more `zerobus_service.proto` into the repo, we should look into unifying to only one and referencing it from all the required places

teodordelibasic-db (commented):
Looks good, we should only decide regarding the Go version.

go/NEXT_CHANGELOG.md
zlata-stefanovic-db: Deleting that part

go/internal/zerobuspb/zerobus_service.proto
zlata-stefanovic-db: Yes, there's already one in each of the rust, java & python SDKs, will make a follow up for this

zlata-stefanovic-db: Filled an issue for this: https://github.com/databricks/zerobus-sdk/issues/435

go/internal/zerobuspb/gen.go
zlata-stefanovic-db: Done

purego/README.md
```diff
@@ -0,0 +1,27 @@
+# Zerobus pure-Go SDK (work in progress)
+
+A native, pure-Go implementation of the Zerobus ingestion SDK that talks to the
+Zerobus gRPC service directly — **no cgo and no Rust FFI**.
```
elenagaljak-db: I don't think we need this "no cgo..."

elenagaljak-db (commented):
Do you know if these files (the `zerobus_service.pb.go` etc) can be dynamically compiled? For example in Rust we don't have these files but the [build.rs](https://github.com/databricks/zerobus-sdk/blob/main/rust/sdk/build.rs) builds them at runtime.

elenagaljak-db:
Also I think this go should be from version 2.0.0. since it's completely different core.

zlata-stefanovic-db:
> Do you know if these files (the `zerobus_service.pb.go` etc) can be dynamically compiled? For example in Rust we don't have these files but the [build.rs](https://github.com/databricks/zerobus-sdk/blob/main/rust/sdk/build.rs) builds them at runtime.

Not automatic functions like build.rs, but there is "go generate" that can be manually ran, however, the generated files still have to be committed - go generate is a manual maintainer step, not part of go build/go get.

zlata-stefanovic-db:
> Also I think this go should be from version 2.0.0. since it's completely different core.

Do you think we should version it now as an isolated project - starting from like 0.1.0 (as it is now) and when it's proven to be able to replace the cgo to make the versioning 2.0.0? @elenagaljak-db @teodordelibasic-db

elenagaljak-db:
Yes, let's start with 0.1.0

elenagaljak-db (approved):
Good start :)

PR #456, opened 2026-06-30
[Java][Python] Default recovery_retries to 4 (#438)

### LINKED ISSUE #438: [Java][Python] recovery_retries default is 3, should be 4 (matches core/Go/TS/C++)
## Summary

The **Java** and **Python** SDKs default `recovery_retries` to **3**, while the Rust core, the C FFI, and every other wrapper (Go, TypeScript, C++) use **4**. This means a stream that hits transient failures gives up after 3 recovery attempts on Java/Python but 4 everywhere else — an unintended cross-SDK behavioral divergence with no documented rationale.

## Evidence

| SDK | `recovery_retries` default | Source |
|---|---|---|
| **Rust core** (ground truth) | **4** | `rust/sdk/src/stream_options.rs:15` — `pub const RECOVERY_RETRIES: u32 = 4;` |
| C FFI | 4 | `zerobus_get_default_config()` → `defaults::RECOVERY_RETRIES` |
| Go | 4 | `go/types.go:73` |
| TypeScript | 4 | seeds from core defaults (`typescript/src/lib.rs` — unset → core default) |
| C++ | 4 | `cpp/include/zerobus/config.hpp` (`recovery_retries = 4`) |
| **Java** | **3** ❌ | `java/src/main/java/com/databricks/zerobus/StreamConfigurationOptions.java:35` — `private int recoveryRetries = 3;` |
| **Python** | **3** ❌ | `python/rust/src/common.rs:330` — `recovery_retries: 3` |

## Root cause (history)

- The core constant has been `4` since the **initial commit** (`fb17329`); it was never `3`.
- Java's `recoveryRetries = 3` was introduced unchanged when the Java SDK was moved into the monorepo (`4614bfa`, "[Monorepo] Move the Java SDK to the Rust SDK repo").
- Python's `recovery_retries: 3` likewise arrived with the Python move (`9db106d`).
- No comment, changelog entry, or commit message justifies `3`. The Java Javadoc (`StreamConfigurationOptions.java:185`) merely documents the value as `3`, reflecting the stale default rather than a deliberate choice.

Conclusion: Java and Python carried their own pre-monorepo default and were never reconciled to the core's `4`. This is a stale-value bug, not an intentional per-SDK tuning.

## Impact

A user who does not override `recovery_retries` gets **one fewer recovery attempt** on Java/Python than on Rust/Go/TS/C++ before the stream fails. For workloads relying on automatic recovery through transient gRPC/network blips, Java/Python streams are marginally less resilient than the rest, with no way for the user to know the default differs.

## Proposed fix

Change the default to **4** in both wrappers, matching the core and all other SDKs:

- Java: `StreamConfigurationOptions.java:35` → `private int recoveryRetries = 4;` (and update the builder's `defaultOptions`-derived field and the Javadoc at line 185).
- Python: `python/rust/src/common.rs:330` → `recovery_retries: 4` (and any docstring/repr that states `3`).

Better still, where feasible, **seed wrapper defaults from the core** (as TypeScript does) rather than re-hardcoding them, so this class of drift can't recur.

## Considerations

- This is a user-visible default change (streams will attempt 4 recoveries instead of 3 when the option is left unset). It aligns Java/Python with the documented core intent, so it's a convergence rather than a surprise, but it should land with appropriate changelog notes and version bumps per the repo's release policy.
- Callers who explicitly set `recovery_retries` are unaffected.
- Worth a quick audit of the *other* default fields in the Java/Python option structs against the core, in case `recovery_retries` isn't the only value that drifted during the monorepo moves.

## Acceptance criteria

- Java and Python default `recovery_retries` to `4`.
- Java Javadoc / Python docs/repr updated to reflect `4`.
- Changelog entries added for Java and Python per repo policy.
- (Optional) Java/Python default-field audit against the core confirms no other drift.

## References

- Core default: `rust/sdk/src/stream_options.rs:15`
- Java: `java/src/main/java/com/databricks/zerobus/StreamConfigurationOptions.java:35` (and `:185`)
- Python: `python/rust/src/common.rs:330`
- For-reference (already correct): `go/types.go:73`, `cpp/include/zerobus/config.hpp`, `typescript/src/lib.rs`

zlata-stefanovic-db:
## Summary

Fixes #438.

The **Java** and **Python** SDKs defaulted `recovery_retries` to **3**, while the Rust core and every other wrapper (Go, TypeScript, C++) use **4**. A stream left with the default gave up after 3 recovery attempts on Java/Python but 4 everywhere else — an unintended cross-SDK divergence that arrived when those SDKs moved into the monorepo carrying their pre-existing default. No comment, changelog, or commit message ever justified `3`.

This aligns both wrappers with the core's documented intent.

| SDK | before | after |
|---|---|---|
| Rust core (ground truth) | 4 | 4 |
| Go / TypeScript / C++ | 4 | 4 |
| **Java** | **3** ❌ | **4** ✅ |
| **Python** | **3** ❌ | **4** ✅ |

## Changes

**Behavior (default 3 → 4):**
- Java: `StreamConfigurationOptions.java` field (the builder derives from it via `getDefault()`).
- Python: `python/rust/src/common.rs` `Default` impl.

**Docs / stub / test kept in sync:**
- Java Javadoc default list + the default-values unit test (`StreamConfigurationOptionsTest`).
- Python type stub `_zerobus_core.pyi` (attribute doc, `__init__` signature, Args doc), `config.py` docstring, `README.md` options table.
- The four `python/examples/*.py` that explicitly passed `recovery_retries=3` (consistency only — explicit overrides, not the default).

**Changelogs:** Bug Fixes entries in `java/NEXT_CHANGELOG.md` and `python/NEXT_CHANGELOG.md`.

## Notes

- **User-visible default change:** streams attempt 4 recoveries instead of 3 when the option is left unset. It's a convergence to the documented core intent, not a surprise. Callers that set `recovery_retries` explicitly are unaffected.
- **Drift audit (per the issue's optional ask):** every other default field in both option structs was checked against the core — `recovery_retries` was the only drift. Python's Arrow config already used `4` and is untouched.
- **JNI path:** `rust/jni/src/options.rs` reads `recoveryRetries` from the Java object (no hardcoded default), so the Java field fix fully covers it.
- Reviewed with `llm review`: 0 findings.

### DIFF
--- a/java/NEXT_CHANGELOG.md
+++ b/java/NEXT_CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Bug Fixes
 
+- Fixed the default `recoveryRetries`, which was `3` instead of the `4` used by the Rust core and every other SDK (Go, TypeScript, C++). A stream left with the default now makes 4 recovery attempts on transient failures instead of 3, matching the documented cross-SDK behavior. Callers that set `recoveryRetries` explicitly are unaffected. (#438)
+
 ### Documentation
 
 - Reworked README and Javadoc to steer users toward the high-throughput ingestion pattern: ingest with `ingestRecordOffset()` in a loop without waiting, then `flush()` (or `waitForOffset()` on the last offset) once. Added a performance callout warning against waiting for an acknowledgment after every record, documented `AckCallback` as the non-blocking alternative, and rewrote the Quick Start example to use the offset-based API instead of the deprecated `ingestRecord().join()` loop.
--- a/java/src/main/java/com/databricks/zerobus/StreamConfigurationOptions.java
+++ b/java/src/main/java/com/databricks/zerobus/StreamConfigurationOptions.java
@@ -32,7 +32,7 @@ public class StreamConfigurationOptions {
   private boolean recovery = true;
   private int recoveryTimeoutMs = 15000;
   private int recoveryBackoffMs = 2000;
-  private int recoveryRetries = 3;
+  private int recoveryRetries = 4;
   private int flushTimeoutMs = 300000;
   private int serverLackOfAckTimeoutMs = 60000;
   private Optional<Consumer<IngestRecordResponse>> ackCallback = Optional.empty();
@@ -182,7 +182,7 @@ public class StreamConfigurationOptions {
    *   <li>recovery: true
    *   <li>recoveryTimeoutMs: 15000
    *   <li>recoveryBackoffMs: 2000
-   *   <li>recoveryRetries: 3
+   *   <li>recoveryRetries: 4
    *   <li>flushTimeoutMs: 300000
    *   <li>serverLackOfAckTimeoutMs: 60000
    *   <li>ackCallback: empty
--- a/java/src/test/java/com/databricks/zerobus/StreamConfigurationOptionsTest.java
+++ b/java/src/test/java/com/databricks/zerobus/StreamConfigurationOptionsTest.java
@@ -21,7 +21,7 @@ public class StreamConfigurationOptionsTest {
     assertTrue(options.recovery());
     assertEquals(15000, options.recoveryTimeoutMs());
     assertEquals(2000, options.recoveryBackoffMs());
-    assertEquals(3, options.recoveryRetries());
+    assertEquals(4, options.recoveryRetries());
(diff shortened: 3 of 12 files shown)


elenagaljak-db (approved):
Thanks for fixing it

PR #457, opened 2026-06-30
[Rust] Add mux acknowledgment callback bridge

danilonajkov-db:

This PR implements the acknowledgment callback bridge required by multiplexed streams.

- Generalize `AckCallback` over its identifier type while preserving `OffsetId` as the default, so existing individual-stream callback implementations remain unchanged.
- Add a crate-private multiplexed callback adapter that captures one sub-stream index.
- Convert each acknowledged stream-local `OffsetId` into the `MessageId` exposed by `MultiplexedStream`.
- Forward both successful acknowledgments and errors through `AckCallback<MessageId>`.
- Validate that callback adapters are only created for supported sub-stream indices.

The multiplexed stream builder is being added separately in #433. Its only callback-specific responsibility will be to install one adapter per sub-stream:

```rust
let callback = multiplexed_ack_callback(stream_index, callback.clone());
grpc_config.ack_callback = Some(callback);
```

This gives multiplexed streams the same user-facing callback API as individual streams, except the identifier is `MessageId` instead of `OffsetId`:

```rust
impl AckCallback<MessageId> for MyCallback {
    fn on_ack(&self, message_id: MessageId) {
        println!("Acknowledged {message_id}");
    }

    fn on_error(&self, message_id: MessageId, error_message: &str) {
        eprintln!("Error for {message_id}: {error_message}");
    }
}
```

Each adapter already knows which sub-stream it belongs to, so callback delivery requires no lookup, routing table, or locking. It only packs the captured stream index and acknowledged offset into a `MessageId` before invoking the user callback.

## How is this tested?

- Verifies callbacks from different sub-streams produce different `MessageId` values for the same stream-local offset.
- Verifies errors preserve both the correct `MessageId` and error message.
- Verifies invalid sub-stream indices are rejected.
- `cargo test -p databricks-zerobus-ingest-sdk`
- `cargo fmt --all --check`

### DIFF
--- a/rust/sdk/src/callbacks.rs
+++ b/rust/sdk/src/callbacks.rs
@@ -26,7 +26,9 @@ use crate::offset_generator::OffsetId;
 /// Callback trait for receiving acknowledgment notifications.
 ///
 /// Implement this trait to receive callbacks when records/batches are acknowledged
-/// by the server or when errors occur.
+/// by the server or when errors occur. Single streams use the default [`OffsetId`]
+/// identifier. Other stream types can select an identifier that matches their
+/// ingestion API.
 ///
 /// # Thread Safety and Performance
 ///
@@ -58,7 +60,7 @@ use crate::offset_generator::OffsetId;
 ///     }
 /// }
 /// ```
-pub trait AckCallback: Send + Sync {
+pub trait AckCallback<Id = OffsetId>: Send + Sync {
     /// Called when a record/batch is successfully acknowledged by the server.
     ///
     /// **Note**: This runs synchronously in a dedicated callback handler task.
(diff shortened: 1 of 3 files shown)


danilonajkov-db:
@elenagaljak-db's comments from previous PR
image

danilonajkov-db:
@elenagaljak-db this should be much cleaner now, we have the adapter/bridge to translate the message_id in mux streams to the offset id in individual streams and made the ack callback IDs generic (so we can reuse the same one for both mux and normal streams)

rust/sdk/src/callbacks.rs
```diff
@@ -76,11 +78,12 @@ pub trait AckCallback: Send + Sync {
...
     /// # Parameters
     ///
-    /// * `offset_id` - The logical offset ID that encountered an error
+    /// * `id` - The identifier that encountered an error
     /// * `error_message` - Human-readable error description
-    fn on_error(&self, offset_id: OffsetId, error_message: &str);
+    fn on_error(&self, id: Id, error_message: &str);
 }
 
+#[allow(dead_code)]
```
teodordelibasic-db, line 86: nit: I don't think this is needed.

teodordelibasic-db (approved):
LGTM

rust/sdk/src/callbacks.rs
danilonajkov-db: ill remove it anyway when I integrate this in the stream builder

PR #458, opened 2026-06-30
[CI] Run CI on external fork PRs via approved-for-ci-run label

elenagaljak-db:
Fork PRs can't run CI: GitHub denies them the OIDC token, so the JFrog auth step fails (ACTIONS_ID_TOKEN_REQUEST_TOKEN unbound).

This adds a pull_request_target workflow: a maintainer (write access)reviews the fork code and adds the `approved-for-ci-run` label, whichmirrors the PR into an internal `ci-run/pr-<N>` branch + PR. As a same-repo PR it gets OIDC, so JFrog auth and full CI run normally.
The label is auto-removed so each push needs re-approval.

Requires: `approved-for-ci-run` label, `CI_APP_ID` + `CI_APP_PRIVATE_KEY` secrets, and PR-write on the CI GitHub App.

### DIFF
--- a/.github/workflows/approved-for-ci-run.yml
+++ b/.github/workflows/approved-for-ci-run.yml
@@ -1,0 +1,154 @@
+name: Handle `approved-for-ci-run` label
+# This workflow helps to run the CI pipeline for PRs made by external
+# contributors (from forks).
+
+on:
+  pull_request_target:
+    branches:
+      - main
+    types:
+      # Default types that trigger a workflow:
+      # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
+      - opened
+      - synchronize
+      - reopened
+      # Handled in addition to keep labels tidy:
+      - closed
+      # Actual magic happens here:
+      - labeled
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: false
+
+env:
+  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  PR_NUMBER: ${{ github.event.pull_request.number }}
+  BRANCH: "ci-run/pr-${{ github.event.pull_request.number }}"
+
+# No permissions for GITHUB_TOKEN by default; the **minimal required** set is granted per job.
+permissions: {}
+
+defaults:
+  run:
+    shell: bash -euo pipefail {0}
+
+jobs:
+  remove-label:
+    # Remove `approved-for-ci-run` whenever the PR code changes (or it is closed).
+    # Re-approval requires a human to review and re-apply the label.
+
+    permissions:
+      pull-requests: write # for `gh pr edit`
+
+    if: |
+      contains(fromJSON('["opened", "synchronize", "reopened", "closed"]'), github.event.action) &&
+      contains(github.event.pull_request.labels.*.name, 'approved-for-ci-run')
+
+    runs-on:
+      group: databricks-protected-runner-group
+      labels: linux-ubuntu-latest
+
+    steps:
+      - run: gh pr --repo "${GITHUB_REPOSITORY}" edit "${PR_NUMBER}" --remove-label "approved-for-ci-run"
+
+  create-or-update-pr-for-ci-run:
+    # Create local PR for an `approved-for-ci-run` labelled PR to run CI pipeline in it.
+
+    permissions:
+      pull-requests: write # for `gh pr edit` 
+      # For `git push` and `gh pr create` we use CI_ACCESS_TOKEN
+
+
+    if: |
+      github.event.action == 'labeled' &&
+      contains(github.event.pull_request.labels.*.name, 'approved-for-ci-run')
+
+    runs-on:
+      group: databricks-protected-runner-group
+      labels: linux-ubuntu-latest
+
+    steps:
+      - name: Generate a GitHub App installation token
+        id: app-token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          app-id: ${{ secrets.CI_APP_ID }}
+          private-key: ${{ secrets.CI_APP_PRIVATE_KEY }}
+
+      - name: Authorize — only users with write access may trigger fork CI
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          ACTOR: ${{ github.event.sender.login }}
+        run: |
+          # The `.permission` field collapses roles to admin/write/read/none
+          # (maintain reports as "write"), which matches "can approve CI runs".
+          PERM="$(gh api "repos/${GITHUB_REPOSITORY}/collaborators/${ACTOR}/permission" --jq '.permission')"
+          echo "Actor '${ACTOR}' has permission: ${PERM}"
+          if [ "${PERM}" != "admin" ] && [ "${PERM}" != "write" ]; then
+            echo "::error::Only users with write access may approve CI runs for fork PRs (actor '${ACTOR}' has '${PERM}')."
+            gh pr --repo "${GITHUB_REPOSITORY}" edit "${PR_NUMBER}" --remove-label "approved-for-ci-run" || true
+            exit 1
+          fi
+
+      - run: gh pr --repo "${GITHUB_REPOSITORY}" edit "${PR_NUMBER}" --remove-label "approved-for-ci-run"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/approved-for-ci-run.yml
```diff
@@ -0,0 +1,154 @@
...
+          echo "Actor '${ACTOR}' has permission: ${PERM}"
+          if [ "${PERM}" != "admin" ] && [ "${PERM}" != "write" ]; then
+            echo "::error::Only users with write access may approve CI runs for fork PRs (actor '${ACTOR}' has '${PERM}')."
+            gh pr --repo "${GITHUB_REPOSITORY}" edit "${PR_NUMBER}" --remove-label "approved-for-ci-run" || true
+            exit 1
+          fi
+
+      - run: gh pr --repo "${GITHUB_REPOSITORY}" edit "${PR_NUMBER}" --remove-label "approved-for-ci-run"
+
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
teodordelibasic-db: `actions/checkout` defaults to `fetch-depth: 1`. When a fork is more than one unique commit ahead of `main`, the base repository does not have the head commit's fork-only parents and the shallow checkout cannot supply them. The force-push is rejected with `shallow update not allowed`.

Could we use `fetch-depth: 0`, or otherwise fetch the full ancestry of the immutable approved SHA, before pushing the mirror branch?

.github/workflows/approved-for-ci-run.yml
```diff
@@ -0,0 +1,154 @@
...
+
+  create-or-update-pr-for-ci-run:
+    # Create local PR for an `approved-for-ci-run` labelled PR to run CI pipeline in it.
+
+    permissions:
+      pull-requests: write # for `gh pr edit` 
+      # For `git push` and `gh pr create` we use CI_ACCESS_TOKEN
+
+
+    if: |
```
teodordelibasic-db: The job currently runs for any label event while `approved-for-ci-run` remains present, and it also runs for same-repository PRs that do not need mirroring. If the collaborator lookup fails, `bash -e` exits before either label-removal step, so a later unrelated label can replay the workflow.

Could we:

- require `github.event.label.name == 'approved-for-ci-run'`;
- require `github.event.pull_request.head.repo.full_name != github.repository`;
- treat a failed collaborator lookup as denial; and
- consume `approved-for-ci-run` in a separate `if: always()` step?

That keeps the label tied to one explicit approval attempt and ensures every new head SHA needs a new approval.

.github/workflows/approved-for-ci-run.yml
```diff
@@ -0,0 +1,154 @@
...
+      github.event.pull_request.head.repo.full_name != github.repository
+
+    runs-on:
+      group: databricks-protected-runner-group
+      labels: linux-ubuntu-latest
+
+    steps:
+      - name: Close CI-run PR and delete the `ci-run/pr-${{ env.PR_NUMBER }}` branch
+        run: |
+          CLOSED="$(gh pr --repo "${GITHUB_REPOSITORY}" list --head "${BRANCH}" --json 'closed' --jq '.[].closed')"
```
teodordelibasic-db: `gh pr list` defaults to open PRs, and the branch is deleted only through `gh pr close --delete-branch`. A branch remains if the push succeeds but PR creation fails, or if the mirror PR was already closed without deleting its branch.

Could cleanup close any open mirror PR and then independently delete `refs/heads/ci-run/pr-N` through the GitHub API, tolerating a missing ref? The existing job-scoped `GITHUB_TOKEN` already has `contents: write`; generating another App token is not necessary.

.github/workflows/approved-for-ci-run.yml
teodordelibasic-db, line 96: This depth-one checkout breaks when a fork is more than one unique commit ahead of `main`. The base repository does not have the fork-only parent commits, and the shallow checkout cannot supply them when pushing `ci-run/pr-N`, so the remote rejects the push.

Could we fetch the approved SHA's full ancestry before pushing?

```yaml
with:
  fetch-depth: 0
```

.github/workflows/approved-for-ci-run.yml
teodordelibasic-db, line 63: This condition accepts any label event while `approved-for-ci-run` remains present, and it also accepts same-repository PRs that do not need mirroring. If the collaborator lookup fails, `bash -e` exits before the label is removed, so a later unrelated label can replay the job.

Could we restrict the job to the exact approval label on a fork PR?

```yaml
if: |
  github.event.action == 'labeled' &&
  github.event.label.name == 'approved-for-ci-run' &&
  github.event.pull_request.head.repo.full_name != github.repository
```

It would also be good to treat a failed permission lookup as denial and consume `approved-for-ci-run` in a separate step with `if: always()`, so every approval attempt removes the label.

.github/workflows/approved-for-ci-run.yml
teodordelibasic-db, line 151: Cleanup only deletes the branch while closing an open mirror PR. If the push succeeds but PR creation fails, or the mirror PR was already closed without deleting its branch, `ci-run/pr-N` remains in the repository.

Could we close any open mirror and then delete the ref independently?

```bash
gh pr list --state open --head "$BRANCH" --json number --jq '.[].number' |
  while read -r number; do
    gh pr close "$number"
  done
gh api --method DELETE "repos/$GITHUB_REPOSITORY/git/refs/heads/$BRANCH" || true
```

The existing job-scoped `GITHUB_TOKEN` has `contents: write`, so this does not need another App token.

.github/workflows/approved-for-ci-run.yml
```diff
@@ -0,0 +1,154 @@
...
+    if: |
+      github.event.action == 'labeled' &&
+      contains(github.event.pull_request.labels.*.name, 'approved-for-ci-run')
+
+    runs-on:
+      group: databricks-protected-runner-group
+      labels: linux-ubuntu-latest
+
+    steps:
+      - name: Generate a GitHub App installation token
```
teodordelibasic-db, line 72: The App is already scoped correctly, but neither this token request nor the PR description records the complete boundary. Could we request only `Contents: write` and `Pull requests: write`, switch from the deprecated `app-id` input to `client-id`, and document that the App is installed only on `databricks/zerobus-sdk` with no `Workflows: write` permission?

```yaml
with:
  client-id: ${{ secrets.CI_APP_CLIENT_ID }}
  private-key: ${{ secrets.CI_APP_PRIVATE_KEY }}
  permission-contents: write
  permission-pull-requests: write
```

The missing workflow permission means external PRs that change `.github/workflows/**` cannot use this mirror path. It would be good to document the alternate process for testing those contributions as well.

.github/workflows/approved-for-ci-run.yml
```diff
@@ -0,0 +1,154 @@
...
+      labels: linux-ubuntu-latest
+
+    steps:
+      - run: gh pr --repo "${GITHUB_REPOSITORY}" edit "${PR_NUMBER}" --remove-label "approved-for-ci-run"
+
+  create-or-update-pr-for-ci-run:
+    # Create local PR for an `approved-for-ci-run` labelled PR to run CI pipeline in it.
+
+    permissions:
+      pull-requests: write # for `gh pr edit`
```
teodordelibasic-db, line 59: Could we remove the trailing whitespace here and update the following comment to say GitHub App token instead of `CI_ACCESS_TOKEN`?

.github/workflows/approved-for-ci-run.yml
```diff
@@ -0,0 +1,154 @@
...
+      - run: git checkout -b "${BRANCH}"
+
+      - run: git push --force origin "${BRANCH}"
+
+      - name: Create a Pull Request for the CI run (if required)
+        if: steps.get-pr.outputs.ALREADY_CREATED == ''
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+        run: |
+          cat << EOF > body.md
```
teodordelibasic-db, line 118: `body.md` is created inside the fork worktree, so a committed symlink at that path can redirect this fixed-content write and fail the run. Could we pass the body directly to `gh pr create` and include the immutable approved SHA for auditability?

```yaml
env:
  APPROVED_SHA: ${{ github.event.pull_request.head.sha }}
run: |
  gh pr create \
    --body "CI run for PR #$PR_NUMBER. Approved SHA: $APPROVED_SHA" \
    --head "$BRANCH" \
    --base main \
    --draft
```

PR #460, opened 2026-06-30
[Typescript] Fix SDK identifier in user-agent header and expose applicationName

irinatomic-db:

- `ZerobusSdk` constructor now correctly sets `zerobus-sdk-ts/<version>` as the SDK identifier in the HTTP user-agent header, instead of falling back to `zerobus-sdk-rs`.
- Added optional `applicationName` constructor parameter, appended to the user-agent (e.g. `zerobus-sdk-ts/1.2.0 my-app/1.0`).
- Removed dead `user-agent` insertions from the headers providers, which were never effective at the transport level.

## Why

Without explicitly setting the SDK identifier, the server cannot distinguish TypeScript traffic from direct Rust usage, making per-language telemetry impossible. The applicationName feature already exists in Rust and Python - this brings TypeScript to parity.

## How is this tested?

- Built locally (`npm run build`) - compiles and NAPI-RS generates updated TypeScript types correctly.
- `npm run test:unit`
- Integration tests run against staging (shinkansen_eastus1.default.air_quality) — all passed
- Confirmed the user-agent header value on the wire: added a temporary `eprintln!` in the constructor and observed `zerobus-sdk-ts/1.1.0 irina-test/1.0` in the test output

### DIFF
--- a/typescript/src/lib.rs
+++ b/typescript/src/lib.rs
@@ -836,6 +836,13 @@ impl RustHeadersProvider for TsOAuthHeadersProvider {
     }
 }
 
+#[napi(object)]
+#[derive(Default)]
+pub struct ZerobusSdkOptions {
+    /// Identifier appended to the `user-agent` header
+    pub application_name: Option<String>,
+}
+
 /// The main SDK for interacting with the Databricks Zerobus service.
 ///
 /// This is the entry point for creating ingestion streams to Delta tables.
@@ -846,7 +853,7 @@ impl RustHeadersProvider for TsOAuthHeadersProvider {
 /// const sdk = new ZerobusSdk(
 ///   "https://workspace-id.zerobus.region.cloud.databricks.com",
 ///   "https://workspace.cloud.databricks.com",
-///   "my-app/1.0"
+///   { applicationName: "my-app/1.0" }
 /// );
 ///
 /// const stream = await sdk.createStream(
@@ -874,10 +881,8 @@ impl ZerobusSdk {
     ///   (e.g., "https://workspace-id.zerobus.region.cloud.databricks.com")
     /// * `unity_catalog_url` - The Unity Catalog endpoint URL
     ///   (e.g., "https://workspace.cloud.databricks.com")
-    /// * `application_name` - Optional application identifier appended to the
-    ///   HTTP `user-agent` header, conventionally `"<product>/<version>"`.
-    ///   The SDK prefix `zerobus-sdk-ts/<version>` is always present; this
-    ///   value is appended after it (e.g. `"zerobus-sdk-ts/1.2.0 my-app/1.0"`).
+    /// * `options` - Optional SDK configuration (see `ZerobusSdkOptions`),
+    ///   including `applicationName` for server-side attribution.
     ///
     /// # Errors
     ///
@@ -887,7 +892,7 @@ impl ZerobusSdk {
     pub fn new(
         zerobus_endpoint: String,
         unity_catalog_url: String,
-        application_name: Option<String>,
+        options: Option<ZerobusSdkOptions>,
     ) -> Result<Self> {
         let workspace_id = zerobus_endpoint
             .strip_prefix("https://")
@@ -900,11 +905,13 @@ impl ZerobusSdk {
                 )
             })?;
 
+        let options = options.unwrap_or_default();
+
         let builder = RustZerobusSdk::builder()
             .endpoint(&zerobus_endpoint)
             .unity_catalog_url(&unity_catalog_url)
             .sdk_identifier(TS_SDK_USER_AGENT);
-        let builder = match application_name {
+        let builder = match options.application_name {
             Some(name) => builder.application_name(name),
             None => builder,
         };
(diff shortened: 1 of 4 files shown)


typescript/src/lib.rs
```diff
@@ -879,13 +874,21 @@ impl ZerobusSdk {
...
     ///
     /// - Invalid endpoint URLs
     /// - Failed to extract workspace ID from the endpoint
     #[napi(constructor)]
-    pub fn new(zerobus_endpoint: String, unity_catalog_url: String) -> Result<Self> {
+    pub fn new(
+        zerobus_endpoint: String,
+        unity_catalog_url: String,
+        application_name: Option<String>,
+    ) -> Result<Self> {
```
elenagaljak-db, line 896: Is this a breaking change? We can't have those in minor releases.

irinatomic-db: It isn't. 

In NAPI-RS, `Option<String>` on a constructor parameter generates `applicationName?: string | null` in TypeScript - an optional parameter. All existing callers passing two arguments continue to work without any changes.

teodordelibasic-db (commented):
Similar to the Java PR, the internals look good, but I think we should change how this is exposed in order to have an API that can more easily evolve and expand.

The idiomatic way in Rust to do this is the builder pattern, however it's different in other languages.

In TypeScript, adding `applicationName` as a positional optional works for this one parameter, but I'd suggest we switch to an options object right away to avoid shooting ourselves in the foot in the future if we want to add more arguments.

The issue is that JS binds arguments by position and has no named arguments. Say we keep going down the positional path and end up with three optionals after the required args:

```typescript
new ZerobusSdk(endpoint, ucUrl, x?, y?, z?)
```

What positional lets you do: omit optionals from the right.

```typescript
new ZerobusSdk(endpoint, ucUrl, xVal, yVal)   // z omitted, fine
```

What it does not let you do: skip an earlier optional while setting a later one. To set `y` but not `x`, you can't leave `x` out or `y` lands in `x`'s slot - you have to pad with `undefined`:

```typescript
new ZerobusSdk(endpoint, ucUrl, undefined, yVal)              // set only y
new ZerobusSdk(endpoint, ucUrl, undefined, undefined, zVal)   // set only z
```

With real knobs that becomes unreadable and error-prone:

```typescript
new ZerobusSdk(endpoint, ucUrl, undefined, 5000, undefined, true)
```

You can't tell what `5000` or `true` configure without counting positions, and if two options share a type the compiler won't catch a value put in the wrong slot.

An options object fixes both problems, because fields are addressed by name so order is irrelevant and you specify only what you want:

```typescript
new ZerobusSdk(endpoint, ucUrl, { y: yVal });
new ZerobusSdk(endpoint, ucUrl, { z: zVal, x: xVal });
new ZerobusSdk(endpoint, ucUrl, { applicationName: "my-app/1.0" });
```

We already use this pattern one level down: `createStream(tableProps, clientId, clientSecret, options?)` takes `StreamConfigurationOptions` with all the optional knobs in one bag.

Concretely, something like:

```rust
#[napi(object)]
pub struct ZerobusSdkOptions {
    /// Application identifier appended to the user-agent header, e.g. "my-app/1.0".
    pub application_name: Option<String>,
    // future knobs slot in here with no signature change
}

#[napi(constructor)]
pub fn new(
    zerobus_endpoint: String,
    unity_catalog_url: String,
    options: Option<ZerobusSdkOptions>,
) -> Result<Self>
```

I am in no way a TS expert and it's not easy to keep track of API design in each language, so let me know if this makes sense. 🙂

irinatomic-db:
@teodordelibasic-db  

Good point, agreed. Switching to a `ZerobusSdkOptions` object so it's expandable without signature churn - consistent with the existing createStream options pattern. Will update the PR.

teodordelibasic-db (commented):
Looks good, thanks for changing the approach! Could we just look into adding a basic sanity check test, something along the lines of the following (or feel free to do something else):

```typescript
it('should accept an options object with applicationName', () => {
    const sdk = new ZerobusSdk(
        'https://1234567890.zerobus.us-west-2.cloud.databricks.com',
        'https://test-workspace.cloud.databricks.com',
        { applicationName: 'my-app/1.0' }
    );
    assert.ok(sdk);
});
```

irinatomic-db:
@teodordelibasic-db 

Added the happy-path tests (options object with applicationName, and empty options object).

Skipped an invalid-applicationName rejection test: this crate pins `databricks-zerobus-ingest-sdk = 2.0.1`, which does no `user-agent` validation, it accepts anything. That validation only landed in `2.0.3`. 

Should I bump the dependency version?

teodordelibasic-db:
> @teodordelibasic-db
> 
> Added the happy-path tests (options object with applicationName, and empty options object).
> 
> Skipped an invalid-applicationName rejection test: this crate pins `databricks-zerobus-ingest-sdk = 2.0.1`, which does no `user-agent` validation, it accepts anything. That validation only landed in `2.0.3`.
> 
> Should I bump the dependency version?

We can do it, but no need for this PR.

teodordelibasic-db (approved):
LGTM

PR #461, opened 2026-06-30
feat(zeroparser): add MessageRegistry::from_descriptors for sibling/imported types

pixie79:

Add a new constructor, `MessageRegistry::from_descriptors(root, others)`, to the
`zeroparser` registry so a parse registry can be built from the **whole**
`FileDescriptorSet` — the root message *plus* the sibling/imported top-level
messages it references — rather than from the root descriptor alone.

Concretely:

- **New API** — `from_descriptors(root: &DescriptorProto, others: &[(String, &DescriptorProto)])`.
  It registers the root exactly as `from_descriptor` does (so root-relative
  lookups are identical), then registers every other top-level message under its
  **true package-qualified FQN** (leading dot), e.g. package `google.protobuf` +
  message `StringValue` → key `.google.protobuf.StringValue`. That key matches
  the `type_name` form carried on message-typed `FieldDescriptorProto`s, so
  field lookups now resolve.
- **Shared collector** — `collect_messages` now delegates to a small
  `collect_messages_with_prefix(root, initial_prefix, acc)` helper that threads
  the enclosing namespace as a prefix. An empty prefix yields `.Name` (the
  existing root behaviour); a package prefix like `.google.protobuf` yields
  `.google.protobuf.Name`.
- **No behavioural change to existing API** — `from_descriptor` is unchanged
  (it simply calls the helper with the root prefix), so all current callers and
  semantics are preserved byte-for-byte. The change is purely additive.

## Why are these changes needed?

`MessageRegistry::from_descriptor` only recurses `nested_type`. It therefore
registers the root message and its *nested* children, but **never** the sibling
or imported top-level messages that live in other files of the same
`FileDescriptorSet`.

The most common casualty is the Protobuf well-known types. A root message with a
field typed `.google.protobuf.StringValue` (imported from
`google/protobuf/wrappers.proto`) compiles and loads fine, but at parse time the
lookup for `.google.protobuf.StringValue` misses the registry and decoding fails
on **every record that populates such a field**:

```
UnknownTypeName: Unknown message type '.google.protobuf.StringValue' not in registry
```

This affects any schema that uses imported messages — the `google.protobuf`
wrappers (`StringValue`, `Int64Value`, …), `Timestamp`, `Duration`, `Any`, or a
user's own imported `.proto` types. There is currently no public way to populate
the registry with those sibling/imported descriptors, because the `messages` map
and the collector are private. `from_descriptors` provides that path without
disturbing the existing root-only constructor, so callers that have the full
descriptor set can opt in and decode imported-type fields correctly.

## How is this tested?

`cargo test -p databricks-zerobus-ingest-sdk --features zeroparser` (the
`zeroparser` module is feature-gated):

- **`zeroparser::registry`** — 6 passed, 0 failed, including the two new tests:
  - `message_registry_from_descriptors_registers_siblings` — proves
    `from_descriptor` alone does **not** register an imported
    `.google.protobuf.StringValue` (lookup is `None`), while `from_descriptors`
    registers it under its true FQN, keeps the root as the parse root, and the
    field still resolves to `.google.protobuf.StringValue`.
  - `message_registry_from_descriptors_empty_package_and_nesting` — covers the
    empty (unnamed) package case (sibling registers at `.Name`) and confirms a
    sibling's own nested types are collected (`.Sibling.Inner`).
- **Full `zeroparser` suite** — 82 passed, 0 failed; the e2e example binary
  compiles green.

The change is also exercised downstream (a consumer that builds the registry
from the full `FileDescriptorSet` via `from_descriptors`): a record carrying a
populated `StringValue` and `Timestamp` on the wire previously failed with the
`UnknownTypeName` error above and now decodes and round-trips correctly.

---

Signed-off-by: Mark Olliver <mark.olliver@flutter.com>

### DIFF
--- a/rust/sdk/src/zeroparser/registry.rs
+++ b/rust/sdk/src/zeroparser/registry.rs
@@ -185,29 +185,12 @@ impl MessageRegistry {
         }
     }
 
-    /// Build a registry from a root descriptor plus the other top-level message
-    /// descriptors in the same `FileDescriptorSet`, each paired with its proto
-    /// package so it is registered under its TRUE fully-qualified name.
+    /// Build a registry from the parse root plus additional top-level messages.
     ///
-    /// `from_descriptor` only recurses `nested_type`, so it registers the root
-    /// and its nested children but NOT sibling or imported top-level messages
-    /// from other files in the set (e.g. `google.protobuf.StringValue` from
-    /// `google/protobuf/wrappers.proto`). A `root` field typed
-    /// `.google.protobuf.StringValue` then misses on lookup and decode fails
-    /// with `UnknownTypeName`. This constructor closes that gap.
-    ///
-    /// `others` is a slice of `(package, descriptor)` pairs — one per top-level
-    /// message reachable from the set. `package` is the proto `package` of the
-    /// file the message came from (may be empty for the unnamed package). Each
-    /// message and its nested types are inserted under their package-prefixed
-    /// FQN with a leading dot, e.g. package `google.protobuf` + message
-    /// `StringValue` → key `.google.protobuf.StringValue`, matching the
-    /// `type_name` form on message-typed `FieldDescriptorProto`s.
-    ///
-    /// `root` remains the parse root (`root_descriptor`); the existing
-    /// `from_descriptor` API is unchanged.
+    /// Use this when root fields reference sibling/imported messages that are
+    /// not nested under `root`.
     #[inline(always)]
-    pub fn from_descriptors(root: &DescriptorProto, others: &[(String, &DescriptorProto)]) -> Self {
+    pub fn from_descriptors(root: &DescriptorProto, others: &[(&str, &DescriptorProto)]) -> Self {
         let mut messages = HashMap::new();
 
         // Register the root (and its nested types) at the root prefix, exactly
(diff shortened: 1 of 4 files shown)


rust/sdk/src/zeroparser/registry.rs
```diff
@@ -185,15 +185,75 @@ impl MessageRegistry {
...
+    /// file the message came from (may be empty for the unnamed package). Each
+    /// message and its nested types are inserted under their package-prefixed
+    /// FQN with a leading dot, e.g. package `google.protobuf` + message
+    /// `StringValue` → key `.google.protobuf.StringValue`, matching the
+    /// `type_name` form on message-typed `FieldDescriptorProto`s.
+    ///
+    /// `root` remains the parse root (`root_descriptor`); the existing
+    /// `from_descriptor` API is unchanged.
+    #[inline(always)]
+    pub fn from_descriptors(root: &DescriptorProto, others: &[(String, &DescriptorProto)]) -> Self {
```
brankogrb-db: Change seems reasonable but let's add e2e tests for it

rust/sdk/src/zeroparser/registry.rs
```diff
@@ -185,15 +185,75 @@ impl MessageRegistry {
         }
     }
 
+    /// Build a registry from a root descriptor plus the other top-level message
```
brankogrb-db: let's try to keep comments short and concise.

rust/sdk/src/zeroparser/registry.rs
brankogrb-db: use `&str` instead of `String`

brankogrb-db:
Let's also run existing benchmarks to make sure we didn't hit some perf regression - even though we shouldn't

brankogrb-db:
Since I see you generated this PR with AI, I suggest running self-review on this PR with AI also. I think it can be super helpful on finding out some of the issues

PR #462, opened 2026-07-01
[Java] Expose applicationName parameter in ZerobusSdk constructor

irinatomic-db:

- Added a three-argument `ZerobusSdk(String serverEndpoint, String unityCatalogEndpoint, String applicationName)` constructor. The existing two-argument constructor is unchanged and delegates to the new one with null.
- The `applicationName` value is passed through JNI to the Rust builder's `.application_name()`, appending it to the HTTP `user-agent` header (e.g. `zerobus-sdk-java/1.3.0 my-app/1.0`).
- Updated `nativeCreate` JNI signature in both Java and Rust to carry the new nullable parameter.

## Why

The `applicationName` feature already exists in the Rust and Python SDKs (and was just added to TypeScript in [#460](https://github.com/databricks/zerobus-sdk/pull/460)). This brings Java to parity.

## How is this tested?

- Rust JNI crate builds cleanly (cargo build in rust/jni/).
- Maven not available in the dev environment; Java compilation not verified locally.
- Not tested end-to-end against a live Zerobus service.

### DIFF
--- a/java/NEXT_CHANGELOG.md
+++ b/java/NEXT_CHANGELOG.md
@@ -9,9 +9,11 @@
 - Added native library support for Linux musl (Alpine) on x86_64 and aarch64. The libc flavor is detected automatically at runtime; override with `-Dzerobus.libc=musl|glibc`.
 - Added `ZerobusSdk.streamBuilder()`, a fluent builder for creating streams that mirrors the Rust SDK's `stream_builder()`. It supports JSON, Protocol Buffer, and Arrow Flight streams through a single chainable API and is now the recommended way to create streams. See `StreamBuilder`.
 - `ZerobusSdk` now has a three-argument constructor accepting an optional `applicationName`
-  parameter. When set, it is appended to the HTTP `user-agent` header sent on every request
-  (e.g. `zerobus-sdk-java/1.3.0 my-app/1.0`), enabling server-side attribution per
-  customer application. The existing two-argument constructor is unchanged.
+  parameter. When set, it is appended to the HTTP `user-agent` header on gRPC requests to the
+  Zerobus service (it is not sent on requests to the login service that mint the OAuth token),
+  so callers can be identified in server-side telemetry. The wire value becomes
+  `zerobus-sdk-java/<version> <applicationName>` (e.g. `zerobus-sdk-java/1.3.0 my-app/1.0`).
+  The existing two-argument constructor is unchanged.
 
 ### Bug Fixes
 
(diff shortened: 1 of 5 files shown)


teodordelibasic-db (commented):
Let's just wait for @danilonajkov-db to merge his Java stream builder [PR](https://github.com/databricks/zerobus-sdk/pull/425) and then expose this via the builder. We don't want to add a new constructor for every combination of configuration knobs such as this one.

teodordelibasic-db:
Danilo's PR is merged. 🥳

irinatomic-db:
Synced offline - application_name is a value on SDK level, not on stream level

java/NEXT_CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
 
 - Added native library support for Linux musl (Alpine) on x86_64 and aarch64. The libc flavor is detected automatically at runtime; override with `-Dzerobus.libc=musl|glibc`.
 - Added `ZerobusSdk.streamBuilder()`, a fluent builder for creating streams that mirrors the Rust SDK's `stream_builder()`. It supports JSON, Protocol Buffer, and Arrow Flight streams through a single chainable API and is now the recommended way to create streams. See `StreamBuilder`.
+- `ZerobusSdk` now has a three-argument constructor accepting an optional `applicationName`
+  parameter. When set, it is appended to the HTTP `user-agent` header sent on every request
```
teodordelibasic-db: For Python we slightly corrected this to say it isn't on every request since we don't use it for fetching the token, so let's just correct it here as well to something like `sent on gRPC requests to the Zerobus service` and maybe explicitly mention it's not used for requests towards the login service.

java/NEXT_CHANGELOG.md
teodordelibasic-db, line 39: Let's add the constructor to this section as well.

java/src/main/java/com/databricks/zerobus/ZerobusSdk.java
```diff
@@ -98,9 +98,24 @@ public class ZerobusSdk implements AutoCloseable {
...
+   *
+   * @param serverEndpoint The gRPC endpoint URL for the Zerobus service.
+   * @param unityCatalogEndpoint The Unity Catalog endpoint URL.
+   * @param applicationName Optional application identifier appended to the HTTP {@code user-agent}
+   *     header, conventionally {@code "<product>/<version>"} (e.g. {@code "my-app/1.0"}). When set,
+   *     the header becomes {@code "zerobus-sdk-java/<version> <applicationName>"}. Pass {@code
+   *     null} to omit.
+   * @throws ZerobusException if the SDK cannot be initialized
+   */
+  public ZerobusSdk(String serverEndpoint, String unityCatalogEndpoint, String applicationName) {
```
teodordelibasic-db, line 115: Could you look into adding a construction test that will run in CI: `ci-java.yml` builds and stages the JNI lib (`--features test-helpers`) before `mvn test`, and `build()` only parses the endpoint (no server/creds needed), so dummy endpoints suffice.

We can assert with dummy endpoints that `null` and `"my-app/1.0"` construct successfully. We can also guard on native-lib availability like `StreamBuilderTest.assumeNativeLibrary()` so bare local runs skip rather than fail.

teodordelibasic-db (commented):
Code changes look good, just a few test/doc nits that would probably fall through the cracks if we don't address it here. 🙂

teodordelibasic-db (approved):

PR #463, opened 2026-07-01
[PureGo] Add gRPC transport and stream layer

zlata-stefanovic-db:

Adds the lowest layer of the pure-Go SDK — the `internal/transport` package — structured so the proto/JSON path and the coming Arrow/Flight path share as much code as possible.

**Connection (`transport.go`)** — `Conn` wraps the gRPC connection. `Dial` secures with TLS (system root CAs, TLS 1.2 floor) by default, overridable via `WithTLSConfig` / `WithInsecure` / `WithGRPCDialOptions`. A `Conn` is safe for concurrent use and backs many streams.

**Protocol-agnostic core (`rawstream.go`)** — the shared half of an open stream:
- `bidiRPC[Req, Resp]` — the minimal `Send`/`Recv`/`CloseSend` interface a bidirectional gRPC stream client exposes. The `EphemeralStream` client satisfies it today; the Arrow Flight `DoPut` client will too.
- `rawStream[Req, Resp]` — send/receive plumbing, error wrapping, idempotent teardown, and a shared **handshake** (send setup message → await readiness → validate → record stream id). Only two per-protocol hooks vary.

**Shared metadata (`metadata.go`)** — `streamContext` applies the table-name and authorization headers and wraps a cancelable child context. Both headers are protocol-agnostic (Arrow uses the same table-name header), so this setup is shared, not duplicated per protocol.

**Proto/JSON stream (`stream.go`)** — `Stream` embeds `rawStream[EphemeralStreamRequest, EphemeralStreamResponse]` and supplies only what's proto-specific: opening the `EphemeralStream` RPC, framing the create-stream setup message, and recognizing the readiness response (the stream id).

**Why:** this is the foundation the ingest and recovery layers build on, and it deliberately shares everything but the irreducible per-protocol pieces. When the Arrow path lands, it embeds the same `rawStream`, reuses `streamContext` and the handshake flow, and adds only its three protocol-specific pieces (which RPC to open, how to frame the setup message, how to recognize "ready"). The layer deals only in wire types from `internal/zerobuspb` and knows nothing about batching, offsets, or acknowledgements — those come in follow-up layers.

## Notes for reviewers

- **No changelog entry.** All code is under `purego/internal/` — there is no public API yet (Go blocks external imports of `internal/`), so nothing user-facing to log. `purego/` has no `NEXT_CHANGELOG.md`/`CHANGELOG.md`; those will be introduced with the public `zerobus` package (see the planned build order), not in this transport PR.
- **Auth header normalization.** `authHeaderValue` (`metadata.go`) treats a token containing a space as already carrying an auth scheme and sends it verbatim; a bare token is prefixed with `"Bearer "`. Real credentials (OAuth/JWT/PAT) never contain spaces, and the contract is documented at `metadata.go:35-38`.
- Merged latest `main`, which includes the canonical-proto unification (#464): the per-SDK `zerobus_service.proto` copies were removed and `internal/zerobuspb/gen.go` now regenerates from the shared `rust/sdk/zerobus_service.proto`.

## How is this tested?

Unit tests in `transport_test.go` drive the package through an in-memory gRPC listener: dial/TLS option handling, the create-stream handshake (including auth-scheme passthrough, table-name trimming, descriptor and record-type validation, and rejecting unexpected handshake responses), `Send`/`Recv`, and `Close` idempotency/teardown. All pass under `go test`, with `go vet` and `gofmt` clean.


zlata-stefanovic-db:
Tracking issue for the overall pure-Go SDK design: #467. This PR lands the transport foundation (first checklist item).

purego/internal/transport/metadata.go
teodordelibasic-db, line 1: Could you look into whether it makes more sense to name this file `headers.go` or something along those lines. I know currently we are going with `metadata.go` due to that being idiomatic to Go so to say, but it kind of differs from regular Zerobus concepts. If you deem current name is more apt I am fine with it.

purego/internal/transport/metadata.go
```diff
@@ -0,0 +1,49 @@
...
+// token is prefixed with "Bearer "; a value that already carries a scheme (for
+// example "Bearer ..." or "Basic ...") is returned verbatim. An empty token
+// yields "".
+func authHeaderValue(token string) string {
+	token = strings.TrimSpace(token)
+	if token == "" {
+		return ""
+	}
+	// If the caller already supplied a scheme (for example "Bearer ..."), keep it.
+	if strings.Contains(token, " ") {
```
teodordelibasic-db: Why do we need this handling? `if strings.Contains(token, " ") { return token }` means a malformed token like `"my token"` is sent with no scheme prefix. If we want this, maybe we can make it more robust by matching a known scheme prefix (`Bearer`/`Basic`/`DPoP`) instead of any-space.

purego/internal/transport/rawstream.go
```diff
@@ -0,0 +1,111 @@
...
+// opens the stream — concrete streams embed it and supply those via wire types
+// and handshake hooks. Stream (proto/JSON over EphemeralStream) is the current
+// implementation.
+//
+// Like the underlying gRPC stream, a rawStream is not safe for concurrent send;
+// pair it with a single writer goroutine. Calling send and recv from separate
+// goroutines is fine.
+type rawStream[Req, Resp any] struct {
+	rpc    bidiRPC[Req, Resp]
+	name   string // identifies the stream in error messages
```
teodordelibasic-db: Kind of a weird looking comment, not sure if it matches the standard.

purego/internal/transport/ephemeral_stream.go
teodordelibasic-db, line 1: As far as I understand, this file is meant to hold the "basic gRPC" stream implementation. If so, let's denote it more clearly, either via file and struct name or by introducing a parent folder.

purego/internal/transport/stream.go
```diff
@@ -0,0 +1,150 @@
...
+	case zerobuspb.RecordType_PROTO, zerobuspb.RecordType_JSON:
+		// Supported.
+	default:
+		return nil, fmt.Errorf("transport: open %q: unsupported record type %v", p.TableName, p.RecordType)
+	}
+	if p.RecordType == zerobuspb.RecordType_PROTO && len(p.DescriptorProto) == 0 {
+		return nil, fmt.Errorf("transport: open %q: descriptor proto required for PROTO records", p.TableName)
+	}
+
+	ctx, cancel := streamContext(ctx, p.TableName, p.Token)
```
teodordelibasic-db: `streamContext` wraps the caller ctx in `context.WithCancel(ctx)`, and the same ctx
is what the live stream runs on and what the handshake `Recv()` blocks on. 

We can have two bad outcomes:

1.  A caller who bounds `Open` with `context.WithTimeout(ctx, 5s)` to avoid a hung handshake silently gets the whole stream torn down 5s later mid-ingest.
2. A caller who passes `context.Background()` gets an `Open` that hangs forever if
the server half-opens, since the handshake has no deadline of its own.

To fix we can use the caller ctx only for dial/handshake and run the live stream on an internal ctx whose only cancel is `Close()`; at minimum give the handshake `Recv()` an independent deadline so `Open` can't hang.

purego/internal/transport/metadata.go
```diff
@@ -0,0 +1,49 @@
...
+const (
+	mdTableName     = "x-databricks-zerobus-table-name"
+	mdAuthorization = "authorization"
+)
+
+// streamContext builds the outgoing context an ingestion stream needs: table
+// name and authorization headers, wrapped in a cancelable child whose cancel
+// func tears the stream down.
+func streamContext(ctx context.Context, tableName, token string) (context.Context, context.CancelFunc) {
+	ctx = metadata.AppendToOutgoingContext(ctx, mdTableName, tableName)
```
teodordelibasic-db: `metadata.AppendToOutgoingContext` is used for both `x-databricks-zerobus-table-name`
and `authorization`. If the caller's ctx already carries either key (an interceptor,
a reused ctx, a future higher layer), the stream sends two values and server-side
first-value-wins can route to the wrong table or send a stale token.

We can set the Zerobus-owned keys to exactly one value (read existing outgoing MD, replace these two
keys, keep unrelated user metadata). We can also add a test that opens with a ctx already carrying these keys and asserts a single intended value reaches the server.

purego/internal/transport/rawstream.go
```diff
@@ -0,0 +1,111 @@
...
+// Errors carry a concise cause (from a hook, or the await step); the caller is
+// expected to annotate them with its operation context (e.g. the table name).
+func (s *rawStream[Req, Resp]) handshake(
+	sendSetup func(rpc bidiRPC[Req, Resp]) error,
+	confirmReady func(resp *Resp) (id string, err error),
+) error {
+	if err := sendSetup(s.rpc); err != nil {
+		return err
+	}
+	resp, err := s.rpc.Recv()
```
teodordelibasic-db: We call `s.rpc.Recv()` directly here, so handshake errors miss the `"transport: recv on stream %s"` wrapping every other path uses, and open-time vs steady-state error strings diverge. Minor consistency

We can route it through `s.recv()`.

purego/internal/transport/ephemeral_stream.go
```diff
@@ -0,0 +1,150 @@
...
+// CloseSend signals that no more requests will be sent, half-closing the stream
+// while leaving Recv open to drain remaining responses. For a graceful shutdown,
+// call CloseSend, read until io.EOF, then Close.
+func (s *Stream) CloseSend() error { return s.closeSend() }
+
+// Close aborts the stream and releases its resources. It is idempotent and safe
+// to call after a graceful CloseSend/drain. Unlike CloseSend it does not wait
+// for the server: any in-flight Send or Recv is unblocked with a cancellation
+// error.
+func (s *Stream) Close() { s.close() }
```
teodordelibasic-db, line 178: `close()` only fires the context cancel, so `Close()` without a preceding drain makes
the server see RST_STREAM rather than END_STREAM. The correct order (CloseSend; drain
to `io.EOF`; Close) is documented on `Stream.CloseSend`, but nothing encapsulates it,
so every higher layer has to re-implement it and can get it wrong.

This is the mistake the Rust gRPC path had to unwind (`stream/grpc/receiver.rs:212` drains
inline on close for this reason). Not a bug today, but better to fix now.

Once ingest lands, we can add one `GracefulClose(ctx)` that does the sequence and
keep `Close()` as the hard-abort escape hatch. 

We can add a graceful-close test now (`CloseSend`; drain to EOF; `Close`); current tests only cover abortive close (`transport_test.go:317`).

purego/internal/transport/rawstream.go
teodordelibasic-db: Why did we decide to go with `name` and not `ID` or something similar here?

purego/internal/transport/rawstream.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return fmt.Errorf("await ready response: %w", err)
+	}
+	id, err := confirmReady(resp)
+	if err != nil {
+		return err
+	}
+	if id != "" {
```
teodordelibasic-db, line 148: This function mutates `s.name`, which `ID()`, `send`, `recv`, and `closeSend` read.

Safe today because the write only happens during single-threaded `Open`, before the `*Stream` escapes to the caller, so no concurrent reader exists and `go test -race` is clean.

But recovery mints a new server stream ID on each reconnect, so `name` genuinely has to be reassigned on a background supervisor goroutine while the writer goroutine is calling `send`/`recv`/`ID()`.

That is a real data race. A Go string is a two-word (pointer+len) value, so a concurrent read can tear, and `-race` (Go's data-race detector, run in CI) would hard-fail the recovery PR.

We should make the field concurrency-safe: store it in an `atomic.Pointer[string]` and have `send`/`recv`/`closeSend`/`ID()` read it through a `name()` accessor that falls back to the fixed `"ephemeral-stream"` label when nil; recovery does `id.Store(&newID)`

We should also probably have a stable client-side stream ID for log correlation across recoveries. This can be a follow up and we should decide what does `ID` actually return

purego/internal/transport/metadata.go
zlata-stefanovic-db: Switching from any-space to a known-scheme prefix check (case-insensitive Bearer/Basic/DPoP); anything else is treated as a bare token and gets Bearer . Updating the doc comment on authHeaderValue and StreamParams.Token to match.

purego/internal/transport/rawstream.go
zlata-stefanovic-db: Rewording this

purego/internal/transport/metadata.go
zlata-stefanovic-db: Fixed. Renamed streamContext -> withStreamMetadata, which reads the existing outgoing metadata, then Sets (replaces) the two Zerobus-owned keys to exactly one value each - and Deletes authorization when no token is supplied so a stale inherited one can't leak through. Unrelated caller metadata is copied through untouched

purego/internal/transport/rawstream.go
zlata-stefanovic-db: Fixed this

purego/internal/transport/stream.go
zlata-stefanovic-db: Fixed as suggested

purego/internal/transport/ephemeral_stream.go
zlata-stefanovic-db: Added a graceful-close test for now

purego/internal/transport/metadata.go
zlata-stefanovic-db: Kept it as metadata for now since the file manipulates gRPC metadata.MD directly and the name mirrors that package, might consider to rename it later, but this might make more sense for the context now

purego/internal/transport/rawstream.go
zlata-stefanovic-db: It actually holds the ID, renaming this

purego/internal/transport/rawstream.go
zlata-stefanovic-db: Fixed as suggested

purego/internal/transport/rawstream.go
```diff
@@ -0,0 +1,156 @@
...
+//
+// hctx (the caller's Open context) bounds only the readiness wait, defaulting to
+// defaultHandshakeTimeout when it has no deadline; the live stream runs on a
+// separate Close-only context, so this never tears down an established stream.
+//
+// The two hooks are protocol-specific: sendSetup writes the first message, and
+// confirmReady validates the first response and returns the stream ID ("" if the
+// protocol assigns none). Errors carry a concise cause for the caller to
+// annotate.
+func (s *rawStream[Req, Resp]) handshake(
```
davidtosovic-db, line 110: `handshake` starts a `Recv` in a goroutine, and if `hctx` is canceled it relies on the caller to cancel the underlying stream context so that goroutine can exit.

That is true for the current `Stream.Open` path, so I don’t think this is broken today, but it feels easy to miss when the Arrow/Flight path reuses this helper.

Could we either make that contract explicit in the `handshake` comment, or pass a cleanup/cancel function into `handshake` and call it before returning on timeout/cancel? I slightly prefer the latter since it keeps the teardown behavior next to the goroutine that needs it.

purego/internal/transport/transport.go
```diff
@@ -0,0 +1,87 @@
...
+
+// WithTLSConfig secures the connection with a custom TLS configuration,
+// replacing the default of system root CAs.
+func WithTLSConfig(tc *tls.Config) DialOption {
+	return func(cfg *dialConfig) { cfg.creds = credentials.NewTLS(tc) }
+}
+
+// WithInsecure disables transport security. It is intended for tests and local,
+// plaintext endpoints, and must not be used against production.
+func WithInsecure() DialOption {
```
davidtosovic-db: Nit: Since tokens are a part of gRPC metadata, an accidental `WithInsecure()` outside of tests would put bearer tokens on the wire in plaintext. This is ok for now, but might be good to gate this behind a test-only build tag or something similar.

purego/internal/transport/ephemeral_stream.go
```diff
@@ -0,0 +1,157 @@
...
+		return nil, fmt.Errorf("transport: open %q: unsupported record type %v", p.TableName, p.RecordType)
+	}
+	if p.RecordType == zerobuspb.RecordType_PROTO && len(p.DescriptorProto) == 0 {
+		return nil, fmt.Errorf("transport: open %q: descriptor proto required for PROTO records", p.TableName)
+	}
+
+	// The live stream must outlive Open: detach from ctx's cancel/deadline
+	// (WithoutCancel keeps its values, so caller metadata survives) and make it
+	// cancelable only via Close. ctx still bounds the handshake below.
+	streamCtx := withStreamMetadata(context.WithoutCancel(ctx), p.TableName, p.Token)
```
davidtosovic-db, line 78: There might be a small lifecycle issue here.

The stream context is detached with `context.WithoutCancel(ctx)`, which makes sense once `Open` has succeeded, because we don’t want a short setup timeout to kill a stream that runs for a long time. But today that detached context is also used while the stream is still being opened and while the first create-stream message is sent. So if the caller’s context is canceled during that part, `Open` may not stop promptly.

Could we keep the stream detached after a successful handshake, but have the caller context still cancel the in-progress open attempt? In practice that probably means wiring the open context to cancel the stream context until the handshake succeeds, then removing that wiring before returning the stream.

LLM generated this suggestion for the fix:
```go
func (c *Conn) Open(ctx context.Context, p StreamParams) (*Stream, error) {
	// validate params first...
	openCtx := ctx
	if _, ok := openCtx.Deadline(); !ok {
		var cancelOpen context.CancelFunc
		openCtx, cancelOpen = context.WithTimeout(openCtx, defaultHandshakeTimeout)
		defer cancelOpen()
	}
	// This context is for the eventual live stream, so it does not inherit
	// the caller's cancellation/deadline.
	streamCtx := withStreamMetadata(context.WithoutCancel(ctx), p.TableName, p.Token)
	streamCtx, cancelStream := context.WithCancel(streamCtx)
	// While Open is still in progress, caller cancellation should still tear
	// down the stream attempt so EphemeralStream/Send/Recv can unblock.
	stopOpenCancel := context.AfterFunc(openCtx, cancelStream)
	stream, err := c.open(openCtx, streamCtx, p)
	if err != nil {
		cancelStream()
		return nil, err
	}
	// From this point on, the returned stream should outlive openCtx.
	if !stopOpenCancel() {
		cancelStream()
		return nil, fmt.Errorf("transport: open %q: %w", p.TableName, openCtx.Err())
	}
	stream.cancel = cancelStream
	return stream, nil
}
```

purego/internal/transport/rawstream.go
zlata-stefanovic-db: Agreed, implemented the latter option.

purego/internal/transport/transport.go
zlata-stefanovic-db: Done, moved the WithInsecure() into export_test.go in package transport so it compiles only under go test & production code can't reach it.

purego/internal/transport/ephemeral_stream.go
zlata-stefanovic-db: Very good catch & a real issue. I went with your preferred option - handshake now takes a teardown context.CancelFunc and calls it and waits for the recv goroutine to exit before returning

teodordelibasic-db (commented):
LGTM, but could we add a `purego` job that runs `go test -race ./...` in the CI, wired in the same way as the other modules?

davidtosovic-db (approved):

PR #465, opened 2026-07-01
[Rust] Dynamic protobuf support

### LINKED ISSUE #372: [Rust] Dynamic protobuf: runtime descriptor fetch + dynamic message ingestion

### LINKED ISSUE #521: [Rust] Add option to fetch schema from UC when using dynamic proto
### SDK

Rust

### Description

https://github.com/databricks/zerobus-sdk/pull/465 introduced dynamic proto in rust. To make it more convenient, add option to automatically fetch the table schema from Unity Catalog and create `MessageDescriptor` from it.

### Proposed Solution

_No response_

### Additional Context

_No response_

andrijast-db:

Closes #372

Adds a native dynamic-protobuf path for ingesting records whose schema is known only at runtime, with no compiled prost::Message type.

New public API in the dynamic module:
- `StreamBuilder::dynamic_proto(descriptor)` - select the format from a runtime DescriptorProto.
- `ZerobusStream::message_descriptor()` / `new_record()` - resolve the schema (cached) and start a record.
- `dynamic::DynamicRecord` - build records field-by-field with a validating, chainable `.set(field, value)?`.
- `dynamic::message_descriptor()` standalone resolver, the `IntoDynamicValue` conversion trait, and re-exports of `prost_reflect::{DynamicMessage, MessageDescriptor, Value}`.

Follow-ups:
- Add ability to fetch schema from UC

## How is this tested?

Manually tested with custom Rust client


rust/sdk/src/dynamic.rs
```diff
@@ -0,0 +1,251 @@
...
+            .map_err(|e| {
+                ZerobusError::InvalidArgument(format!("cannot set field '{field}': {e}"))
+            })?;
+        Ok(self)
+    }
+}
+
+impl From<DynamicRecord> for EncodedRecord {
+    fn from(record: DynamicRecord) -> Self {
+        EncodedRecord::Proto(record.message.encode_to_vec())
```
zlata-stefanovic-db: Hit this exact thing in the C FFI dynamic-proto path ([#371)](https://github.com/databricks/zerobus-sdk/pull/371): prost-reflect doesn't enforce proto2 required presence, so a record missing a required field encodes fine and only fails server-side. Non-nullable UC columns become required (descriptor_from_uc_columns - Label::Required), and .set() only checks fields you actually set - so omitting one (typo, sparse data, a new NOT NULL column) slips through here. I wrote a collect_missing_required_fields helper there that walks the message before encode_to_vec() and rejects locally with the field path. [#431](https://github.com/databricks/zerobus-sdk/pull/431)  Maybe we lift it into shared core and call it here too?

andrijast-db: Yeah, let's do that. Thanks!

rust/examples/proto/dynamic/single.rs
teodordelibasic-db, line 1: Not 100% confident, but I think it's more clear to have another folder under `examples/proto/`, i.e. `dynamic/`, which then also branches into `single.rs` and `batch.rs` or something like that. We can also rename current `proto/` to `compiled/`.

rust/examples/proto/README.md
```diff
@@ -28,6 +31,7 @@ Protocol Buffers examples provide type safety and better performance. **No schem
 **Available examples:**
 - **`single.rs`** - Ingest records one at a time using `ingest_record_offset()` / `ingest_record()`
 - **`batch.rs`** - Ingest multiple records at once using `ingest_records_offset()` / `ingest_records()`
+- **`dynamic.rs`** - Build the schema at runtime (no compiled `.proto`) and fill records field-by-field with `DynamicRecord`
```
teodordelibasic-db: Adding to comment in `examples/proto/dynamic.rs`, it feels off to have Proto Dynamic as a dimension here. IMO Proto should branch into dynamic and compiled.

rust/sdk/src/dynamic_proto.rs
teodordelibasic-db, line 1: I think this file is ambiguously named, we should probably be more specific/descriptive.

rust/sdk/Cargo.toml
```diff
@@ -21,6 +21,7 @@ all-features = true
 async-trait.workspace = true
 prost.workspace = true
 prost-types.workspace = true
+prost-reflect = { workspace = true, features = ["serde"] }
```
teodordelibasic-db: The `serde` feature is unused I think.

rust/README.md
```diff
@@ -501,6 +532,7 @@ The SDK provides flexible ways to ingest data with different levels of abstracti
 | `ProtoMessage<T>` | Proto | Auto-encoding: pass structs, SDK handles encoding |
 | `ProtoBytes` | Proto | Pre-encoded: pass bytes with explicit wrapper |
 | `Vec<u8>` | Proto | Backward-compatible: raw bytes without wrapper |
+| `DynamicRecord` | Proto | Dynamic: build fields at runtime against a descriptor ([`dynamic_proto`](#dynamic-protobuf-stream)) |
```
teodordelibasic-db: We list `DynamicRecord` as if it can be passed straight to `ingest_record_offset()` like `ProtoMessage`/`ProtoBytes`, but there's no `From<DynamicRecord> for EncodedRecord` (and there can't be an easy one, since `encode()` is fallible). A reader copying from this row will write `stream.ingest_record_offset(record)`, which won't compile.

Can we change the row to show the actual usage (`ProtoBytes(record.encode()?)`?

rust/sdk/src/builder/stream_builder.rs
```diff
@@ -194,6 +196,18 @@ impl<'a> StreamBuilder<'a> {
...
     }
 
+    /// Select dynamic protobuf record format, for a schema known only at runtime.
+    ///
+    /// Build the `descriptor` with [`crate::schema::descriptor_from_uc_columns`]
+    /// (or fetch it from Unity Catalog) and fill records with
+    /// [`crate::dynamic::DynamicRecord`]. Wire-identical to
+    /// [`compiled_proto`](Self::compiled_proto); see the [`dynamic`](crate::dynamic)
+    /// module for the full pattern.
+    pub fn dynamic_proto(mut self, descriptor: prost_types::DescriptorProto) -> Self {
```
teodordelibasic-db: This now has the same signature as `compiled_proto` and collapses into the same `(RecordType::Proto, Some(*desc))` build arm, the `DynamicProto` variant currently differs only in its `Debug` label, and `message_descriptor()`/`new_record()` work on `compiled_proto` streams too since they only check for a descriptor. So the second method adds a name but no distinct behavior.

I think we should look into `dynamic_proto` accepting the resolved descriptor instead:

```rust
pub fn dynamic_proto(mut self, descriptor: dynamic::MessageDescriptor) -> Self
```

The caller resolves once up front via `dynamic::message_descriptor(&proto)?` (or we add a `descriptor_from_uc_columns` to resolved-descriptor path), and the builder pulls the raw `DescriptorProto` back out for the CreateStream payload with `MessageDescriptor::descriptor_proto().encode_to_vec()`, wire-identical to today. This makes the method genuinely different from `compiled_proto`, fails fast on a bad descriptor at `build()` instead of at the first `new_record()`, and lets us drop the lazy `OnceLock<MessageDescriptor>` on `ZerobusStream` (the descriptor is now supplied, not lazily rebuilt).

rust/sdk/src/dynamic_proto.rs
```diff
@@ -0,0 +1,377 @@
...
+                "record missing required field(s): {}",
+                missing.join(", ")
+            )));
+        }
+        Ok(self.message.encode_to_vec())
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
teodordelibasic-db, line 291: Let's try to improve test coverage by building a descriptor via `descriptor_from_uc_columns()` with a nested struct containing a required scalar, asserting `encode()` reports the nested missing field, then round-tripping a valid record through `DynamicMessage::decode`.

We can also maybe encode the same record via `dynamic_proto` and `compiled_proto` and assert byte-equality.

rust/sdk/src/dynamic_proto.rs
```diff
@@ -0,0 +1,377 @@
...
+    /// Encode the record to protobuf wire bytes, ready to ingest as
+    /// [`ProtoBytes`](crate::ProtoBytes).
+    ///
+    /// # Errors
+    ///
+    /// [`ZerobusError::InvalidArgument`] if any proto2 `required` field is unset
+    /// (`prost-reflect` does not enforce this on encode). The message is
+    /// otherwise valid by construction, since [`set`](Self::set) type-checks each
+    /// field.
+    pub fn encode(&self) -> ZerobusResult<Vec<u8>> {
```
teodordelibasic-db, line 278: Potential follow up to improve perf can be computing "descriptor has any required field (recursively)" once when the `MessageDescriptor` is resolved and short-circuiting the walk when false.

rust/sdk/src/dynamic_proto.rs
andrijast-db: Renamed to dynamic_proto.rs

rust/sdk/src/dynamic_proto.rs
andrijast-db: Added the tests.

rust/sdk/src/builder/stream_builder.rs
andrijast-db: Makes sense. Switched to `MessageDescriptor`.

rust/examples/proto/compiled/single.rs
teodordelibasic-db, line 1: This file reads [`output/orders.descriptor`](https://github.com/databricks/zerobus-sdk/blob/d973c7446a122fa4b9e8a2c2e51692e0627553ff/rust/examples/proto/compiled/single.rs#L40) relative to the process working directory, but the README now tells the reader to run:

```bash
cargo run -p rust-examples-proto --example proto_compiled_single
```

This command panics when we run it from `rust/` or the workspace root:

```text
thread 'main' panicked at examples/proto/compiled/single.rs:137:43:
Failed to read proto descriptor file: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

The same relative read is in `compiled/batch.rs` at line 32.

rust/sdk/src/dynamic_proto.rs
teodordelibasic-db: Let's just create an issue to track this.

rust/sdk/src/builder/stream_builder.rs
```diff
@@ -194,6 +198,23 @@ impl<'a> StreamBuilder<'a> {
         self
     }
 
+    /// Select dynamic protobuf record format, for a schema known only at runtime.
+    ///
+    /// Takes a resolved [`MessageDescriptor`](crate::MessageDescriptor), so you
+    /// can supply one from any source — build it against your own
+    /// [`prost_reflect::DescriptorPool`] (e.g. when the message references types
```
teodordelibasic-db: We don't support descriptors which are not self-contained on server side, so we should probably reword this comment.

rust/sdk/src/dynamic_proto.rs
andrijast-db: Created https://github.com/databricks/zerobus-sdk/issues/520

rust/examples/proto/compiled/single.rs
andrijast-db: Command is back working. I changed that compiled descriptor file is read at compile time instead of runtime.

teodordelibasic-db (dismissed):
LGTM, we just need to rebase and resolve conflicts.

teodordelibasic-db (approved):

PR #466, opened 2026-07-02
[Rust][FFI] Expose ack callback through the C FFI

### LINKED ISSUE #437: [C++][FFI] Add push-based AckCallback (expose ack callback through the C FFI)
## Summary

The C++ SDK currently offers only the **pull-based** ack model (`ingest` → offset → `wait_for_offset` / `flush`). It has **no push-based `AckCallback`**, which Python and Java both expose. This is a follow-up to consider adding an `AckCallback` to the C++ SDK — which first requires exposing it through the **C FFI**, where it does not exist today.

> Not a blocker for any in-flight PR — raised as a follow-up from review of PR #420 (@irinatomic-db, comment on `cpp/include/zerobus/stream.hpp:82`) to evaluate whether it's worth doing.

## Motivation (why a callback matters more for C++ than for Go)

The pull model means the application blocks a thread on `wait_for_offset` / `flush` until the server acknowledges. The cost of that blocking differs sharply by language runtime:

- **Go** — goroutines are userspace-scheduled and multiplexed onto a small pool of OS threads. Blocking a goroutine on `wait_for_offset` just suspends it and yields its OS thread to another goroutine; no kernel involvement. Thousands of goroutines can sit blocked on ack tracking at negligible cost. So the pull-only gap is fine for Go.
- **C++** — `std::thread` maps 1:1 to an OS thread and is kernel-scheduled. Blocking one on `wait_for_offset` **parks that OS thread doing nothing** until the ack arrives. Scaling this across many concurrent streams can get expensive fast (one parked kernel thread per outstanding wait).

The idiomatic C++ answer to "notify me when something happens" is a **callback**, not a blocked thread. With an `AckCallback`, the Rust tokio runtime fires the callback from its own thread pool when the ack arrives — the application thread never blocks waiting for acks, and no extra OS threads are tied up.

The pull model is fully functional for now; this is about whether the push model is worth adding for scale/idiomatic ergonomics.

## Current state across SDKs (from a cross-SDK audit)

| SDK | Push-based `AckCallback`? | Notes |
|---|---|---|
| **Rust core** | trait exists | `rust/sdk/src/callbacks.rs` — `AckCallback { on_ack(offset), on_error(offset, msg) }`. **Not exposed through the C FFI.** |
| **Python** | **yes** | `AckCallback` (subclassable), bridged via `AckCallbackWrapper` in `python/rust/src/common.rs`; set through `StreamConfigurationOptions(ack_callback=...)`. |
| **Java** | **yes** | `AckCallback` interface (`onAck`/`onError`), bridged via `JavaAckCallback` in `rust/jni/src/callbacks.rs`; set through `setAckCallback(...)`. |
| **C++** | **no** | pull-only (`wait_for_offset` / `flush`). |
| **Go** | **no** | pull-only (acceptable per the goroutine reasoning above). |
| **TypeScript** | **no** | pull-only. |

Note the binding mechanism: Python (PyO3) and Java (JNI) bind the Rust core directly, so they could wire the existing core `AckCallback` trait through their own bindings. **C++ (and Go) go through the C FFI, which exposes no ack callback at all** — so the C FFI is the prerequisite. Adding it there would unblock C++, Go, and any future C-FFI consumer at once.

## Implementation considerations to think about

- **Language boundaries** — marshalling a callback across the C FFI: a C function-pointer trampoline + `user_data`, with the wrapper translating `on_ack(offset)` / `on_error(offset, msg)` into the native callback (the same pattern already used for `HeadersProviderCallback`).
- **Callback object lifetime** — the callback fires from tokio's thread pool, asynchronously, and may run after the application's ingest call returns. The C++ callback object must outlive the stream and all pending acks. This is the **same class of use-after-free risk** tracked for the headers-provider `_nowait` path in #436 — whatever ownership/lifetime contract we settle on there should inform this. The C FFI needs a clear "who keeps the callback alive, and until when" rule.
- **Thread-safety** — the callback runs on a core/tokio thread, not the application thread. The user's callback implementation must be thread-safe with respect to its own state; this must be documented (mirroring how `HeadersProvider` thread-safety is documented).
- **Error delivery** — `on_error(offset, message)` semantics: which errors surface here vs. through the existing throwing/`CResult` path, and how retryability is conveyed.
- **Ordering / semantics** — confirm whether `on_ack` is per-record or "acked up to this offset," and that it matches the core trait's contract and the Python/Java behavior, so the model is consistent across SDKs.
- **Free/teardown ordering** — on `close()` / stream free, define whether pending callbacks are drained, dropped, or delivered, and ensure no callback fires after the callback object is released.

## Acceptance criteria (if pursued)

- A push-based ack callback is exposed through the C FFI (`zerobus.h`), with a documented lifetime + thread-safety contract.
- The C++ SDK exposes an `AckCallback` (e.g. via `StreamOptions`) with `on_ack` / `on_error`, consistent in semantics with the Python/Java callbacks and the core trait.
- Lifetime is safe by construction (no callback fires after the callback object is freed); covered by an ASan/stress test.
- Docs + examples + changelog updated per repo policy.
- (Optional) Go can adopt the same FFI callback if desired, though the pull model remains acceptable there.

## References

- Origin: PR #420 review (@irinatomic-db), comment on `cpp/include/zerobus/stream.hpp:82`
- Core trait: `rust/sdk/src/callbacks.rs` (`AckCallback`)
- Python bridge: `python/rust/src/common.rs` (`AckCallbackWrapper`); `python/zerobus/sdk/shared/config.py`
- Java bridge: `rust/jni/src/callbacks.rs` (`JavaAckCallback`); `java/.../AckCallback.java`, `StreamConfigurationOptions.setAckCallback`
- C++ pull-based path: `cpp/src/stream.cpp` (`wait_for_offset`, `flush`)
- Related lifetime concern: #436 (callback-object UAF across the FFI)

zlata-stefanovic-db:
## Summary

Exposes the core push-based **ack callback** through the C FFI. The Rust core already has an `AckCallback` trait (`on_ack` / `on_error`) that Python and Java bind directly, and a background callback-handler task that dispatches acks/errors to it — but there was no way to register a callback across the C boundary, so C++ and Go (and any future C-FFI consumer) were limited to the pull model (`wait_for_offset` / `flush`).

This is the FFI-layer prerequisite: it does not add a C++ or Go public callback API — those build on top of it in follow-up PRs.

## What changed

- **`rust/ffi/src/common.rs`** — new `CallbackAckCallback` that wraps C function pointers and implements the core `AckCallback` trait (mirroring the existing `CallbackHeadersProvider`). `on_error` marshals the Rust `&str` into a call-scoped `CString`; both entry points contain panics with `catch_unwind` (unwinding across `extern "C"` is UB) and log the panic rather than silently swallowing it. `CStreamConfigurationOptions` gains `ack_on_ack`, `ack_on_error`, and `ack_user_data`.
- **`rust/ffi/src/stream.rs`** — `apply_c_stream_options` registers the callback via the existing `builder.ack_callback(...)` only when a pointer is supplied, so both create paths (`zerobus_sdk_create_stream` and `zerobus_sdk_create_stream_with_headers_provider`) pick it up. `zerobus_get_default_config` zero-initializes the new fields.
- **`rust/ffi/zerobus.h`** — regenerated by cbindgen; the fields render as nullable C function pointers. The doc block documents the full contract C consumers build against: threading/serialization, per-record + monotonic ack semantics, error delivery (relayed as-is, no retryability signal, does not replace the pull path), and the teardown/lifetime rule for `user_data`.
- **`go/ffi.go`** — the cgo struct mirror gains the matching trailing fields so its layout stays byte-identical to the Rust struct (a mismatch would let Rust read past the struct Go allocates). Go stays pull-only and leaves the pointers null.
- **Tests** — unit tests covering ack dispatch, error-message marshalling, null-callback no-op, and default-config-has-no-callback.
- **Changelogs** — `rust/ffi` and `go`.

## Design notes

- **Additive, ABI-safe**: existing struct fields keep their order/offsets; new fields are appended and default to null. No signatures changed, and `zerobus.h` struct layout is byte-identical (only doc comments changed).
- **Scoped to gRPC streams**, matching the core `ack_callback` (Arrow Flight has its own independent config struct and is untouched).
- **Lifetime contract**: callbacks fire from a background task, so the callback and its `user_data` must be thread-safe and must outlive the stream. On `close()` the handler is drained up to `callback_max_wait_time_ms` then aborted (or waits indefinitely when unset); callbacks may fire during `close()` but never after it returns.

## Test plan

- `cargo fmt --check`, `cargo clippy`, `cargo test -p zerobus-ffi` (all green; new ack tests pass).
- `go build` / `go vet` against the regenerated FFI (struct layout in sync).

## Follow-ups

Part of #437 (not `Closes`): #437's acceptance criteria also cover a C++ `AckCallback` (via `StreamOptions`) and an ASan/stress test that drives a real callback through a live stream + `close()` to prove no use-after-free. Those land in the follow-up PR that adds the C++/Go public callback API, where an actual `user_data` lifetime exists to exercise — this PR is the FFI prerequisite only.


rust/ffi/src/common.rs
```diff
@@ -257,6 +273,64 @@ impl HeadersProvider for CallbackHeadersProvider {
...
+
+// Safety: We assume the caller's callback and user_data are thread-safe.
+unsafe impl Send for CallbackAckCallback {}
+unsafe impl Sync for CallbackAckCallback {}
+
+impl AckCallback for CallbackAckCallback {
+    fn on_ack(&self, offset_id: OffsetId) {
+        if let Some(cb) = self.on_ack {
+            // Contain panics: unwinding across the FFI boundary is UB.
+            let _ = catch_unwind(AssertUnwindSafe(|| cb(offset_id, self.user_data)));
```
irinatomic-db: `let _ = catch_unwind(...)` is correct for containing UB, but it silently discards a panicking user callback -  a caller whose `on_ack` panics gets zero signal. Please log the `Err` arm, e.g.:

```rust
if catch_unwind(AssertUnwindSafe(|| cb(offset_id, self.user_data))).is_err() {
    tracing::error!(offset_id, "ack on_ack callback panicked; panic contained at FFI boundary");
}
```

Same for on_error at :327.

rust/ffi/src/common.rs
```diff
@@ -257,6 +273,64 @@ impl HeadersProvider for CallbackHeadersProvider {
...
+    ) -> Self {
+        Self {
+            on_ack,
+            on_error,
+            user_data,
+        }
+    }
+}
+
+// Safety: We assume the caller's callback and user_data are thread-safe.
```
irinatomic-db: This `// we assume the caller's callback and user_data are thread-safe comment` is too vague.

The dispatch task is a single `tokio::spawn` with a sequential `recv` loop (`rust/sdk/src/stream/grpc/callback_handler.rs:26`), so `on_ack`/`on_error` are serialized - never re-entered concurrently by the SDK. The real requirement is narrower than "be thread-safe": 
(a) calls arrive on a background thread, so `user_data` shared with app threads needs its own synchronization 
(b) the SDK never invokes the callback concurrently with itself. Please spell both out here so the `unsafe impl`is justified precisely

rust/ffi/zerobus.h
```diff
@@ -128,6 +128,16 @@ typedef struct CStreamConfigurationOptions {
   bool has_stream_paused_max_wait_time_ms;
   uint64_t callback_max_wait_time_ms;
   bool has_callback_max_wait_time_ms;
+  /**
```
irinatomic-db, line 131: The core trait doc warns callbacks run synchronously on the dispatch task, so slow callbacks accumulate ack backlog (`rust/sdk/src/callbacks.rs:61`). That warning isn't propagated to the C header, and it matters more here than for the headers provider because acks are the hot path. Please add to this block: 
(1) callbacks fire on a background SDK thread, serialized (no concurrent re-entry); 
(2) keep them lightweight - offload heavy work to your own queue. 
Since this header is the contract C++/Go build against, it needs to be self-contained.

rust/ffi/zerobus.h
```diff
@@ -128,6 +128,16 @@ typedef struct CStreamConfigurationOptions {
...
+   * Optional ack callback. When either pointer is non-null, acks/errors are
+   * delivered asynchronously instead of only via wait_for_offset / flush.
+   * ack_user_data is passed to each call and must outlive the stream.
+   * For ack_on_error, error_message is valid only during the call (copy to keep).
+   * Callback implementations must not unwind across the C boundary.
+   */
+  void (*ack_on_ack)(int64_t offset_id, void *user_data);
```
irinatomic-db, line 144: The [issue](https://github.com/databricks/zerobus-sdk/issues/437) explicitly asked to confirm and document whether `on_ack(offset)` means "this record" or "everything acked through this offset," so the model matches Python/Java. The core emits one `CallbackMessage::Ack(offset)` per ack (callback_handler.rs:37) and this PR forwards it faithfully, but the header says nothing. Please document the exact semantics on this pointer (and that acks are monotonic, so a later offset implies all prior). This is a contract C consumers will hard-code against.

rust/ffi/zerobus.h
```diff
@@ -128,6 +128,16 @@ typedef struct CStreamConfigurationOptions {
...
+  void (*ack_on_error)(int64_t offset_id, const char *error_message, void *user_data);
```
irinatomic-db, line 145: The issue's "error delivery" consideration asks which errors surface through `on_error` vs the existing CResult/throwing path, and how retryability is conveyed. Right now `on_error` just relays whatever the core sends with no documented classification. Please document: which error classes reach on_error(offset, msg), whether they overlap with errors returned from ingest/flush, and whether the message conveys retryability. Without this a C++ consumer can't tell whether to also check the pull path.

rust/ffi/src/tests.rs
```diff
@@ -442,6 +442,78 @@ mod tests {
...
+        LAST_ERROR_OFFSET.store(offset_id, AtomicOrdering::SeqCst);
+        // The message must be a valid NUL-terminated string for the call's duration.
+        assert!(!error_message.is_null());
+        let msg = unsafe { CStr::from_ptr(error_message) }
+            .to_str()
+            .expect("error message must be valid UTF-8");
+        assert_eq!(msg, "boom");
+    }
+
+    #[test]
```
irinatomic-db, line 479: These unit tests exercise the bridge in isolation but nothing drives a callback through a live stream + close(), which is exactly where the "user_data must outlive the stream" contract and the drain-then-abort teardown (`rust/sdk/src/stream/grpc/close.rs:88`) actually get tested. Please add a test that: registers a real callback, ingests + closes, and asserts (a) callbacks stop after close() returns and (b) no use-after-free

rust/ffi/zerobus.h
```diff
@@ -128,6 +128,16 @@ typedef struct CStreamConfigurationOptions {
   bool has_stream_paused_max_wait_time_ms;
   uint64_t callback_max_wait_time_ms;
   bool has_callback_max_wait_time_ms;
+  /**
+   * Optional ack callback. When either pointer is non-null, acks/errors are
+   * delivered asynchronously instead of only via wait_for_offset / flush.
+   * ack_user_data is passed to each call and must outlive the stream.
```
irinatomic-db: Good that the lifetime contract is stated, but the teardown behavior isn't: on close() the handler task is drained up to callback_max_wait_time_ms then aborted (close.rs:88), so callbacks may still fire during close but never after free. Please document that here - "who keeps user_data alive, and until when" was called out in the issue, and C consumers need to know when it's safe to release user_data.

rust/ffi/src/common.rs
zlata-stefanovic-db: Both fixed

rust/ffi/src/common.rs
zlata-stefanovic-db: Made it more precise now

rust/ffi/zerobus.h
zlata-stefanovic-db: Header now says: fires once per record, in order, and acks are monotonic (getting offset N means everything <= N is confirmed).

rust/ffi/zerobus.h
zlata-stefanovic-db: Header now says: on_error relays the core error text as-is, carries no retryability signal, and the same failure can also show up via ingest/flush - so we don't rely on on_error alone.

rust/ffi/zerobus.h
zlata-stefanovic-db: Header now says: callbacks may fire during close() but never after it returns - so keep user_data alive until close() returns, then free it.

rust/ffi/zerobus.h
zlata-stefanovic-db: Added to the header: callbacks are serialized, keep them lightweight.

rust/ffi/src/tests.rs
zlata-stefanovic-db: Good call, but I skipped this one on purpose - I'd like to land this test with the follow-up rather than here. Driving a callback through a live stream + close() and asserting no use-after-free needs a real user_data lifetime to violate - which only exists once there's a Go/C++ public callback API. 

This PR is FFI-only with no C consumer to drive it, and #437 gates the ASan criterion on the full effort. I've changed "Closes #437" to "Part of #437" and noted the deferred ASan/stress test in the description so it's tracked, not dropped. Sound OK?

Follow up: https://github.com/databricks/zerobus-sdk/issues/469

irinatomic-db: Sounds good

irinatomic-db (approved):

PR #474, opened 2026-07-06
[C++] Arrow Flight stream (Beta)

zlata-stefanovic-db:
## Summary

Adds the C++ Arrow Flight ingestion stream (Beta) on top of the current core, and fixes a set of doc/type inconsistencies found while reviewing it.

### Arrow Flight stream (Beta)
- New `ArrowStream` class (`arrow_stream.hpp`/`.cpp`): move-only handle over the `zerobus_arrow_stream_*` FFI, with `ingest_batch`, `wait_for_offset`, `flush`, `get_unacked_batches`, `close`, and `is_closed`.
- `Sdk::create_arrow_stream` overloads for OAuth client-credentials and custom headers-provider auth.
- `ArrowStreamOptions` config struct + `IpcCompression` enum, converted to the C config via `to_c()`.

### Doc/type fixes
- Correct an `ArrowStream` reference in the `create_arrow_stream` comment (was "Stream").
- Document all three `stream_paused_max_wait_time_ms` cases (`nullopt` / `0` / `> 0`) to match the FFI header and Rust core.
- Change `stream_paused_max_wait_time_ms` to `std::optional<std::uint64_t>` to match the core's `Option<u64>`, and cast the sentinel explicitly in `to_c()` so the ternary doesn't fold both branches to unsigned.
- Correct the header note on why `zerobus_arrow_stream_ingest_batch_via_record_batch` is omitted: it is a functional duplicate of `zerobus_arrow_stream_ingest_batch` (same IPC-bytes signature, both deserialise then re-encode), not a method taking an unconstructable `RecordBatch*`.

### Tests (config-conversion unit tests only)
This PR includes only server-free unit tests of the config-conversion logic — the layer that ships in this diff:
- `arrow_config_defaults_test` pins every `to_c(ArrowStreamOptions{})` scalar to `zerobus_arrow_get_default_config()`, turning default drift into a build failure.
- `arrow_config_convert_test` covers the `stream_paused_max_wait_time_ms` conversion logic: set-value round-trip, `nullopt` → `-1` sentinel, the `INT64_MAX` boundary, and the `> INT64_MAX` overflow rejection.

## Scope / follow-ups

This is an incremental split of the C++ SDK, so some artifacts land in later stacked PRs rather than here:

- **README, `NEXT_CHANGELOG.md`/`CHANGELOG.md`, and runnable examples** for the Arrow path are deferred to the following stacked PRs. The `cpp/` tree has no changelog/README/`examples/` yet; they will be bootstrapped alongside the rest of the SDK docs.
- **Behavioral / lifecycle tests** (live-`ArrowStream` coverage: move single-close, `close()` idempotency, post-close operation guards, `get_unacked_batches` after a failed close) are deferred to the same later PRs, since they exercise a live stream and generally require a server. The unit tests above intentionally cover only the config-conversion logic introduced in this diff.

## Notes
- The API is Beta and may change before GA.
- `stream_paused_max_wait_time_ms` changing from signed to unsigned is a source-level change to the (unreleased) Beta config struct.

### DIFF
--- a/cpp/include/zerobus/config.hpp
+++ b/cpp/include/zerobus/config.hpp
@@ -46,17 +46,21 @@ struct StreamOptions {
 };
 
 /// Arrow IPC compression codec. Matches the `ipc_compression` field encoding
-/// in `CArrowStreamConfigurationOptions` (-1 = None, 0 = LZ4_FRAME, 1 = ZSTD).
+/// in `CArrowStreamConfigurationOptions` (-1 = none, 0 = LZ4_FRAME, 1 = ZSTD).
+///
+/// Named `NoCompression`, not `None`, to avoid X11's `#define None 0L`, which
+/// the preprocessor expands even inside an `enum class`.
 enum class IpcCompression : std::int32_t {
-  None = -1,
+  NoCompression = -1,
   Lz4Frame = 0,
   Zstd = 1,
 };
 
 /// Configuration for an Arrow Flight ingestion stream (Beta).
 ///
-/// Defaults mirror `zerobus_arrow_get_default_config()`; keep them in sync by
-/// hand (not yet pinned by `config_defaults_test`).
+/// The scalar defaults below are hand-kept in sync with the Rust core and sent
+/// to the FFI verbatim (see `to_c()`). `arrow_config_defaults_test` fails the
+/// build if they drift from `zerobus_arrow_get_default_config()`.
 struct ArrowStreamOptions {
   /// Maximum number of in-flight (unacknowledged) batches.
   std::size_t max_inflight_batches = 1'000;
@@ -75,7 +79,7 @@ struct ArrowStreamOptions {
   /// Connection establishment timeout.
   std::uint64_t connection_timeout_ms = 30'000;
   /// Arrow IPC compression codec.
-  IpcCompression ipc_compression = IpcCompression::None;
+  IpcCompression ipc_compression = IpcCompression::NoCompression;
   /// Max wait during a server-initiated pause before recovering. `nullopt` =
   /// full server duration; `0` = recover immediately; `x` = min(x, server).
   std::optional<std::uint64_t> stream_paused_max_wait_time_ms;
--- a/cpp/include/zerobus/config.hpp
+++ b/cpp/include/zerobus/config.hpp
@@ -14,7 +14,7 @@ enum class RecordType : std::int32_t {
   Json = 2,
 };
 
-/// Configuration for a (non-Arrow) ingestion stream.
+/// Configuration for a gRPC ingestion stream.
 ///
 /// The scalar defaults below are hand-kept in sync with the Rust core and sent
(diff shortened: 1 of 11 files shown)


cpp/src/detail/config_convert.hpp
```diff
@@ -41,6 +44,36 @@ inline CStreamConfigurationOptions to_c(const StreamOptions& opts) {
   return c;
 }
 
+/// Build the C Arrow stream-config struct from `ArrowStreamOptions`.
+inline CArrowStreamConfigurationOptions to_c(const ArrowStreamOptions& opts) {
+  CArrowStreamConfigurationOptions c = zerobus_arrow_get_default_config();
+  c.max_inflight_batches = opts.max_inflight_batches;
```
teodordelibasic-db, line 57: `to_c(const ArrowStreamOptions&)` forwards `opts.max_inflight_batches` straight into the C config. The core then builds a bounded Tokio channel with that capacity (`mpsc::channel::<...>(options.max_inflight_batches)` in `arrow_stream.rs`), and a capacity of `0` is invalid there and panics. The panic happens during `create_arrow_stream` (the connection is established eagerly), so the FFI panic guard catches it and the caller does get a non-retryable `ZerobusException` rather than a process crash. But the message is the panic text, not something that tells the caller the real problem is that the minimum is `1`.

Since this file already guards the two other invalid inputs client-side (empty `schema_ipc_bytes`, and `stream_paused_max_wait_time_ms` above `INT64_MAX`), could we reject `max_inflight_batches == 0` here the same way, with a message that names the minimum, and add a conversion test for it?

```cpp
if (opts.max_inflight_batches == 0) {
  throw ZerobusException("max_inflight_batches must be at least 1", false);
}
```

This issue may be present for some other config fields as well.

cpp/include/zerobus/config.hpp
```diff
@@ -14,7 +14,7 @@ enum class RecordType : std::int32_t {
   Json = 2,
 };
 
-/// Configuration for an ingestion stream.
+/// Configuration for a (non-Arrow) ingestion stream.
```
teodordelibasic-db: Let's use `gRPC` for `non-Arrow`, even though it is a bit strange. I don't think there is a better way to distinguish.

teodordelibasic-db (commented):
Looks good, let's just check regarding erroneous config values.

cpp/src/detail/config_convert.hpp
zlata-stefanovic-db: Fixed. Added a client-side guard at the top of to_c(const ArrowStreamOptions&), matching the existing schema_ipc_bytes / stream_paused_max_wait_time_ms guards. Also traced other fields and found this is the only one that panics.

cpp/include/zerobus/config.hpp
zlata-stefanovic-db: Done, renamed

teodordelibasic-db (approved):
LGTM

PR #483, opened 2026-07-08
[Rust][FFI] Move header helpers to common.rs

### LINKED ISSUE #468: [FFI] Move non-Arrow header functions out of arrow.rs into a common file
Follow-up from a review comment on #420: https://github.com/databricks/zerobus-sdk/pull/420#discussion_r3513892190 (by @davidtosovic-db):

> Nit: This is not important and should not be done in this PR.
> All of these header functions don't really have to do anything with arrow. We should probably move them to the common file.

The header-related functions in `rust/ffi/src/arrow.rs` (around line 670) aren't Arrow-specific, so they'd be better placed in the shared/common FFI module rather than the Arrow file.

This was deliberately deferred out of #420 as a non-blocking nit. No behavior change intended — purely relocating the functions to improve organization.

zlata-stefanovic-db:
Fixes #468.

Follow-up from a deferred review nit on #420 ([discussion](https://github.com/databricks/zerobus-sdk/pull/420#discussion_r3513892190)): the header-callback helpers in `rust/ffi/src/arrow.rs` aren't Arrow-specific, so they belong in the shared FFI module.

## What changed

Moved `zerobus_alloc_header_array`, `zerobus_alloc_cstring`, and `zerobus_free_headers` out of `arrow.rs` into `common.rs`, placing them next to the `CHeader`/`CHeaders` types and the `CallbackHeadersProvider` machinery they serve. The functions, signatures, and semantics are byte-for-byte identical — this is a pure relocation.

Along with the move:
- Dropped the now-unused `use std::ffi::CString;` from `arrow.rs`.
- Switched the two `crate::zerobus_free_headers(...)` call sites in `common.rs` to the now-local path.
- Rewrote the load-bearing cbindgen module-ordering comment in `lib.rs`. It previously stated `common` had no exported functions and that `zerobus.h` must stay byte-identical; both are now updated to reflect that `common` leads with the three header helpers and that C symbols link by name (declaration order is header churn, not an ABI contract).
- Regenerated `zerobus.h` via the `build.rs` cbindgen step. The three declarations now precede the Arrow functions; their content is unchanged.

## No behavior / ABI change

C symbols resolve by name at link time, so the reordering in `zerobus.h` does not affect the Go/Java ABI.

## Testing

- `make fmt` clean; `cargo build` clean (no unused-import warnings).
- `cargo test -p zerobus-ffi` — 71/71 pass, including `test_every_extern_c_entry_point_is_guarded` (guard-coverage scan already covers `common.rs`).
- `cargo clippy -p zerobus-ffi` — only a pre-existing `module_inception` warning in `tests.rs`, unrelated to this change.

Added an Internal Changes entry to `rust/ffi/NEXT_CHANGELOG.md`.

### DIFF
--- a/rust/ffi/NEXT_CHANGELOG.md
+++ b/rust/ffi/NEXT_CHANGELOG.md
@@ -15,6 +15,8 @@
 
 ### Internal Changes
 
+- Move the header-callback helpers `zerobus_alloc_header_array`, `zerobus_alloc_cstring`, and `zerobus_free_headers` from `arrow.rs` into the shared `common.rs` module, next to the `CHeader`/`CHeaders` types they serve. No behavior, signature, or ABI change; their declarations move ahead of the Arrow functions in the generated `zerobus.h`.
+
 ### Behavior Changes
 
 ### Breaking Changes
--- a/rust/ffi/src/arrow.rs
+++ b/rust/ffi/src/arrow.rs
@@ -6,7 +6,6 @@ use bytes::Bytes;
 use databricks_zerobus_ingest_sdk::{
     HeadersProvider, RecordBatch, StreamBuilder, ZerobusArrowStream, ZerobusError, ZerobusResult,
 };
-use std::ffi::CString;
 use std::os::raw::c_char;
 use std::ptr;
 use std::sync::Arc;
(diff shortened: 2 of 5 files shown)


teodordelibasic-db (approved):
LGTM

Issue #469, opened 2026-07-03
[FFI][test] Add live-stream ack callback teardown / use-after-free test


zlata-stefanovic-db:
## Context

Deferred from PR #466 (which exposed the ack callback through the C FFI). See review thread: https://github.com/databricks/zerobus-sdk/pull/466#discussion_r3518491341

PR #466 added unit tests that exercise the `CallbackAckCallback` bridge in isolation, but nothing drives a callback through a **live stream + `close()`**, which is where the lifetime contract and the drain-then-abort teardown (`rust/sdk/src/stream/grpc/close.rs`) actually get exercised.

This test was intentionally not landed in #466 because meaningfully validating the `user_data` use-after-free risk needs a real `user_data` lifetime to violate — which only exists once there is a Go/C++ public callback API on top of the FFI. #466 is the FFI prerequisite only.

## What to add

A test (with ASan where practical) that:
1. Registers a real ack callback with live `user_data`.
2. Ingests records and calls `close()`.
3. Asserts callbacks **stop** after `close()` returns (none fire post-teardown).
4. Asserts **no use-after-free** — `user_data` is safe to release once `close()` returns.

Should cover both teardown paths: drain up to `callback_max_wait_time_ms` then abort, and the wait-indefinitely case when that option is unset.

## References

- Part of #437 (the ASan/stress acceptance criterion)
- Callback dispatch: `rust/sdk/src/stream/grpc/callback_handler.rs`
- Teardown: `rust/sdk/src/stream/grpc/close.rs`
- FFI bridge: `rust/ffi/src/common.rs` (`CallbackAckCallback`)

PR #484, opened 2026-07-08
[C++] Expose gRPC ack callback in the C++ SDK

zlata-stefanovic-db:
## What

Exposes the gRPC ack callback in the C++ SDK. Adds the `AckCallback` abstract class and the `StreamOptions::ack_callback` field, so callers can track durability asynchronously via ack/error notifications instead of blocking in `wait_for_offset()` / `flush()`.

- `AckCallback` with `on_ack(offset)` / `on_error(offset, message)`, registered through `StreamOptions::ack_callback`.
- `extern "C"` trampolines that recover the callback from `user_data` and forward to it. They are `noexcept` and contain (and log) any exception at the C FFI boundary — a callback that throws terminates rather than unwinding across the boundary (UB).
- The `Stream` holds a `shared_ptr` to the callback for its lifetime; move ctor/assignment carry it alongside the handle and headers provider.
- `StreamOptions::callback_wait_forever` (default `false`): when set, `close()` drains the callback task with no deadline instead of aborting it on the `callback_max_wait_time_ms` timeout. It is the only option that guarantees no `AckCallback` is still running after `close()` returns; the tradeoff is that `close()` blocks on a wedged callback. It overrides `callback_max_wait_time_ms`, wiring the FFI presence flag to `false` so the core reads the budget as `None` (drain indefinitely).
- Documented ordering, threading, and lifetime contracts on the public header.
- Tests covering `to_c()` wiring (trampolines installed only when a callback is set), `callback_wait_forever` clearing the wait-time presence flag, trampoline dispatch, null `user_data`, null `error_message`, and exception containment.

## Docs / contract fixes along the way

While aligning the C++ lifetime docs with the core, I found the C FFI header (`rust/ffi/zerobus.h` and its cbindgen source `common.rs`) over-promised that callbacks "never fire after `close()` returns." The core aborts rather than joins the synchronous callback task on the `callback_max_wait_time_ms` timeout, and `abort()` only cancels at an await point — so a synchronously-running callback can still be executing after `close()` returns. Corrected the FFI wording (relevant to Go/Java, which link the same header) to say `ack_user_data` must stay alive until the callback object is destroyed, not merely until `close()` returns. Also fixed the `config.hpp` description of `callback_max_wait_time_ms` (it bounds the ack-callback drain, not a headers-provider callback) and added the caveat that a terminal error may also surface via `ingest`/`flush`.

## Verification

- `make test` in `cpp/`: all test executables passed (`config_defaults_test`, `embedded_nul_test`, `ack_callback_test`).
- `cargo check` on the FFI crate: clean.
- `clang-format --Werror` clean on all touched C++ files.

## Related

Unblocks #469 ([FFI][test] Add live-stream ack callback teardown / use-after-free test). This PR pins down the exact teardown/use-after-free window — `close()` aborts the callback task at `callback_max_wait_time_ms` and `abort()` only cancels at an await, so a synchronously-running callback can outlive `close()` — and adds `callback_wait_forever` as the deterministic close-the-window option. That behavior and the corrected FFI lifetime contract are the prerequisites for writing the live-stream teardown test #469 asks for.


cpp/include/zerobus/stream.hpp
```diff
@@ -25,10 +26,12 @@ class Sdk;
...
 ///   whereas the destructor swallows it.
 /// - Closing flushes synchronously and can block up to the stream's
-///   `flush_timeout_ms` (default 5 minutes) if the server is unresponsive. A
-///   `Stream` that falls out of scope therefore drags that blocking close into
-///   the destructor at an unpredictable point. Call `close()` at a controlled
-///   point in your code instead.
+///   `flush_timeout_ms` (default 5 minutes) if the server is unresponsive, and
+///   then drains any registered `ack_callback` task per
+///   `StreamOptions::callback_wait_policy` (which may be unbounded). A `Stream`
+///   that falls out of scope therefore drags that blocking close into the
```
irinatomic-db, line 32: Nit: with `CallbackWaitPolicy::forever()`, a `Stream` dropped during exception unwinding drags an unbounded wait into the destructor - deadlocking the unwind on a wedged callback. The general lifetime warnings already cover this indirectly; a one-liner here calling it out specifically would help the "safe option" reader who wouldn't otherwise think of unwinding as the risky path.

cpp/tests/ack_callback_test.cpp
```diff
@@ -0,0 +1,165 @@
...
+int g_failures = 0;
+
+void check(const char* what, bool ok) {
+  if (!ok) {
+    std::fprintf(stderr, "FAIL: %s\n", what);
+    ++g_failures;
+  }
+}
+
+// Records what the trampolines deliver.
```
irinatomic-db: Nit: `last_ack` / `last_error_offset` / counters are plain (non-atomic). Fine for the current pure-synchronous tests, but the live-stream teardown test tracked in #469 will fire callbacks from the real core (background thread) and race on these. 

Maybe add a `// synchronous tests only` comment or upgrade to `std::atomic` now so #469 doesn't have to touch this file.

irinatomic-db (dismissed):
LGTM. Left some nits

cpp/include/zerobus/stream.hpp
zlata-stefanovic-db: Done, added the additional comment

cpp/tests/ack_callback_test.cpp
zlata-stefanovic-db: Added the comment

irinatomic-db (approved):
LGTM

Issue #467, opened 2026-07-02
[PureGo] Design a pure-Go SDK (no cgo/FFI)


zlata-stefanovic-db:
## Summary

Build a **pure-Go** Zerobus SDK: a native Go implementation that talks gRPC to Zerobus directly, with no cgo or Rust FFI dependency.

**Outcome:** `purego/v0.1.0` released 2026-08-07 (#700). Arrow/Flight is the remaining protocol and is in progress.

## Motivation

The existing Go SDK (`go/`) binds the Rust core through `rust/ffi/` using cgo and a prebuilt static library. A pure-Go SDK removes several costs:

- **Simpler builds** — ordinary Go tooling, cross-compilation, and `CGO_ENABLED=0` builds.
- **Independent releases** — no dependency on an `ffi/v*` release before Go can consume core changes.
- **Smaller memory-safety surface** — no `runtime.Pinner`, `cgo.Handle` registry, opaque Rust handles, or manual cross-FFI cleanup.
- **Idiomatic Go** — goroutines, channels, functional options, and `context.Context` work end to end without FFI marshalling.

**OSS integrations.** Beyond those costs, a native Go SDK is what makes it viable to ship Zerobus plugins upstream — a [Telegraf](https://github.com/influxdata/telegraf) output plugin and a [Grafana Alloy](https://github.com/grafana/alloy) component. Both compile plugins into a single binary that they release cross-compiled for every supported platform, and the current Go SDK asks them to take on not just cgo but a prebuilt Rust archive published on our own `ffi/v*` cadence. For Telegraf that is disqualifying: `CGO_ENABLED=0` is hardcoded in its build to keep cross-compilation and static linking working, so the plugin would never make the official release — leaving users to compile Telegraf themselves, or to run Zerobus out-of-process behind the `execd` shim with its own binary, config file, and process hop per metric. Alloy does enable cgo and maintains a C cross-compiler image, but gates cgo-only dependencies behind `//go:build cgo` with stubs that error in no-cgo builds, so a cgo-based component would be second-class there too. A pure-Go module drops into both as an ordinary `go get` dependency.

The tradeoff is that streaming, acknowledgement, recovery, and ingestion behavior must be implemented and kept in parity with the Rust core.

## What shipped

Grouped by layer, bottom up. Each layer took several PRs, and the ingestion core was deliberately split into a four-part stack to keep reviews tractable.

**1. Transport and handshake** — #463, #489, #636

[...]

PR #489, opened 2026-07-09
[PureGo] Add GracefulClose to transport Stream

### LINKED ISSUE #487: [Go][PureGo] Add GracefulClose(ctx) to the transport Stream
## Summary

Add a `GracefulClose(ctx)` method to the pure-Go transport `Stream` that encapsulates the clean-shutdown sequence: half-close the send side (`CloseSend`), drain remaining responses to `io.EOF`, then release resources (`Close`).

## Motivation

Today the transport layer exposes only `Close()`, which fires the context cancel. Closing without first draining makes the server observe an abrupt `RST_STREAM` rather than an orderly `END_STREAM`. The correct sequence (`CloseSend` → drain to `io.EOF` → `Close`) is documented on `Stream.CloseSend`, but nothing encapsulates it — so every higher layer (ingest, recovery) has to re-implement it and can get the order wrong.

This is the same mistake the Rust gRPC path had to unwind (it drains inline on close for this reason). Not a bug today since there is no higher layer yet, but worth fixing before ingest lands and starts depending on it.

Raised in review of #463 (thread on `ephemeral_stream.go`).

## Proposal

- Add the generic drain-to-EOF logic on `rawStream` so the future Arrow/Flight `wireStream` inherits it for free.
- Expose `func (s *Stream) GracefulClose(ctx context.Context) error` on the proto/JSON stream.
- `ctx` bounds the drain: on expiry or a non-EOF stream error, hard-abort like `Close` and return the cause; a clean drain returns `nil`.
- Keep `Close()` as the hard-abort escape hatch (unchanged).

## Part of

Pure-Go SDK effort — tracking issue #467 (transport foundation landed in #463).

zlata-stefanovic-db:

Adds `GracefulClose(ctx)` to the pure-Go transport `Stream`, encapsulating the clean-shutdown sequence so higher layers don't re-implement it and get the order wrong.

- **`rawstream.go`** — generic `gracefulClose(ctx)` on `rawStream`: half-close the send side (`CloseSend`), drain and discard remaining responses to `io.EOF`, then release resources (`close`). The future Arrow/Flight `wireStream` inherits it for free. Draining to EOF lets the server see an orderly close; a bare `Close` cancels the context and the server sees an abrupt reset instead.
- **`ephemeral_stream.go`** — exposes `func (s *Stream) GracefulClose(ctx context.Context) error`; `CloseSend`/`Close` docs updated to cross-reference it. `Close()` stays as the hard-abort escape hatch.
- `ctx` bounds the drain: on expiry or a non-EOF stream error it hard-aborts (via `close`) and returns the cause (ctx error preferred), else `nil`. The ctx is bridged to `close` so a caller deadline unblocks the in-flight `recv`, which otherwise waits on the stream's Close-only context.

## Why

Before this, the transport layer exposed only `Close()`, which fires the context cancel — closing without a preceding drain makes the server observe an abrupt `RST_STREAM` rather than an orderly `END_STREAM`. The correct sequence was documented on `CloseSend` but nothing encapsulated it, so every higher layer (ingest, recovery) would have to re-implement it. This is the same trap the Rust gRPC path had to unwind.

Raised in review of #463.

## How is this tested?

`transport_test.go`:
- `TestStreamGracefulClose` — server ends the stream on half-close; drain reaches a clean `io.EOF` and `GracefulClose` returns nil.
- `TestStreamGracefulCloseDrainsPending` — an in-flight ack precedes `io.EOF`; the drain must discard it and keep going.
- `TestStreamGracefulCloseHonorsDeadline` — server never ends the stream (new `hangDrain` fake-server flag); `GracefulClose` returns its ctx error promptly and leaves the stream torn down.

`gofmt`, `go vet`, and `go test -race ./...` all clean.

## Notes for reviewers

- **No changelog entry** — all code is under `purego/internal/`, no public API yet.
- **Timing** — @teodordelibasic-db suggested adding this "once ingest lands." Pulling it slightly forward so the ingestion layer has a correct primitive to call from day one rather than open-coding the sequence; happy to defer into the ingestion PR instead if preferred.

Closes #487. Part of #467.


purego/internal/transport/ephemeral_stream.go
```diff
@@ -168,11 +168,22 @@ func (s *Stream) Recv() (*zerobuspb.EphemeralStreamResponse, error) { return s.r
...
+// GracefulClose half-closes the send side, drains remaining responses until the
+// server ends the stream, then releases resources. Prefer it over Close when done
+// sending: draining to end-of-stream lets the server see an orderly close rather
+// than the abrupt reset Close produces.
+//
+// ctx bounds the drain; on expiry or a stream error it hard-aborts like Close and
+// returns the cause, else nil. Must not be called concurrently with Recv, and no
+// Send may follow (the send side is half-closed). A later Close is a no-op, since
+// GracefulClose always releases resources before it returns.
+func (s *Stream) GracefulClose(ctx context.Context) error { return s.gracefulClose(ctx) }
```
davidtosovic-db, line 183: Nothing guarantees that this context has a deadline. Meaning we could hang here if the server is not sending acks on `recv()` channel.
For `Open(....)` you have explicit guards against this, where you check for a deadline and apply a default one if it's not present.
```go
	openCtx := ctx
	if _, ok := ctx.Deadline(); !ok {
		var cancelOpen context.CancelFunc
		openCtx, cancelOpen = context.WithTimeout(ctx, defaultHandshakeTimeout)
		defer cancelOpen()
	}
```

For safety we should have the same or equivalent mechanism here.

purego/internal/transport/rawstream.go
```diff
@@ -94,6 +94,42 @@ func (s *rawStream[Req, Resp]) close() {
...
+// then releases resources. Draining to EOF lets the server see an orderly close;
+// a bare close cancels the context and the server sees an abrupt reset instead.
+//
+// ctx bounds the drain: on ctx expiry or a non-EOF error it hard-aborts and
+// returns the cause (ctx error preferred); a clean drain returns nil. Not safe to
+// call concurrently with recv, and no send may follow (the send side is
+// half-closed). Every return path calls close first, so a later close is a no-op.
+func (s *rawStream[Req, Resp]) gracefulClose(ctx context.Context) error {
+	if err := s.closeSend(); err != nil {
+		s.close() // can't half-close cleanly; hard-abort
```
davidtosovic-db: Do we have a test for this case? (when `s.closeSend()` fails)

purego/internal/transport/transport_test.go
```diff
@@ -542,18 +548,79 @@ func TestStreamGracefulClose(t *testing.T) {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
+	defer cancel()
+
+	start := time.Now()
+	err = stream.GracefulClose(ctx)
+	if err == nil {
+		t.Fatal("GracefulClose against a server that never ends the stream: got nil, want deadline error")
+	}
+	if elapsed := time.Since(start); elapsed > 5*time.Second {
```
davidtosovic-db: Nit: not blocking for this PR, but consider improving the test to be more tight + less flaky (in this change or another one).

This looks to be a "we hope the timing works fine for us" test.
We're expecting the cancellation to happen after ~200ms, but are asserting on 5s. (so a 4.9s is ok, even though we expect 200ms, at least that's what the test asserts)

In general, we rely here on wall-clock, so this is a flaky test.

davidtosovic-db (dismissed):

purego/internal/transport/rawstream.go
teodordelibasic-db: Let's tidy up these inline comments.

purego/internal/transport/transport_test.go
```diff
@@ -99,6 +102,10 @@ func (f *fakeServer) EphemeralStream(stream zerobuspb.Zerobus_EphemeralStreamSer
 	for {
 		req, err := stream.Recv()
 		if err == io.EOF {
+			if f.hangDrain {
+				<-stream.Context().Done() // withhold the clean end
```
teodordelibasic-db: ditto

purego/internal/transport/transport_test.go
```diff
@@ -542,18 +548,79 @@ func TestStreamGracefulClose(t *testing.T) {
...
-	// Drain to the clean end-of-stream the server sends after the half-close.
-	for {
-		_, err := stream.Recv()
-		if err == io.EOF {
-			break
-		}
-		if err != nil {
-			t.Fatalf("Recv during drain: got %v, want io.EOF", err)
-		}
+	stream.Close() // idempotent no-op after a graceful shutdown
```
teodordelibasic-db: ditto

purego/internal/transport/rawstream.go
zlata-stefanovic-db: No, adding this now.

purego/internal/transport/transport_test.go
zlata-stefanovic-db: Replaced the wall-clock test entirely with assertion of what the deadline mechanism actually returned rather than inferring it from timing, removing the flakiness entirely. Now we check if the error actually returned the deadline exceeded message.

purego/internal/transport/ephemeral_stream.go
teodordelibasic-db, line 37: Unrelated to this PR and we can discuss and do a follow up if we decide to do so, but since we will probably have an `ArrowStream`, I think leaving this as  `Stream` might be confusing. Maybe more natural would be `GrpcStream`. What do you think?

purego/internal/transport/raw_stream.go
teodordelibasic-db, line 1: Also unrelated and potential follow up, do you think `raw_stream.go` might be more per Go file naming standards?

teodordelibasic-db (commented):
First pass just regarding nits and styling.

purego/internal/transport/ephemeral_stream.go
zlata-stefanovic-db: Maybe we can name it EphemeralStream as the rpc it wraps? Because Arrow is also a grpc stream technically so it might be confusing - so we'd have rawStream as a shared base and then 1. EphemeralStream and 2. ArrowStream

purego/internal/transport/raw_stream.go
zlata-stefanovic-db: Yes, I can rename it

davidtosovic-db (dismissed):

davidtosovic-db (approved):

PR #490, opened 2026-07-09
[C++] SDK README and contributor docs

zlata-stefanovic-db:
## Summary

Documentation for the C++ SDK: `cpp/README.md` (build, install, quickstart for
JSON / proto / Arrow Flight ingestion, ingestion-format guidance, credential
model, API overview, `StreamOptions` / `ArrowStreamOptions` configuration
tables, and an HTTP-proxy note), `cpp/CLAUDE.md` (contributor guide: FFI
boundary, RAII/memory ownership, thread-safety, release process), and
`cpp/CONTRIBUTING.md` (C++-specific dev setup and workflow). Also adds the C++
rows to the root `README.md` and `CLAUDE.md`, reconciles the root Arrow-Flight
and `examples/arrow/` notes with the C++ SDK's `0.1.0` state, and adds a
`NEXT_CHANGELOG.md` entry. Docs only.

Revives and supersedes the closed #418 (`split/415/docs`), rebased onto current
`main`. The docs were re-verified against the current `cpp/` tree, so they
describe the SDK as it actually is today:

- Documents `Stream` (proto/JSON, single + batched), `ProtoSchema` (dynamic
  descriptor + JSON encoder from Unity Catalog metadata), `ArrowStream` (Arrow
  Flight, Beta), `HeadersProvider`, and the configuration/error types.
- No `*_nowait` variants or `test_harness.hpp` — those are not in the tree.
- No ack-callback API (`AckCallback` / `StreamOptions::ack_callback`) — that is
  not on `main` yet, so it is intentionally omitted here and will be documented
  alongside the branch that introduces it.
- Examples are referenced as forthcoming rather than linked, since `cpp/examples/`
  is not part of this branch.

## Verification

- `make test` passes on a clean checkout (all 4 tests, including the Arrow
  config tests).
- Every public API name, build/release claim, and internal detail in the docs
  was cross-checked against the headers, `CMakeLists.txt`, and the FFI header.

## Notes

Docs-only; no code or behavior changes.


cpp/CLAUDE.md
```diff
@@ -0,0 +1,177 @@
...
+`Stream` variant expresses optional fields (e.g. `stream_paused_max_wait_time_ms`)
+with the C struct's `has_*` presence flags; the Arrow variant uses a `-1`
+sentinel for "wait the full server-specified pause duration". `config_defaults_test`
+and `arrow_config_defaults_test` fail the build if the hand-kept scalar defaults
+in `StreamOptions` / `ArrowStreamOptions` drift from the FFI defaults.
+
+## Thread safety
+
+- A `Stream` / `ArrowStream` is **not** safe for concurrent use — serialize
+  access externally (same contract as Java and the Rust core).
```
teodordelibasic-db: Let's not mention Java here.

cpp/README.md
```diff
@@ -0,0 +1,303 @@
...
+  authorized by doing the OAuth client-credentials exchange first (which is what
+  the schema-generation tooling does), needing no PAT.
+
+## Generating a `.proto` for the static path
+
+You do not hand-write the `.proto` for a real table — the monorepo ships a
+generator that emits it (plus a binary descriptor) from the live Unity Catalog
+schema, authenticated with the OAuth client credentials:
+
+```bash
```
teodordelibasic-db, line 322: The generator command uses `--table "catalog.schema.table"` with `--output-dir ./out` and no `--output`. With no `--output`, `generate_files` defaults the filename to the cleaned last component of the table name, so it writes `./out/table.proto`, not `catalog.schema.table.proto` (see [`rust/tools/generate_files/src/main.rs:77`](https://github.com/databricks/zerobus-sdk/blob/98862fa7b1d46349ec90facb3fa9c18d6c05bb1a/rust/tools/generate_files/src/main.rs#L77)). The next paragraph then tells the reader to compile `catalog.schema.table.proto`, which will not exist if they copy the documented command.

We can add `--output catalog.schema.table.proto` to the command.

cpp/README.md
```diff
@@ -0,0 +1,303 @@
...
+- CMake ≥ 3.16
+- A Rust toolchain (only when building the FFI library from source, which is the
+  default)
+
+## Building
+
+From `cpp/`:
+
+```bash
+make build        # configure + build the SDK, tests, and examples
```
teodordelibasic-db: nit: Comment states "and examples", however we haven't added them yet.

cpp/CLAUDE.md
```diff
@@ -0,0 +1,177 @@
...
+│   ├── config.hpp        # StreamOptions / ArrowStreamOptions / enums
+│   ├── record.hpp        # UnackedRecord
+│   ├── error.hpp         # ZerobusException
+│   └── version.hpp       # ZEROBUS_CPP_VERSION
+├── src/                  # Implementation (the only place that includes zerobus.h)
+│   ├── sdk.cpp / stream.cpp / arrow_stream.cpp / proto_schema.cpp
+│   ├── headers_callback.cpp  # extern "C" trampoline for HeadersProvider
+│   └── detail/           # Internal: ffi_util, config_convert, headers_callback
+├── tests/                # Dependency-free unit tests (plain executables)
+├── examples/             # Runnable usage examples
```
teodordelibasic-db: nit: Same as in `cpp/README.md`, `examples` have not been added yet.

CLAUDE.md
teodordelibasic-db: Let's change from `four` to `five.`

cpp/README.md
```diff
@@ -0,0 +1,303 @@
...
+  carries a message and an `is_retryable()` flag.
+- **Proto and JSON** ingestion, single and batched.
+- **Dynamic protobuf** — build a descriptor and encode records straight from
+  Unity Catalog table metadata, with no `.proto` file or `protoc` required.
+- **Arrow Flight** ingestion (Beta) — stream Arrow record batches with optional
+  LZ4/ZSTD compression.
+
+> Status: `0.1.0` — initial development. The API may change before `1.0.0`.
+
+## Requirements
```
teodordelibasic-db, line 22: Other SDKs have a prerequisite pointer to the root `README.md` regarding workspace, Delta table, service principal, etc. Let's add it here as well maybe.

cpp/README.md
```diff
@@ -0,0 +1,303 @@
...
+**Never wait for an acknowledgement after every ingest.** Calling
+`wait_for_offset()` (or `flush()`) inside the ingest loop forces a full server
+round-trip per record and collapses throughput. Instead, loop and flush once at
+the end (or periodically for a continuous stream):
+
+```cpp
+for (const auto& record : records) {
+  stream.ingest_json_record(record);   // queue only — do NOT wait here
+}
+stream.flush();                        // wait once for all pending acks
```
teodordelibasic-db, line 119: Let's also mention `wait_for_offset` here is identical, last offset ack means all prior requests are acked.

cpp/CLAUDE.md
```diff
@@ -0,0 +1,177 @@
...
+  `ProtoSchema::encode_json` frees its buffer with `zerobus_free_proto_bytes`;
+  `descriptor_bytes()` copies the borrowed bytes.
+
+### Headers provider callback
+
+When using a custom `HeadersProvider`:
+- `detail::zerobus_cpp_headers_trampoline` is an `extern "C"` function passed to
+  `zerobus_sdk_create_stream_with_headers_provider`. `user_data` is the raw
+  `HeadersProvider*`.
+- The `Stream` / `ArrowStream` keeps a `std::shared_ptr<HeadersProvider>` alive
```
teodordelibasic-db, line 87: The public header at [`cpp/include/zerobus/headers_provider.hpp:22`](https://github.com/databricks/zerobus-sdk/blob/98862fa7b1d46349ec90facb3fa9c18d6c05bb1a/cpp/include/zerobus/headers_provider.hpp#L22) says the `shared_ptr` is necessary but not always sufficient: a `get_headers()` call still running when `close()` times out (about 1s) can be invoked on a freed provider.

Let's be consistent and mirror the header's caveat here, if that one is correct?

teodordelibasic-db (commented):
Mostly nits.

cpp/CLAUDE.md
zlata-stefanovic-db: Deleted Java

cpp/CLAUDE.md
zlata-stefanovic-db: Changed the comment to be consistent

cpp/README.md
zlata-stefanovic-db: Added here

cpp/CLAUDE.md
zlata-stefanovic-db: Referenced them as "forthcoming" and "when they land" etc.

cpp/README.md
zlata-stefanovic-db: Done

teodordelibasic-db (approved):
LGTM

PR #493, opened 2026-07-13
[Rust SDK] Fix generating VARIANT types in nested schemas

brankogrb-db:

VARIANT columns nested inside a STRUCT, ARRAY, or MAP failed when building a descriptor from a Unity Catalog table schema:
- Rust core: unknown primitive type 'variant'
- Java GenerateProto: Unsupported column type (and Unsupported array/map element type when nested)

Top-level VARIANT was already handled; only the nested code paths were missing the mapping.

TODO:
Improve Java's `GenerateProto` to handle structs

## How is this tested?

- `nested_variant_maps_to_string` — VARIANT nested in STRUCT/ARRAY/MAP → proto string
- `nested_variant_maps_to_large_utf8` — Arrow Flight path → LargeUtf8

Also tested ingesting data

### DIFF
--- a/rust/NEXT_CHANGELOG.md
+++ b/rust/NEXT_CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bug Fixes
 
+- Fixed `VARIANT` columns nested inside a `STRUCT`, `ARRAY`, or `MAP` failing with `unknown primitive type 'variant'` when building a descriptor from a Unity Catalog schema. Nested `VARIANT` now maps to `string` (unshredded JSON-encoded text) at any depth, matching the top-level behavior. Applies to both the protobuf and Arrow Flight schema paths (`descriptor_from_uc_columns` / `arrow_schema_from_uc_columns`), and so also fixes the Go and C++ SDKs, which build descriptors through the C FFI.
+
 ### Documentation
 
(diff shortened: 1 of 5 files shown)


rust/NEXT_CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### Bug Fixes
 
+- Fixed `VARIANT` columns nested inside a `STRUCT`, `ARRAY`, or `MAP` failing with `unknown primitive type 'variant'` when building a descriptor from a Unity Catalog schema. Nested `VARIANT` now maps to `string` (unshredded JSON-encoded text) at any depth, matching the top-level behavior. Applies to both the protobuf and Arrow Flight schema paths (`descriptor_from_uc_columns` / `arrow_schema_from_uc_columns`), and so also fixes the Go and C++ SDKs, which build descriptors through the C FFI.
```
teodordelibasic-db: I'd say the `and so also fixes the Go and C++ SDKs, which build descriptors through the C FFI` is slightly confusing. Separating out these two SDKs is a bit weird, Go SDK doesn't expose this path, C++ does but is not finished/released. Eventually we will want to expose this API in all SDKs, so I think there is no need to mention any other SDK other than Rust. Might just be my subjective feeling, so think about it, not a blocker.

teodordelibasic-db (dismissed):
LGTM, nit in `rust/NEXT_CHANGELOG.md` to think about.

rust/NEXT_CHANGELOG.md
brankogrb-db: removed `and so also fixes the Go and C++ SDKs, which build descriptors through the C FFI.` line

teodordelibasic-db (approved):

PR #494, opened 2026-07-13
[PureGo] Add custom auth-header hook to transport

zlata-stefanovic-db:
## Summary

First of a 3-PR stack splitting the original purego OAuth branch; targets `main`.
Builds on the transport layer landed in #463, replacing its raw-token field with a
pluggable `HeadersProvider` as the single source of stream credentials.

### What changes
- **`StreamParams.Token` → `StreamParams.HeadersProvider`.** The raw-token field is
  dropped in favor of a provider hook, matching the auth model of the other SDKs.
  When a provider is set, `GetHeaders(ctx, tableName)` supplies the auth (and any
  other) metadata headers; when nil, the stream opens with no auth header.
- **`metadata.go` → `headers.go`.** All header/auth concerns — the interface,
  resolution, validation, normalization, and the auth-rejection predicate — now
  live in one file. `resolveHeaders` is extracted from `Open`.
- **Header validation at the wire boundary.** Keys are checked against gRPC's
  `[0-9a-z-_.]` rule, values reject control/non-ASCII characters, and duplicate
  case-folded keys are rejected — so a malformed header fails loudly in `Open`
  rather than opaquely inside gRPC.
- **Auth normalization stays in transport.** A bare authorization value is
  prefixed with `"Bearer "`; a value already carrying a known scheme
  (`Bearer`/`Basic`/`DPoP`, case-insensitive) is sent verbatim.
- **Table-name stays authoritative.** A provider-supplied table-name header is
  ignored in favor of the stream-open param, and inherited stale metadata for the
  Zerobus-owned keys is replaced (not appended), so gRPC first-value-wins can't
  mis-route or send a stale token.
- **Auth rejection → invalidate.** An `Unauthenticated`/`PermissionDenied` on open
  invalidates the provider so the next open re-derives credentials.
- **Bounded open.** `GetHeaders`, the RPC start, and the handshake all run under
  the open deadline (the caller's `ctx`, defaulted to `defaultHandshakeTimeout`
  when it has none), so a blocking token mint can't hang `Open` indefinitely.
- **Handshake Send `io.EOF` fix.** Per the gRPC `ClientStream` contract, a `Send`
  that aborts server-side returns `io.EOF` with the real status on `Recv`. The
  handshake now falls through to `Recv` on that `io.EOF` instead of returning it,
  so an auth rejection racing onto the send path reaches `isAuthRejection` and
  triggers invalidation rather than surfacing as an opaque EOF.

Transport declares its own local `HeadersProvider` interface and consumes it
duck-typed — no dependency on the `auth` package (which lands in #495).

## Stack
1. **this PR** — transport header hook
2. #495 — auth `HeadersProvider` + token cache
3. #496 — UC OAuth provider

## Test plan
- `go build ./...`, `go vet ./...`, and `gofmt -l` are clean.
- `go test -race ./internal/transport/` passes (24 test functions), covering:
  auth-scheme normalization, invalid/duplicate header rejection, table-name
  authority, inherited-metadata replacement, auth-rejection invalidation (and its
  negative case), provider-error propagation, deadline-bounded `GetHeaders`, and
  the handshake `Send`-`io.EOF` fall-through.
- Deadline/cancel tests assert on the returned context error rather than
  wall-clock elapsed time, so they don't flake.


purego/internal/transport/ephemeral_stream.go
```diff
@@ -23,7 +25,17 @@ type StreamParams struct {
 	// Token is the credential sent in the authorization header. A bare token is
 	// prefixed with "Bearer "; a value that already carries a known scheme (e.g.
 	// "Bearer ..." or "Basic ...") is sent verbatim. Empty means no header.
+	// Ignored when HeadersProvider is set.
 	Token string
+	// HeadersProvider supplies custom auth headers, similar to other SDKs.
```
teodordelibasic-db: I would emit the `similar to other SDKs` part probably.

purego/internal/transport/ephemeral_stream.go
```diff
@@ -23,7 +25,29 @@ type StreamParams struct {
 	// Token is the credential sent in the authorization header. A bare token is
 	// prefixed with "Bearer "; a value that already carries a known scheme (e.g.
 	// "Bearer ..." or "Basic ...") is sent verbatim. Empty means no header.
+	// Ignored when HeadersProvider is set.
 	Token string
```
teodordelibasic-db: Is there a reason to keep this at all?

purego/internal/transport/ephemeral_stream.go
```diff
@@ -63,6 +88,23 @@ func (c *Conn) Open(ctx context.Context, p StreamParams) (*Stream, error) {
...
+		headers, err = p.HeadersProvider.GetHeaders(ctx, p.TableName)
+		if err != nil {
+			return nil, fmt.Errorf("transport: open %q: headers provider: %w", p.TableName, err)
+		}
+		for k, v := range headers {
+			if !isUsableAsHeader(v) {
+				return nil, fmt.Errorf("transport: open %q: header %q contains invalid value characters", p.TableName, strings.TrimSpace(k))
+			}
+		}
+	} else if !isUsableAsHeader(p.Token) {
```
teodordelibasic-db: You might have this planned out already, but if HeadersProvider is not set, I think we should ideally default to OAuth currently and the other mandatory header(s).

purego/internal/transport/metadata.go
```diff
@@ -23,21 +24,57 @@ var authSchemes = []string{"bearer", "basic", "dpop"}
 // are replaced (gRPC is first-value-wins, so a duplicate could mis-route or send
 // a stale token); unrelated caller metadata is preserved.
 func withStreamMetadata(ctx context.Context, tableName, token string) context.Context {
```
teodordelibasic-db: Since I think this is now unused, should we remove it?

purego/internal/transport/ephemeral_stream.go
```diff
@@ -63,6 +88,23 @@ func (c *Conn) Open(ctx context.Context, p StreamParams) (*Stream, error) {
 	if p.RecordType == zerobuspb.RecordType_PROTO && len(p.DescriptorProto) == 0 {
 		return nil, fmt.Errorf("transport: open %q: descriptor proto required for PROTO records", p.TableName)
 	}
+	var headers map[string]string
+	if p.HeadersProvider != nil {
+		var err error
+		headers, err = p.HeadersProvider.GetHeaders(ctx, p.TableName)
```
teodordelibasic-db: Should we maybe add a default timeout to this context as well? I mean either construct `openCtx` before this call or something like that.

purego/internal/transport/metadata.go
teodordelibasic-db: Predates this PR but are we sure it is okay for the authorization header to not start with anything other than `"Bearer..."`?

purego/internal/transport/ephemeral_stream.go
zlata-stefanovic-db: The reason was simplifying the tests on the lower level, I'll remove it in this PR for consistency with other SDKs

purego/internal/transport/metadata.go
zlata-stefanovic-db: Yes, removed it along with the token

purego/internal/transport/ephemeral_stream.go
zlata-stefanovic-db: This will land in one of the stacked PRs (#496), but noted.

purego/internal/transport/ephemeral_stream.go
zlata-stefanovic-db: Yes, doing this

purego/internal/transport/raw_stream.go
```diff
@@ -11,12 +12,12 @@ import (
...
 // defaultHandshakeTimeout bounds the open attempt when the caller's context has
 // no deadline, so Open can't hang if the server half-opens the stream.
-const defaultHandshakeTimeout = 30 * time.Second
+const defaultHandshakeTimeout = 15 * time.Second
 
 // defaultDrainTimeout bounds gracefulClose when the caller's context has no
 // deadline, so it can't hang on an unresponsive server. A var so tests can
 // shrink it.
-var defaultDrainTimeout = 30 * time.Second
+var defaultDrainTimeout = 500 * time.Millisecond
```
danilonajkov-db, line 22: this means that we wait only for 0.5s for draining the pending records queue? 30s was probably too much, but I would leave 5s

purego/internal/transport/metadata.go
zlata-stefanovic-db: Good push I've now removed the scheme handling from the transport entirely (dropped the Bearer prefixing and the Bearer/Basic/DPoP detection). The transport sends the authorization value verbatim, exactly as the provider formats it - it makes no assumption about the scheme.

This is what the Rust core and the other SDKs already do: the transport treats authorization as opaque, and formatting "Bearer <token>" is the provider's job (the UC OAuth provider in #496 does exactly that). We still trim, drop the header on an empty value so a stale inherited one can't leak, and reject control/non-ASCII characters.

purego/internal/transport/raw_stream.go
zlata-stefanovic-db: This is not the ack drain timeout, actually the 500ms was done to match the rust core.

More explanation:
In Rust there is a flush() operation which is called before close() and has a flush_timeout_ms (much larger than 500ms - default is 300s), and during this we wait for acks. Afterwards there is a teardown with stream_teardown_drain_timeout_ms (500ms) which is the value we're trying to match here. By the time this runs, the flush has already gotten all the acks so this serves as a bound to let recv task unwind to a clean finish so the socket closes with end_stream cleanly.

In Go, we still haven't implemented the ack tracking so there is no records being drained right now so this might look a little strange. When the flush() lands in Go, it will make sense and the longer ack-wait budget will live there.

zlata-stefanovic-db: Also it's a no-deadline fallback so if the caller/ctx has a deadline it wont default to 500ms

danilonajkov-db (approved):

PR #495, opened 2026-07-13
[PureGo] Add auth provider interfaces and per-table token cache

zlata-stefanovic-db:
## Summary

Second of a 3-PR stack. **Stacked on #494** (base branch `purego-transport-headers`).

Introduces the `auth` package foundations:
- `HeadersProvider` interface — the single auth extension point, matching the Rust SDK (and duck-type compatible with the transport's local `HeadersProvider` from #494).
- `StaticHeadersProvider` — returns a fixed headers map, for tests or externally managed credentials.
- A per-table token cache with proactive refresh and single-flight minting, which bounds a failing token endpoint to one mint per burst rather than N serial attempts.

OAuth-backed providers land in the follow-up PR.

## Stack
1. #494 — transport header hook
2. **this PR** — auth HeadersProvider + token cache
3. #496 — UC OAuth provider

## Test plan
- `go build ./internal/auth/` and `go test ./internal/auth/` pass.

## Note
Rebased on the updated #494, which dropped `StreamParams.Token` in favor of the `HeadersProvider` hook. Auth is expressed solely through `HeadersProvider`; a speculative `TokenProvider` interface (and `StaticTokenProvider`) was dropped as unused, mirroring the Rust SDK where `HeadersProvider` is the only extension point. The header-provider types live in `headers_provider.go`.

### DIFF
--- a/purego/README.md
+++ b/purego/README.md
@@ -21,9 +21,11 @@ Implemented packages:
   the `StreamParams.HeadersProvider`. The authorization value is sent verbatim as
   the provider formats it (e.g. `"Bearer <token>"`); the transport does not add a
   scheme prefix.
+- `internal/auth` — the `HeadersProvider` interface and `StaticHeadersProvider`
+  for fixed-header cases, plus a per-`(clientID, secret, table)` token cache
+  with single-flight minting, proactive refresh, and retryable-error fallback.
 
-Planned layers: OAuth/auth, ingest/ack state management, recovery, and the
-public API.
+Planned layers: ingest/ack state management, recovery, and the public API.
 
 ## Regenerating the protobuf bindings
 
--- a/purego/internal/auth/headers_provider.go
+++ b/purego/internal/auth/headers_provider.go
@@ -1,13 +1,9 @@
-// Package auth provides OAuth 2.0 and custom-header authentication for the
-// Zerobus pure-Go SDK.
+// Package auth provides authentication for the Zerobus pure-Go SDK.
 //
-// Authentication is expressed through [HeadersProvider], which supplies the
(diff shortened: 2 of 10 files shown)


danilonajkov-db:
AI found this:
```
Single-flight can be bypassed by cache pruning in purego/internal/auth/token_cache.go:321-345. slot() returns an entry before its caller acquires entry.mu. A concurrent slot() for another key can prune that uncached/expired entry from entries; a subsequent call for the original key creates a new entry. Both entries can then mint concurrently for the same (clientID, secret, table), violating the promised per-key single-flight bound. Avoid evicting entries that have no cached token, or coordinate eviction with entry acquisition/removal so detached entries cannot mint.
```
Seems like a very low impact problem, as the probability of this happening is low and in the worst case we request token 2 times. But it would be good if we can address it.

danilonajkov-db (dismissed):

zlata-stefanovic-db:
> AI found this:
> 
> ```
> Single-flight can be bypassed by cache pruning in purego/internal/auth/token_cache.go:321-345. slot() returns an entry before its caller acquires entry.mu. A concurrent slot() for another key can prune that uncached/expired entry from entries; a subsequent call for the original key creates a new entry. Both entries can then mint concurrently for the same (clientID, secret, table), violating the promised per-key single-flight bound. Avoid evicting entries that have no cached token, or coordinate eviction with entry acquisition/removal so detached entries cannot mint.
> ```
> 
> Seems like a very low impact problem, as the probability of this happening is low and in the worst case we request token 2 times. But it would be good if we can address it.

Fixed by adding a minted flag to each cache slot.

danilonajkov-db (approved):

PR #496, opened 2026-07-13
[PureGo] Add UC OAuth token provider with per-table caching

zlata-stefanovic-db:
## Summary

Third of a 3-PR stack. **Stacked on #495** (base branch \`purego-auth-cache\`).

Adds the Unity Catalog OAuth 2.0 provider:
- \`OAuthTokenProvider\` — client-credentials flow minting downscoped per-table tokens, with proactive refresh, endpoint/table validation, structured \`slog\` observability, and an injectable \`*http.Client\`. A concrete type, not an interface — the \`HeadersProvider\` interface is the extension point.
- \`OAuthHeadersProvider\` — wraps \`OAuthTokenProvider\` and emits the Zerobus metadata headers (\`authorization: Bearer <token>\` and \`x-databricks-zerobus-table-name\`), satisfying the transport headers-provider hook from #494.
- \`SharedTokenCache\` / \`NewSharedTokenCache\` — allows multiple providers to share one token cache so tokens are reused across streams for the same credentials.
- \`FetchToken\` — bypasses the cache for a guaranteed-fresh mint; useful for diagnostics and forced refreshes.
- Treats all 4xx token responses (including 429/408) as non-retryable and anchors cached token TTL at response receipt, matching the Rust SDK.
- Requires \`https\` for non-loopback \`ucEndpoint\` so client credentials are never sent over plaintext.

## Stack
1. #494 — transport header hook
2. #495 — auth \`HeadersProvider\` + token cache
3. **this PR** — UC OAuth provider

## Test plan
- \`go build ./...\` and \`go test ./internal/...\` pass.

## Note
\`OAuthHeadersProvider\` lives in \`headers_provider.go\` alongside \`HeadersProvider\`/\`StaticHeadersProvider\` (added in #495), since it depends on \`oauth.go\` in this PR. The \`deriveWorkspaceIDFromEndpoint\` helper extracts the workspace ID from the Zerobus endpoint host prefix; the \`zerobusHost\` return value was dropped in a later revision as it was unused.

### DIFF
--- a/purego/internal/auth/oauth.go
+++ b/purego/internal/auth/oauth.go
@@ -11,6 +11,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
 	"time"
 	"unicode"
@@ -66,9 +67,10 @@ func WithHTTPClient(c *http.Client) OAuthOption {
 // [NewSharedTokenCache]. It exposes no methods and is safe for concurrent use.
 type SharedTokenCache = tokenCache
 
-// WithSharedTokenCache installs a shared token cache. Use this when multiple
-// OAuthTokenProviders are created for the same credentials but different tables
-// so they share one cache map rather than each allocating their own.
+// WithSharedTokenCache installs a shared token cache. A single provider already
+// caches all of its tables; use this only when multiple OAuthTokenProvider
+// instances should pool their tokens in one cache rather than each allocating
+// its own (a single provider serves many tables on its own).
 //
 // Obtain a cache with [NewSharedTokenCache]. A nil cache is ignored so the
 // provider's own default cache is preserved.
@@ -246,9 +248,9 @@ func (p *OAuthTokenProvider) mint(ctx context.Context, tableName string, reason 
 	return fetched, err
 }
 
-// Invalidate drops the cached token for this provider's table so the next
-// Token call re-mints from Unity Catalog. Call this when the server rejects
-// the token with an authentication error.
+// Invalidate drops the cached token for the given table so the next Token call
+// re-mints from Unity Catalog. Call this when the server rejects the token with
+// an authentication error.
 //
 // The table name is not re-validated here: it was validated at stream open, and
 // invalidate is a no-op for a key with no cached entry, so an unrecognized name
@@ -303,8 +305,10 @@ func (p *OAuthTokenProvider) fetchToken(ctx context.Context, tableName string) (
 
 	var body struct {
 		AccessToken string `json:"access_token"`
-		// RFC 6749 §4.2.2 defines expires_in as an integer number of seconds.
-		ExpiresIn int64 `json:"expires_in"`
+		// RFC 6749 §4.2.2 defines expires_in as an integer number of seconds. It
+		// is captured raw and parsed tolerantly (see parseExpiresIn) so a server
+		// that sends it as a string or float doesn't fail the whole mint.
+		ExpiresIn json.RawMessage `json:"expires_in"`
 	}
 	if err := json.NewDecoder(resp.Body).Decode(&body); err != nil {
 		return fetchedToken{}, &TokenError{
@@ -321,10 +325,37 @@ func (p *OAuthTokenProvider) fetchToken(ctx context.Context, tableName string) (
 	}
 
 	ft := fetchedToken{token: body.AccessToken}
-	if d, ok := secondsToDuration(body.ExpiresIn); ok {
+	if d, ok := secondsToDuration(parseExpiresIn(body.ExpiresIn)); ok {
 		ft.expiresIn = &d
 	}
 	return ft, nil
+}
+
+// parseExpiresIn tolerantly reads an OAuth expires_in value. RFC 6749 specifies
+// an integer number of seconds, but some servers send a JSON string or float;
+// all are accepted. A missing, null, or unparseable value yields 0, which
+// secondsToDuration treats as "no usable TTL" so the token is still returned
+// (just not cached) rather than failing the mint on a cosmetic type mismatch.
+func parseExpiresIn(raw json.RawMessage) int64 {
+	s := strings.TrimSpace(string(raw))
+	if s == "" || s == "null" {
+		return 0
+	}
+	// Unwrap a quoted JSON string ("3600") to its inner value before parsing.
+	if unquoted, err := strconv.Unquote(s); err == nil {
+		s = unquoted
+	}
+	if n, err := strconv.ParseInt(s, 10, 64); err == nil {
+		return n
+	}
+	if f, err := strconv.ParseFloat(s, 64); err == nil && f >= 0 {
+		// Truncate toward zero; sub-second precision is irrelevant for a TTL.
+		if f > float64(math.MaxInt64) {
+			return math.MaxInt64
+		}
+		return int64(f)
+	}
+	return 0
 }
 
 // maxExpiresInSeconds bounds a server-reported expires_in so the seconds→
--- a/purego/internal/auth/oauth.go
+++ b/purego/internal/auth/oauth.go
@@ -200,9 +200,9 @@ func (p *OAuthTokenProvider) tokenAudience() string {
 }
 
 // FetchToken mints a token directly from Unity Catalog, bypassing the cache. It
-// neither reads nor writes the cache, so callers get a guaranteed-fresh token;
-// most callers should use [OAuthTokenProvider.Token] instead so tokens are
-// reused across streams.
+// neither reads nor writes the cache and does not invalidate any existing cached
+// entry, so callers get a guaranteed-fresh token; most callers should use
+// [OAuthTokenProvider.Token] instead so tokens are reused across streams.
 func (p *OAuthTokenProvider) FetchToken(ctx context.Context, tableName string) (string, error) {
 	tableName = strings.TrimSpace(tableName)
 	if err := validateTableName(tableName); err != nil {
@@ -291,11 +291,14 @@ func (p *OAuthTokenProvider) fetchToken(ctx context.Context, tableName string) (
 	if err != nil {
 		return fetchedToken{}, &TokenError{msg: fmt.Sprintf("token request: %v", err), retryable: isRetryableTransportError(ctx, err), cause: err}
 	}
-	// Drain any unread bytes before closing so the keep-alive connection can be
+	// Drain unread bytes before closing so the keep-alive connection can be
 	// reused: json.Decode stops at the end of the JSON value and classifyHTTPError
-	// caps its read, either of which can leave a trailing tail on the body.
+	// caps its read, either of which can leave a trailing tail on the body. The
+	// drain is itself capped so a server appending an unbounded tail after a valid
+	// response can't force an arbitrarily long read; past the cap the connection
+	// is simply closed instead of reused.
 	defer func() {
-		_, _ = io.Copy(io.Discard, resp.Body)
+		_, _ = io.Copy(io.Discard, io.LimitReader(resp.Body, 4096))
 		resp.Body.Close()
 	}()
 
@@ -501,7 +504,7 @@ func isRetryableTransportError(ctx context.Context, err error) bool {
 // authorization header value. gRPC metadata rejects values with ASCII control
 // characters (bytes 0–31 and DEL).
 //
-// Note: transport.isUsableAsHeader applies the same character check but
+// Note: transport.isUsableAsHeaderValue applies the same character check but
 // intentionally treats the empty string as usable ("no header"); this copy
 // returns false for empty because a minted token must be non-empty. Keep the
 // character logic in sync if either is changed.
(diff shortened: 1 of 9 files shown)


purego/internal/auth/oauth.go
```diff
@@ -0,0 +1,551 @@
...
+		}
+	}
+}
+
+// SharedTokenCache is an OAuth token cache shared across multiple
+// [OAuthTokenProvider] instances via [WithSharedTokenCache]. Obtain one with
+// [NewSharedTokenCache]. It exposes no methods and is safe for concurrent use.
+type SharedTokenCache = tokenCache
+
+// WithSharedTokenCache installs a shared token cache. Use this when multiple
```
teodordelibasic-db: Not entirely sure about this interaction regarding different workspaces and same table, but we should check.

Since `tokenKey` is `(clientID, secretDigest, tableName)`, but the minted token is bound to the workspace audience `api://databricks/workspaces/{workspaceID}/zerobusDirectWriteApi` (see [`oauth.go:269`](https://github.com/databricks/zerobus-sdk/blob/4edf0707e4360d2fdac851ca2cbf9a6fd028236f/purego/internal/auth/oauth.go#L269)), two providers with the same credentials and table but different workspace collide on one key. The second is served the first's token, it will probably be rejected on the audience mismatch, and the bad entry will keep being served until it expires.

purego/internal/auth/oauth.go
```diff
@@ -0,0 +1,551 @@
...
+	return &TokenError{msg: msg, retryable: isRetryableStatus(resp.StatusCode)}
+}
+
+func isHTTPSuccess(code int) bool { return code >= 200 && code < 300 }
+
+// isRetryableTransportError reports whether a transport-level error from the
+// token request is transient and safe to retry. Network-level timeouts and
+// connection failures qualify; any other transport error is treated as
+// non-retryable so a genuine failure isn't masked by a stale cached token.
+func isRetryableTransportError(err error) bool {
```
teodordelibasic-db: Here we treat anything wrapping `context.DeadlineExceeded` as non-retryable, and by design that includes `http.Client.Timeout`. The cache only serves a still-valid cached token on a failed refresh when the error is retryable ([`token_cache.go:258`](https://github.com/databricks/zerobus-sdk/blob/4edf0707e4360d2fdac851ca2cbf9a6fd028236f/purego/internal/auth/token_cache.go#L258)). So a proactive refresh that trips the timeout on a briefly slow UC endpoint fails the open even though a valid token was in hand. This can happen with no caller deadline, since `transport.Open` adds a 15s header budget of its own. Rust treats request timeouts as retryable and serves the cached token here.

The caller-cancellation case is fine to keep non-retryable. Could we separate that from a mint attempt timing out, and on a proactive refresh where the cached token is still valid serve it regardless of the failure class? A test that seeds a still-valid-but-refresh-due token, forces a client timeout on refresh, and asserts the cached token is returned would cover it.

purego/internal/auth/token_cache.go
teodordelibasic-db: A caller parked on another goroutine's in-flight mint re-attempts when the leader failed with a context error and its own context is still live. That was meant for genuine leader cancellation, but this PR's mint also wraps `context.DeadlineExceeded` when `http.Client.Timeout` fires, which is not a cancellation. So against a slow UC, a burst of opens for one table turns one mint into many: each waiter re-attempts, becomes the next leader, and mints again, which is the opposite of what single-flight is for.

Could we restrict the re-attempt to genuine caller-context cancellation and not to a transport or client timeout from the leader's mint?

purego/internal/auth/oauth.go
```diff
@@ -0,0 +1,551 @@
...
+func (e *TokenError) Unwrap() error { return e.cause }
+
+// isRetryableStatus reports whether an HTTP status is a transient failure worth
+// suppressing when a cached token can be served. Only 5xx responses qualify;
+// all 4xx (including 429 and 408) are non-retryable.
+func isRetryableStatus(code int) bool {
+	return code >= 500
+}
+
+func classifyHTTPError(resp *http.Response) error {
```
teodordelibasic-db, line 483: Could we log the status code plus the structured OAuth `error` / `error_description` fields, and keep the raw body out of the logged message even if we retain it on the error type for diagnostics?

purego/internal/auth/oauth.go
zlata-stefanovic-db: Included the workspace audience in the token cache key.

purego/internal/auth/oauth.go
zlata-stefanovic-db: Fixed as suggested. Made only a cancel/deadline from the caller's own context stay non-retryable.

purego/internal/auth/token_cache.go
zlata-stefanovic-db: A retryable leader-mint timeout is now shared by waiters.

purego/internal/auth/oauth.go
zlata-stefanovic-db: Kept only those fields, retained the raw body on TokenError.ResponseBody for diagnostics.

purego/internal/auth/oauth.go
```diff
@@ -0,0 +1,623 @@
...
+	}
+
+	var body struct {
+		AccessToken string `json:"access_token"`
+		// RFC 6749 §4.2.2 defines expires_in as an integer number of seconds. It
+		// is captured raw and parsed tolerantly (see parseExpiresIn) so a server
+		// that sends it as a string or float doesn't fail the whole mint.
+		ExpiresIn json.RawMessage `json:"expires_in"`
+	}
+	if err := json.NewDecoder(resp.Body).Decode(&body); err != nil {
```
teodordelibasic-db: Here we read with no size cap, unlike the error path (L453) and the drain (L301), which are both capped at 4096. Should we introduce cap here as well?

purego/internal/auth/token_cache.go
teodordelibasic-db: The `http.Client.Timeout`-with-live-caller-ctx case is now fixed, but I think the path I originally flagged still fails.

When the caller has no deadline, `transport.Open` runs the refresh mint under its own 15s header budget (`ephemeral_stream.go:70-77`), and the client timeout is 30s, so the budget fires first. Since the mint runs under that budget ctx, `ctx.Err() != nil` when it fires, so it's classified as caller cancellation which is non-retryable, so the still-valid cached token isn't served and open fails. The heuristic ("is the caller ctx done?") can't tell the 15s wrapper budget apart from a real caller giving up.

purego/internal/auth/oauth.go
zlata-stefanovic-db: Addressed now, capped at 4096

purego/internal/auth/oauth.go
```diff
@@ -0,0 +1,669 @@
...
+
+	tokenURL := p.ucEndpoint + "/oidc/v1/token"
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, tokenURL, strings.NewReader(form.Encode()))
+	if err != nil {
+		return fetchedToken{}, &TokenError{msg: fmt.Sprintf("build token request: %v", err), retryable: false, cause: err}
+	}
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	req.SetBasicAuth(p.clientID, p.clientSecret)
+
+	resp, err := p.client.Do(req)
```
teodordelibasic-db, line 318: This request sends the client ID and secret to the server, and it does so over an encrypted `https` connection so nobody on the network can read them. The constructor enforces that the endpoint is `https` for exactly this reason.

The gap is that we send this request with the default HTTP client, which automatically follows redirects. A redirect is the server replying "don't use this address, go ask this other one instead," and the client then quietly re-sends the same request, credentials included, to the new address. When Go decides whether to attach the credentials to that second request, it only checks that the new address has the same host name. It does not check whether the new address is still `https`. So if the server redirects from `https://host/...` to `http://host/...` (unencrypted), Go re-sends the ID and secret in the clear, and anyone watching that connection can read them. The `https` guarantee from the constructor is lost the moment a redirect is followed.

A token endpoint never has a legitimate reason to redirect, so the safest fix is to not follow redirects on this request at all. Copying the client first means we do not change a client the caller may have passed in:

```go
client := *p.client
client.CheckRedirect = func(*http.Request, []*http.Request) error {
        return http.ErrUseLastResponse
}
resp, err := client.Do(req)
```

With `ErrUseLastResponse` a redirect is returned as an ordinary non-success response and the mint fails through the existing error path instead of quietly following the redirect. A test with a same-host `https`-to-`http` redirect and a cross-host redirect would lock this in.

For context, the Rust SDK is not exposed to this even though it also follows redirects with the credentials attached. Rust's HTTP library drops the credentials on this kind of redirect, while Go's keeps them, so this is a case where PureGo behaves less safely than Rust rather than matching it.

purego/internal/auth/token_cache.go
```diff
@@ -269,17 +300,27 @@ func (c *tokenCache) tryGetOrFetch(
...
-	} else {
-		keepExisting := entry.cached != nil && !entry.cached.isExpired()
-		if !keepExisting {
-			entry.cached = nil
-		}
+	case alreadyExpired && keepExisting:
+		// Dead mint but a still-valid token is cached: serve the cached one so
+		// the caller doesn't get a token we already know is expired.
+		token = entry.cached.value
+	case !keepExisting:
```
teodordelibasic-db, line 320: When a freshly minted token is already past its receipt-anchored TTL and there is no still-valid cached entry to fall back on, this falls into the `!keepExisting` case, clears `entry.cached`, and then returns the expired `token` with a nil error. `getOrFetch` documents that it returns a valid token, so the caller gets a nil error but an unusable token, and the failure only surfaces later as a server-side auth rejection rather than here where the cause is known.

Could we return an error for the cold-miss expired-on-arrival case, while keeping the existing behavior that serves a still-valid cached token when an early refresh produces an expired replacement?

```go
case alreadyExpired && keepExisting:
        token = entry.cached.value
case alreadyExpired: // expired on arrival, nothing valid to serve
        entry.cached = nil
        err := &TokenError{msg: "minted token already expired on arrival", retryable: true}
        flight.token, flight.err = "", err
        close(flight.done)
        entry.mu.Unlock()
        return "", err, false
case !keepExisting:
        entry.cached = nil
```

This is a narrow branch in practice: it needs a very short `expires_in` plus slow post-receipt work to consume the whole TTL, so real UC tokens rarely hit it. `TestTokenCacheDoesNotCacheAlreadyExpiredToken` currently asserts the token is returned, so it would need to assert the error instead.

purego/internal/auth/token_cache.go
```diff
@@ -288,16 +329,32 @@ func (c *tokenCache) tryGetOrFetch(
...
+// from the caller stopping, not from the request or an SDK-internal budget.
+// Only such failures are non-retryable; a client/transport timeout or a
+// WithTimeoutCause budget stays retryable so a cached token can be served.
+// When ctx is done, context.Cause tells the caller's own cancel/deadline (plain
+// context.Canceled/DeadlineExceeded) apart from an intermediate budget cause.
+func isCallerCancellation(ctx context.Context, err error) bool {
+	if !isContextError(err) || ctx.Err() == nil {
+		return false
+	}
+	cause := context.Cause(ctx)
```
teodordelibasic-db, line 350: `isCallerCancellation` needs to tell a real caller cancellation (which we must honor by failing) apart from our own internal open-budget timeout (which is transient, so we can fall back to a still-valid cached token). It decides by matching `context.Cause(ctx)` against Go's two standard values. But a caller who cancels via `context.WithCancelCause` has a custom cause, so we do not recognize their cancellation, treat it as retryable, and can serve a cached token instead of stopping as they asked.

The effect is mild: the caller still gets a valid token, not a hang or a wrong error, and nothing inside the SDK cancels this way, so it is only reachable if an external caller uses `WithCancelCause`. Still, cancelling with a reason should behave the same as cancelling without one.

The logic is currently "recognize the caller, treat everything else as our budget"; it is more robust flipped: recognize only our own budget, and treat everything else as a caller cancellation.

```go
func isCallerCancellation(ctx context.Context, err error) bool {
        if !isContextError(err) || ctx.Err() == nil {
                return false
        }
        // Our own open budget is the one non-caller cause; anything else is the caller.
        return context.Cause(ctx) != errHeadersBudgetExceeded
}
```

The snag is that `auth` deliberately does not import `transport`, where `errHeadersBudgetExceeded` lives, so this needs that sentinel shared across the two packages rather than a one-line edit.

teodordelibasic-db (commented):
Let's merge the PR, we can address these comments as follow up. First one seems serious, other two are edge cases.

teodordelibasic-db (approved):

purego/internal/auth/oauth.go
zlata-stefanovic-db: Created  #563

purego/internal/auth/token_cache.go
zlata-stefanovic-db: Created #564

purego/internal/auth/token_cache.go
zlata-stefanovic-db: Created #565

PR #499, opened 2026-07-13
[C++] Expand test coverage: unit, concurrency, and integration tests

zlata-stefanovic-db:
## Summary

Expands the C++ SDK test suite to cover the API surface the current tests left uncovered, matching the parity set the other SDKs test (cf. Java's unit tests) and the suite intended in the closed #421. Also adds concurrency and live-integration coverage, plus a ThreadSanitizer CI job. All hermetic tests are plain, dependency-free, network-free executables registered with CTest, in the same style as the existing tests.

### Unit tests (API surface)

- **error_test** — `ZerobusException`: message, `is_retryable()`, catchable as `std::exception`
- **record_test** — `UnackedRecord`: byte/flag preservation, embedded NUL, empty payload
- **version_test** — `version()` pinned to `ZEROBUS_CPP_VERSION`
- **proto_schema_test** — `ProtoSchema`: UC-JSON round trip, invalid JSON, missing required column, move semantics (moved-from null-handle contract)
- **headers_callback_test** — headers-provider FFI trampoline: marshalling, empty / embedded-NUL / throwing / null-`user_data` guards; results freed via `zerobus_free_headers`
- **sdk_test** — `Sdk` / `SdkBuilder`: offline build (no network I/O), move, and `create_stream` / `create_arrow_stream` argument validation

### Concurrency

- **concurrency_test** — exercises the documented "concurrent readers on a shared `ProtoSchema`" contract (`descriptor_bytes` / `encode_json` are `const`) across many threads. Hermetic, so it runs in CI. Aimed at ThreadSanitizer to catch data races the AddressSanitizer job cannot. No concurrent-`Stream` case (a `Stream` is not safe for concurrent use, so that would test misuse).
- Adds a **ThreadSanitizer CI job** (`make test SANITIZE=thread`) to `ci-cpp.yml`.

### Live integration

- **integration_test** — env-var-gated `create_stream -> ingest -> flush -> close` path against a real endpoint, mirroring the Java/TypeScript integration suites. Uses the loop-then-`flush()` pattern (never waits per record). **Skips (exits 0) when credentials are unset**, so `make test` and CI stay hermetic. Requires `ZEROBUS_SERVER_ENDPOINT`, `DATABRICKS_WORKSPACE_URL`, `ZEROBUS_TABLE_NAME`, `DATABRICKS_CLIENT_ID`, `DATABRICKS_CLIENT_SECRET`; optional `ZEROBUS_TEST_RECORD_JSON`.

## Testing

- `make test` — full suite passes (13 tests; `integration_test` skips without credentials)
- `make test SANITIZE=address` — passes under AddressSanitizer (validates the FFI alloc/free pairing in the headers/proto-schema tests)
- `make test SANITIZE=thread` — passes under ThreadSanitizer (validates the shared-`ProtoSchema` concurrency contract)
- `make fmt-check` — clean

## Merge order

**Stacked on #484** (`cpp-ack-callback`). Base is `cpp-ack-callback`; retarget to `main` once #484 merges. Draft until then.


.github/workflows/ci-cpp.yml
```diff
@@ -77,3 +77,32 @@ jobs:
...
+
+      # Build + run the suite under ThreadSanitizer. This is aimed at the
+      # concurrency contracts the wrapper exposes (e.g. concurrent readers on a
+      # shared ProtoSchema); it catches data races the ASan job cannot. Debug
+      # build for readable stack traces.
+      - name: Build + test with ThreadSanitizer
+        env:
+          # Fail the job on the first race rather than merely printing it.
+          TSAN_OPTIONS: halt_on_error=1
+        run: make test SANITIZE=thread BUILD_TYPE=Debug
```
irinatomic-db, line 109: This TSan job can't detect a race in the contract it's meant to validate. The work the threads in `concurrency_test.cpp` do - `descriptor_bytes()` / `encode_json()` - runs inside the Rust FFI static library, which is built with a plain `cargo build --release`. The `-fsanitize=thread` flag is only applied to the C++ compile/link; nothing passes sanitizer `RUSTFLAGS` / `-Zsanitizer=thread` to cargo, so the Rust core is uninstrumented and TSan cannot see memory accesses inside it. On the C++ side each thread only writes into its own thread-local `std::vector`, so there's no shared mutable state for TSan to watch either. (The ProtoSchema is also immutable after construction with read-only borrows, so no race exists to provoke in the first place.)

Two options: 
(a) build the FFI under TSan too (`RUSTFLAGS="-Zsanitizer=thread"` on a nightly toolchain) so the contract is genuinely instrumented
(b) keep it but reframe the test/job comment as a concurrent-liveness/consistency smoke test rather than a race detector.

cpp/tests/integration_test.cpp
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  std::string record_json = env("ZEROBUS_TEST_RECORD_JSON");
+  if (record_json.empty()) {
+    // Fallback record for the common {id, payload} shape. It almost certainly
+    // won't match an arbitrary table's schema, in which case the run fails at
+    // flush() (ingest only queues) with a schema-mismatch error from the
+    // server. Set ZEROBUS_TEST_RECORD_JSON to a record matching
+    // ZEROBUS_TABLE_NAME.
+    record_json = R"({"id": 1, "payload": "zerobus-cpp-integration"})";
```
irinatomic-db: Nit: Let's make the integration test skip (return 0) unless all args including `ZEROBUS_TEST_RECORD_JSON` are given. Right now if the 5 creds are set but this is unset, the default record won't match the table, flush() throws, and the test goes red - a half-configured env shouldn't fail the suite.

cpp/tests/version_test.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+void fail(const char* msg) {
+  std::fprintf(stderr, "FAIL: %s\n", msg);
+  ++g_failures;
+}
+
+}  // namespace
+
+int main() {
+  if (std::strcmp(zerobus::version(), ZEROBUS_CPP_VERSION) != 0) {
```
irinatomic-db, line 22: Nit: maybe assert on `project(... VERSION ...)` instead

.github/workflows/ci-cpp.yml
zlata-stefanovic-db: Will do (b) now and leave (a) as a follow up - doing it properly is probably worth its own PR

irinatomic-db (approved):

Issue #334, opened 2026-06-02
[Java] Native library not found in classpath: /native/linux-x86_64/libzerobus_jni.so


bulv1ne:
### SDK

Java

### SDK Version

1.2.0

### Operating System

Linux

### Description

I upgraded zerobus sdk from 1.0.1 to 1.2.0 and when trying to initialize the ZerobusSdk I get the following error:

```
Exception in thread "main" java.lang.UnsatisfiedLinkError: Native library failed to load: Native library not found in classpath: /native/linux-x86_64/libzerobus_jni.so. Platform: linux-x86_64. Make sure the correct native JAR is on the classpath.
	at com.databricks.zerobus.NativeLoader.ensureLoaded(NativeLoader.java:63)
	at com.databricks.zerobus.ZerobusSdk.<clinit>(ZerobusSdk.java:72)
```

Digger into the published jar:
For version 1.0.1:
```
$ jar tf zerobus-ingest-sdk-1.0.1.jar | grep native
native/
native/osx-aarch64/
native/linux-aarch64/
native/osx-x86_64/
native/windows-x86_64/
native/linux-x86_64/
native/osx-aarch64/libzerobus_jni.dylib
native/linux-aarch64/libzerobus_jni.so
native/osx-x86_64/libzerobus_jni.dylib
native/windows-x86_64/zerobus_jni.dll
native/linux-x86_64/libzerobus_jni.so
```

For version 1.2.0:
```
$ jar tf zerobus-ingest-sdk-1.2.0.jar | grep native
```
No output

### Reproduction

```java
package example;

import com.databricks.zerobus.ZerobusSdk;

public class ReproductionMain {
    public static void main(String[] args) {
        var zerobus = new ZerobusSdk("", "");
    }
}

```

### Expected behavior

For the native library to load.

### Is it a regression?

Yes

### Logs

```shell
Exception in thread "main" java.lang.UnsatisfiedLinkError: Native library failed to load: Native library not found in classpath: /native/linux-x86_64/libzerobus_jni.so. Platform: linux-x86_64. Make sure the correct native JAR is on the classpath.
	at com.databricks.zerobus.NativeLoader.ensureLoaded(NativeLoader.java:63)
	at com.databricks.zerobus.ZerobusSdk.<clinit>(ZerobusSdk.java:72)
	at example.ReproductionMain.main(ReproductionMain.java:7)
```

teodordelibasic-db:
[...]

PR #515, opened 2026-07-14
[Java] Guard against publishing JARs without JNI libs

teodordelibasic-db:

This PR fixes the Java SDK native-library packaging regression reported in #334.

The Java `1.2.0` artifact was published without any `native/.../libzerobus_jni` entries, so `NativeLoader` failed at runtime with errors like:

```text
Native library not found in classpath: /native/linux-x86_64/libzerobus_jni.so
```

This PR adds two release safeguards:

- Adds a Maven enforcer rule that fails `mvn package` / `install` / `deploy` unless the release JNI libraries are staged under `java/src/main/resources/native/`.
- Adds an explicit `release-java.yml` verification step that inspects the built deployable JARs and fails if required JNI entries are missing.

The required entries now match the JNI artifacts built by `release-jni.yml`:

- `native/linux-x86_64/libzerobus_jni.so`
- `native/linux-aarch64/libzerobus_jni.so`
- `native/linux-musl-x86_64/libzerobus_jni.so`
- `native/linux-musl-aarch64/libzerobus_jni.so`
- `native/windows-x86_64/zerobus_jni.dll`

This PR also documents `-Dzerobus.skipNativeLibCheck=true` for local Java-only builds, where developers usually do not have release JNI artifacts staged. Release builds should not use that skip flag.

## How is this tested?

- Parsed `java/pom.xml` with Python's XML parser.
- Parsed `.github/workflows/release-java.yml` with PyYAML.
- Manually verified the required native-library list matches the artifacts produced by `release-jni.yml`.

### DIFF
--- a/java/README.md
+++ b/java/README.md
@@ -215,7 +215,7 @@ Clone and build the SDK:
 ```bash
 git clone https://github.com/databricks/zerobus-sdk.git
 cd zerobus-sdk/java
-mvn clean package
+mvn clean package -Dzerobus.skipNativeLibCheck=true
 ```
 
 This generates two JAR files in the `target/` directory:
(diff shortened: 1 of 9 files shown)


java/README.md
elenagaljak-db, line 509: Should the flag be put here too? Or no

java/tools/README.md
elenagaljak-db: And here too?

elenagaljak-db:
Also here myb https://github.com/databricks/zerobus-sdk/blob/main/java/tools/generate_proto.sh#L22?

java/README.md
teodordelibasic-db: Not needed here, this snippet builds the reader's own application JAR (`my-zerobus-app-…`), not the SDK. The `zerobus.skipNativeLibCheck` property and the native-lib enforcer only exist in the SDK's own `pom.xml`, so a user's build never triggers that check.

java/tools/README.md
teodordelibasic-db: Done.

elenagaljak-db (approved):

PR #517, opened 2026-07-14
[Java] Support glibc 2.26 JNI builds

teodordelibasic-db:

This PR lowers the glibc baseline for the Java SDK's bundled Linux JNI libraries to glibc 2.26.

Specifically, the Java JNI release workflow now builds the GNU/Linux x86_64 and aarch64 artifacts with `cargo-zigbuild` targets:

- `x86_64-unknown-linux-gnu.2.26`
- `aarch64-unknown-linux-gnu.2.26`

The packaged native resource paths stay unchanged (`linux-x86_64`, `linux-aarch64`), so Java runtime loading and downstream packaging continue to work as before.

This is needed because the current published Java SDK JNI libraries require a newer glibc than Amazon Linux 2 provides. That blocks customers running environments such as Mule Runtime 4.11.6e on Amazon Linux 2. Building against a glibc 2.26 baseline lets the standard Java SDK JAR run there without requiring customers to compile their own native library.

The PR also documents the supported glibc floor in the Java README and adds a Java changelog entry.

## How is this tested?

- Built x86_64 JNI library locally:
  - `cargo zigbuild --release -p zerobus-jni --target x86_64-unknown-linux-gnu.2.26`
- Built aarch64 JNI library locally:
  - `cargo zigbuild --release -p zerobus-jni --target aarch64-unknown-linux-gnu.2.26`
- Verified the resulting Linux JNI shared libraries require at most `GLIBC_2.25`, which is compatible with glibc 2.26 environments.
- Parsed `.github/workflows/release-jni.yml` with Python YAML loading.

### DIFF
--- a/.github/workflows/release-jni.yml
+++ b/.github/workflows/release-jni.yml
@@ -13,15 +13,19 @@ jobs:
           - os: linux-ubuntu-latest
             runner: { group: databricks-protected-runner-group, labels: linux-ubuntu-latest }
             target: x86_64-unknown-linux-gnu
+            build_target: x86_64-unknown-linux-gnu.2.26
             artifact: libzerobus_jni.so
             native_dir: linux-x86_64
             cross: false
+            zigbuild: true
           - os: linux-ubuntu-latest
             runner: { group: databricks-protected-runner-group, labels: linux-ubuntu-latest }
             target: aarch64-unknown-linux-gnu
+            build_target: aarch64-unknown-linux-gnu.2.26
             artifact: libzerobus_jni.so
             native_dir: linux-aarch64
-            cross: true
+            cross: false
+            zigbuild: true
           - os: linux-ubuntu-latest
             runner: { group: databricks-protected-runner-group, labels: linux-ubuntu-latest }
             target: x86_64-unknown-linux-musl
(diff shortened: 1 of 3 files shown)


.github/workflows/release-jni.yml
elenagaljak-db, line 114: Build uses matrix.build_target, but upload still looks under matrix.target. Cargo writes artifacts to the exact triple passed to --target, so the glibc Linux jobs should fail at upload with if-no-files-found: error.
i think this should fix it:
`path: rust/target/${{ matrix.build_target || matrix.target }}/release/${{ matrix.artifact }}`

teodordelibasic-db: I checked this against cargo-zigbuild behavior. For glibc-versioned targets like
x86_64-unknown-linux-gnu.2.26, cargo-zigbuild strips the .2.26 suffix before
passing the target to Cargo/Rust, and keeps the suffix only for Zig's linker
target. The artifact is written under target/x86_64-unknown-linux-gnu/release,
not target/x86_64-unknown-linux-gnu.2.26/release.

I also verified this locally with the JNI builds: both x86_64 and aarch64
artifacts landed under the unversioned target directories. So the current upload
path using matrix.target is intentional.

elenagaljak-db (approved):

PR #518, opened 2026-07-14
[Rust][FFI] Add failure-path test that the ack callback Arc is released when create_stream fails

### LINKED ISSUE #491: [Rust][FFI] Add failure-path test that the ack callback Arc is released when create_stream fails
## Summary

Add an FFI-crate test proving that when `zerobus_sdk_create_stream` (and `..._with_headers_provider`) *fails*, the `Arc<CallbackAckCallback>` is dropped rather than leaked to a background task. This pins down an invariant that is currently only safe by construction, so a future refactor can't silently regress it.

## Background

Raised as H1 in review of the C++ ack callback PR (#484). On the create-stream failure path, the callback `Arc` drops with the builder before any background callback-handler task holds it — so there is no use-after-free or leak today. But nothing *enforces* that ordering: a future change that spawns the handler task earlier in the build/create path (before the failure is detected) would leak the callback (and its `user_data`) with no test to catch it.

A C++-side RAII guard cannot enforce this — on the C++ side the callback is owned by a caller `shared_ptr` that outlives the failed call, and the actual ordering lives in the Rust FFI (`rust/ffi/src/common.rs` / `stream.rs`). The invariant must be verified where it lives: the FFI crate.

## Proposal

In `rust/ffi/src/tests.rs`, add a test that:
- registers an ack callback via the C stream-config options,
- forces `create_stream` to fail (e.g. invalid args / a builder that fails to build),
- asserts the callback's `Arc` strong count drops to zero after the failed call (e.g. hold a `Weak` and assert `upgrade()` is `None`, or use a drop-flag `user_data`), i.e. no background task retained it.

Cover both create paths (`zerobus_sdk_create_stream` and `zerobus_sdk_create_stream_with_headers_provider`).

## Related

- #484 (C++ ack callback) — where H1 was raised.
- Part of #437 (ack callback across the C FFI). Complements the live-stream teardown / use-after-free test tracked in #469 (success-path teardown vs. this failure-path release).

zlata-stefanovic-db:
## Summary

Adds the failure-path test requested in #491 (raised as H1 in review of the C++ ack callback PR #484): prove that when `zerobus_sdk_create_stream` — and `..._with_headers_provider` — *fails*, the `Arc<CallbackAckCallback>` is dropped rather than leaked to a background handler task.

Today this is safe only by construction: `apply_c_stream_options` registers the ack `Arc` into the builder, and the builder is dropped when `build()` returns an error. Nothing *enforces* that ordering, so a future refactor that spawned the handler task earlier in the create path would silently leak the callback (and its `user_data`) with no test to catch it. This pins the invariant. It's the failure-path sibling of the success-path teardown lifetime test.

## What the tests do

- `test_create_stream_releases_ack_arc_on_failure` → real `zerobus_sdk_create_stream`
- `test_create_stream_with_headers_provider_releases_ack_arc_on_failure` → real `zerobus_sdk_create_stream_with_headers_provider`

Each registers an ack callback via the C stream-config options, forces creation to fail, asserts the failure surfaced a non-empty error message, and asserts the callback `Arc` was dropped exactly once.

**Hermetic failure trigger:** `RecordType::Json` + an empty table name. JSON skips the "Proto descriptor required" early-return, so the path reaches `apply_c_stream_options` (the ack `Arc` *is* registered into the builder), then fails in `build()` validation on the empty table name — before any channel/network I/O. The empty table name is the only invalid field, so a failed create is attributable to validation; the tests assert failure + a non-empty error message without pinning the exact wording.

## Observation mechanism

`CallbackAckCallback` is created internally and has no observable drop in the shipped build, so a test can't hold a `Weak` to it. `common.rs` adds a **`#[cfg(test)]`-gated** `Drop` impl that bumps a counter — but only when `user_data` matches a dedicated `ACK_DROP_SENTINEL_CREATE_FAIL_TESTS` address, so these tests observe the release of the callback *they* registered and are not perturbed by the existing ack-bridge tests (which use null `user_data`).

- Gated on `#[cfg(test)]` (not a cargo feature), so the shipped library is byte-identical — no `Cargo.toml`/feature change, no production code path.
- Parallelism: the counter is process-global, so the two tests take a dedicated mutex and each captures a before/after delta immediately around the single failing call, asserting `== 1`.

## Also included

- Tightens the ack-callback lifetime doc block and the `zerobus_alloc_header_array` / `zerobus_alloc_cstring` docstrings in `common.rs` (comment-only; same contract, fewer words). `zerobus.h` is regenerated by cbindgen to match — no hand edits.
- Adds a `rust/ffi/NEXT_CHANGELOG.md` entry under Internal Changes.

## Testing

- `cargo test -p zerobus-ffi --lib` → 73 passed, 0 failed (both new tests present; existing ack-bridge tests unaffected)
- `cargo build -p zerobus-ffi` (default features) → clean; the drop hook is absent from the non-test build
- `cargo fmt --check` clean; `cargo clippy -p zerobus-ffi --tests` introduces no new warnings

Closes #491.


irinatomic-db (dismissed):
I like the overall design.

I'm just worried about the fact that we take for granted that the Arc<CallbackAckCallback> was registered. It makes sense that we can't assert on that, but let's make it clear that that's a possibility in case of a test failure. So let's just expand the tests failure message to be smth like "expected the ack callback Arc to be dropped exactly once (0 = leaked OR never registered)"

zlata-stefanovic-db:
> I like the overall design.
> 
> I'm just worried about the fact that we take for granted that the Arc was registered. It makes sense that we can't assert on that, but let's make it clear that that's a possibility in case of a test failure. So let's just expand the tests failure message to be smth like "expected the ack callback Arc to be dropped exactly once (0 = leaked OR never registered)"

Documented now and changed the failure message, thank you!

irinatomic-db (approved):

PR #523, opened 2026-07-15
[Rust] Make mux backpressure tests deterministic

danilonajkov-db:

  - Replace timing-based mux backpressure tests with explicitly gated mock acknowledgments.
  - Poll ingest futures directly to verify they remain pending while their selected lane is full.
  - Release acknowledgments deterministically instead of relying on sleeps and short timeouts.

  ## Testing

  - `cargo test -p tests --test multiplexed_stream_tests`
  - `cargo fmt --all -- --check`

  All 21 multiplexed-stream tests pass.


rust/tests/src/multiplexed_stream_tests.rs
```diff
@@ -371,19 +373,14 @@ mod multi_stream_tests {
...
-            !third_task.is_finished(),
+            matches!(poll!(&mut third_future), Poll::Pending),
             "Expected ingest to wait on the chosen stream even when another sibling is free"
         );
 
-        let third = tokio::time::timeout(Duration::from_millis(500), third_task)
-            .await
-            .expect("ingest should complete once the chosen lane drains")??;
+        slow_ack.release();
+        let third = third_future.await?;
```
teodordelibasic-db: These backpressure tests used to wrap the blocked ingest in `tokio::time::timeout(Duration::from_millis(500), ...)`, so a wiring mistake failed the test in under a second. The rewrite replaces that with `slow_ack.release(); future.await?` and there is no bound on the await anymore. The streams here use `flush_timeout_ms: None` (from `default_options()`), and [`MultiplexedStream::wait_for_capacity`](https://github.com/databricks/zerobus-sdk/blob/becf512c23be0dd05fb4728373ceb8b100125380/rust/sdk/src/multiplexed_stream.rs#L164) is an unbounded backoff loop with no timeout. So if the gated ack is never sent (for example an offset mismatch sends the mock down the non-waiting `else` branch, or the request never reaches `gate.wait()`), the landing-zone permit never frees and the await hangs forever.

Could we keep the gate for determinism but wrap the post-release completion in a generous `tokio::time::timeout` so a regression stays bounded and points at the affected test? The same applies to the two sibling tests at [`multiplexed_stream_tests.rs:461`](https://github.com/databricks/zerobus-sdk/blob/becf512c23be0dd05fb4728373ceb8b100125380/rust/tests/src/multiplexed_stream_tests.rs#L461) and [`multiplexed_stream_tests.rs:537`](https://github.com/databricks/zerobus-sdk/blob/becf512c23be0dd05fb4728373ceb8b100125380/rust/tests/src/multiplexed_stream_tests.rs#L537), and it would be good to cover the trailing `wait_for_message_id` calls too.

rust/tests/src/multiplexed_stream_tests.rs
```diff
@@ -371,19 +373,14 @@ mod multi_stream_tests {
...
 
-        let mux_for_task = Arc::clone(&mux);
-        let third_task =
-            tokio::spawn(async move { mux_for_task.ingest_record(b"waited".to_vec()).await });
-
-        tokio::time::sleep(Duration::from_millis(100)).await;
+        let mut third_future = Box::pin(mux.ingest_record(b"waited".to_vec()));
         assert!(
-            !third_task.is_finished(),
+            matches!(poll!(&mut third_future), Poll::Pending),
```
teodordelibasic-db, line 391: A single `poll!` returning `Pending` only means "blocked on backpressure" because an unblocked ingest happens to resolve in one poll today. Could we make that explicit and harder to break, by adding a `tokio::task::yield_now().await` and a second `poll!` that still expects `Pending` before `release()`? Same at [`multiplexed_stream_tests.rs:456`](https://github.com/databricks/zerobus-sdk/blob/becf512c23be0dd05fb4728373ceb8b100125380/rust/tests/src/multiplexed_stream_tests.rs#L456) and [`multiplexed_stream_tests.rs:532`](https://github.com/databricks/zerobus-sdk/blob/becf512c23be0dd05fb4728373ceb8b100125380/rust/tests/src/multiplexed_stream_tests.rs#L532).

rust/tests/src/mock_grpc.rs
```diff
@@ -32,6 +57,11 @@ pub enum MockResponse {
         ack_up_to_offset: i64,
         delay_ms: u64,
     },
+    /// Successful record acknowledgment released explicitly by a test
+    GatedRecordAck {
```
teodordelibasic-db, line 66: `mock_grpc.rs` is included in three test binaries, but the new `GatedRecordAck` variant and `MockResponseGate::{new, release}` are only used by `multiplexed_stream_tests`. Building `rust_tests` and `proxy_tests` now emits:

```
warning: variant `GatedRecordAck` is never constructed
warning: associated items `new` and `release` are never used
```

The existing `CloseStreamSignal` and `Error` variants already carry `#[allow(dead_code)]` for this reason. It does not fail lint, but it adds noise to every test build. Could we annotate the new variant and methods to match?

rust/tests/src/multiplexed_stream_tests.rs
danilonajkov-db: done

teodordelibasic-db (approved):
LGTM

PR #524, opened 2026-07-15
[Java] Add headers provider

### LINKED ISSUE #386: [Java] Support custom HeadersProvider authentication
### SDK

Java

### Description

The Java SDK currently only supports clientId + clientSecret Oauth flows, while the other SDKs and rust core support custom identity flows via  `HeadersProvider`:
- Rust: `HeadersProvider` trait and `stream_builder().headers_provider(...)`, including Arrow streams.
- Python: subclassable `HeadersProvider`, passed via `headers_provider=` or Arrow-specific provider creation methods.
- TypeScript: public `HeadersProvider` interface and optional `headers_provider` support in stream creation.
- Go: public `HeadersProvider` interface with `CreateStreamWithHeadersProvider(...)` and `CreateArrowStreamWithHeadersProvider(...)`.

This prevents custom identity flows from being used with the Java SDK as is currently supported by all the other language SDKs.

### Proposed Solution

Add Java SDK support for a custom authentication headers provider, matching the capability already exposed in the Rust, Python, TypeScript, and Go SDKs. This will let Java users implement their own identity flow and pass the resulting Zerobus authentication headers into stream creation, without requiring long-lived OAuth client credentials in the SDK configuration.

### Additional Context

_No response_

danilonajkov-db:

Adds custom HeadersProvider authentication to the Java SDK, matching the Rust stream-builder API.

  - Adds the HeadersProvider interface with getHeaders() and optional invalidate().
  - Adds StreamBuilder.headersProvider() as an alternative to .oauth().
  - Supports JSON, Protocol Buffer, and Arrow streams.
  - Retains the provider for stream recovery and explicit stream recreation.
  - Keeps the deprecated legacy create*Stream APIs unchanged.
  - Ensures JNI callbacks run on blocking threads and reclaim local references after every invocation.
  - Validates custom gRPC metadata keys fallibly instead of allowing invalid names to panic.
  - Documents that header names should come from a small, fixed set.

  ### Authentication Behavior

  .oauth() and .headersProvider() are mutually exclusive. Calling either replaces the previously configured authentication method.

  When the server rejects credentials, the SDK calls HeadersProvider.invalidate() before retrying so providers can clear cached tokens.

  ## Testing

  - cargo fmt --all --check
  - cargo check -p zerobus-jni
  - cargo test -p databricks-zerobus-ingest-sdk invalid_custom_header_name_returns_error
  - Java 8-compatible compilation of the SDK and provider harness
  - Staging E2E:
      - Provider fetched an OAuth token inside getHeaders()
      - Created a JSON stream through StreamBuilder.headersProvider()
      - Ingested a record
      - Flushed successfully


java/examples/json/README.md
```diff
@@ -49,6 +49,24 @@ ZerobusJsonStream stream = sdk.streamBuilder()
     .join();
 ```
 
+For custom authentication, provide the authorization and table name headers:
```
teodordelibasic-db, line 52: I suppose you intentionally added this only to this example? I am fine with it, just checking.

rust/jni/src/headers_provider.rs
```diff
@@ -0,0 +1,187 @@
...
+        .call_method(&throwable, "toString", "()Ljava/lang/String;", &[])
+        .ok()?
+        .l()
+        .ok()?;
+    let message = JString::from(message);
+    env.get_string(&message).ok().map(Into::into)
+}
+
+fn provider_error(message: impl Into<String>) -> ZerobusError {
+    ZerobusError::InvalidArgument(format!("Java HeadersProvider error: {}", message.into()))
```
teodordelibasic-db, line 266: Every failure in the JNI bridge, whether an exception thrown by `getHeaders()`, a JVM attach failure, a `spawn_blocking` join failure, or a null map, is wrapped as `ZerobusError::InvalidArgument`. That variant is classified non-retryable in [`rust/sdk/src/errors.rs:70`](https://github.com/databricks/zerobus-sdk/blob/de7aa59e4855e68d50489ba434b6f03aa4ec4712/rust/sdk/src/errors.rs#L70), so if a provider's `getHeaders()` hits a transient failure during background recovery (its own token endpoint returns 503 or times out), the SDK treats it as permanent and stops recovering. The built-in OAuth path behaves differently: it splits transient network and server failures into the retryable `TokenFetchError` ([`rust/sdk/src/errors.rs:86`](https://github.com/databricks/zerobus-sdk/blob/de7aa59e4855e68d50489ba434b6f03aa4ec4712/rust/sdk/src/errors.rs#L86)) and keeps retrying.

Could we give the provider a way to signal retryability? We could add a test where the provider fails once and then succeeds on retry.

java/src/test/java/com/databricks/zerobus/StreamBuilderTest.java
```diff
@@ -273,10 +273,22 @@ void jsonBuildRoutesToCreateJsonStreamInternal() {
...
+            eq("cat.sch.json"), eq("json-id"), eq("json-secret"), eq(null), opts.capture());
     assertEquals(4242, opts.getValue().maxInflightRecords());
-    verify(sdk, never()).createProtoStreamInternal(any(), any(), any(), any(), any());
-    verify(sdk, never()).createArrowStreamInternal(any(), any(), any(), any(), any());
+    verify(sdk, never()).createProtoStreamInternal(any(), any(), any(), any(), any(), any());
+    verify(sdk, never()).createArrowStreamInternal(any(), any(), any(), any(), any(), any());
+  }
+
+  @Test
+  void headersProviderRoutesToCreateJsonStreamInternal() throws Exception {
```
teodordelibasic-db, line 285: It would be good if possible to add few more cases so the behavior we rely on is pinned down: `oauth()` and `headersProvider()` clearing each other, provider-only `validateRequired()`, and the JNI callback in `rust/jni/src/headers_provider.rs` running for real (a successful map and a provider that throws, plus `invalidate()`).

rust/sdk/src/stream/grpc/connection.rs
```diff
@@ -173,3 +171,29 @@ impl ZerobusStream {
         }
     }
 }
+
+fn insert_custom_header(
+    metadata: &mut MetadataMap,
+    key: &'static str,
+    value: &str,
+) -> ZerobusResult<()> {
+    let key = MetadataKey::from_bytes(key.as_bytes())
```
teodordelibasic-db: Could we add an entry in `rust/NEXT_CHANGELOG.md`, something like rejecting invalid custom metadata header names with `InvalidArgument` instead of panicking during stream connection setup?

rust/jni/src/headers_provider.rs
```diff
@@ -0,0 +1,187 @@
...
+            .map_err(|error| java_call_error(env, "iterate headers map", error))?;
+        let Some((key, value)) = entry else {
+            break;
+        };
+
+        if key.is_null() || value.is_null() {
+            return Err(provider_error("header names and values must not be null"));
+        }
+
+        let key_string = JString::from(key);
```
teodordelibasic-db, line 149: After the null checks, `extract_headers` casts each key and value with `JString::from(...)` and hands them to `env.get_string`. A `Map` whose entries are not `String` (possible under type erasure or a buggy implementation) produces an opaque JNI error or a CheckJNI abort rather than a clear `Java HeadersProvider error: ...`. The Python binding avoids this by extracting typed `(str, str)` pairs.

Could we check the entries are `java.lang.String` with `is_instance_of` before `get_string` and return a clear provider error otherwise?

rust/jni/src/headers_provider.rs
```diff
@@ -0,0 +1,187 @@
...
+fn java_call_error(env: &mut JNIEnv<'_>, method: &str, error: JniError) -> ZerobusError {
+    let detail = if matches!(error, JniError::JavaException) {
+        take_java_exception_message(env).unwrap_or_else(|| error.to_string())
+    } else {
+        error.to_string()
+    };
+    provider_error(format!("{method} failed: {detail}"))
+}
+
+fn take_java_exception_message(env: &mut JNIEnv<'_>) -> Option<String> {
```
teodordelibasic-db: If `Throwable.toString()` throws, the `.ok()?` short-circuits and returns, but the new exception is left pending on the thread. These callbacks run on daemon-attached Tokio blocking threads that get reused, so a later JNI call on the same thread can fail immediately because of the stale pending exception.

Could we do the `toString()` extraction and then clear the exception state unconditionally before returning, so the thread is left clean whether or not extraction succeeded? Pulling the extraction into a helper and calling `env.exception_clear()` after it on both paths would do it:

```rust
fn take_java_exception_message(env: &mut JNIEnv<'_>) -> Option<String> {
    let throwable = env.exception_occurred().ok()?;
    env.exception_clear().ok()?;

    // Any call below can raise a fresh pending exception; clear it afterwards
    // so a reused daemon thread is never left dirty.
    let message = extract_throwable_string(env, &throwable);
    let _ = env.exception_clear();
    message
}

fn extract_throwable_string(env: &mut JNIEnv<'_>, throwable: &JObject<'_>) -> Option<String> {
    let message = env
        .call_method(throwable, "toString", "()Ljava/lang/String;", &[])
        .ok()?
        .l()
        .ok()?;
    env.get_string(&JString::from(message)).ok().map(Into::into)
}
```

The trailing `env.exception_clear()` runs on both the success and `None` paths, so it also covers the case where `toString()` succeeds but reading the resulting string fails, which leaves an exception pending too. The caller already falls back to `error.to_string()` when this returns `None`, so no message is lost.

If possible a good test would be a provider that throws a `Throwable` whose `toString()` also throws, then asserts a second `getHeaders()` call on the same provider still succeeds. The second call is what proves the thread was not left dirty.

teodordelibasic-db (commented):
Mostly edge cases and races flagged by LLM.

rust/jni/src/headers_provider.rs
```diff
@@ -0,0 +1,187 @@
...
+        .ok()?;
+    let message = JString::from(message);
+    env.get_string(&message).ok().map(Into::into)
+}
+
+fn provider_error(message: impl Into<String>) -> ZerobusError {
+    ZerobusError::InvalidArgument(format!("Java HeadersProvider error: {}", message.into()))
+}
+
+unsafe impl Send for JavaHeadersProvider {}
```
teodordelibasic-db: Also, not sure in general whether we are okay with this...?

java/examples/json/README.md
danilonajkov-db: yeah, I can add it to all of them but I don't think its needed. Most people would just use the default .oauth

rust/jni/src/headers_provider.rs
danilonajkov-db: made the errors retryable by default, and made NonRetriableException and malformed output non retryable. Thanks

java/src/test/java/com/databricks/zerobus/StreamBuilderTest.java
danilonajkov-db: added a few more tests throughout for these conditions

rust/jni/src/headers_provider.rs
danilonajkov-db: added a check, thanks

rust/jni/src/headers_provider.rs
danilonajkov-db: yep makes sense, fixed

rust/jni/src/headers_provider.rs
danilonajkov-db: yep JavaHeadersProvider already satisfies Send/Sync through its fields, so the unsafe implementations can be dropped

rust/sdk/src/stream/grpc/connection.rs
danilonajkov-db: I decided to not change the rust behaviour in this PR as it affects other sdks as well. I've just added validation on the java side that returns InvalidArgument

rust/jni/src/headers_provider.rs
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    pub fn into_arc(self) -> Arc<dyn HeadersProvider> {
+        Arc::new(self)
+    }
+}
+
+#[async_trait]
+impl HeadersProvider for JavaHeadersProvider {
+    async fn get_headers(&self) -> ZerobusResult<HashMap<&'static str, String>> {
```
teodordelibasic-db, line 48: A callback that exceeds `recoveryTimeoutMs` keeps running after the stream attempt times out. Tokio cannot cancel a `spawn_blocking` task once it has started, so dropping the future in [`rust/sdk/src/stream/grpc/connection.rs:80`](https://github.com/databricks/zerobus-sdk/blob/9633e61025cdff75184d61ec82b6371ac226eafd/rust/sdk/src/stream/grpc/connection.rs#L80) leaves the Java call holding its worker and cloned `GlobalRef`. The next retry can start another callback, and repeated stream attempts can accumulate blocked workers and retained provider objects.

Could we serialize calls per provider and hold the guard inside the blocking closure, so cancellation cannot release it while the Java call is still running?

```rust
pub struct JavaHeadersProvider {
    provider_ref: GlobalRef,
    callback_lock: Arc<tokio::sync::Mutex<()>>,
}

let guard = Arc::clone(&self.callback_lock).lock_owned().await;
let provider_ref = self.provider_ref.clone();
tokio::task::spawn_blocking(move || {
    let _guard = guard;
    get_headers_blocking(provider_ref)
})
.await
```

This would not make a synchronous Java call cancellable, but it would keep one stuck provider adapter from consuming a new worker on every retry. A process-wide bounded executor would give a stronger global bound.

danilonajkov-db: yep makes sense. Added a lock, they are now serialized. Also added a test to match this

teodordelibasic-db (dismissed):
LGTM

teodordelibasic-db (approved):

PR #525, opened 2026-07-16
[Rust] Fix variant Arrow schema

teodordelibasic-db:

Fix Rust SDK Arrow schema generation for Unity Catalog `VARIANT` columns.

The SDK previously mapped `VARIANT` to Arrow `LargeUtf8`, matching the protobuf path but not Arrow Flight ingest. The local Shinkansen server expects Arrow `VARIANT` columns as a nullable struct with non-null `metadata: LargeBinary` and `value: LargeBinary`, so Arrow stream setup failed with a schema mismatch before ingestion could start.

This PR adds a Rust SDK schema type for `VARIANT` and keeps the projections separate:
- Protobuf schema generation still treats `VARIANT` as `string`.
- Arrow schema generation emits `Struct<metadata: LargeBinary not null, value: LargeBinary not null>`.
- Regression tests cover top-level and nested `VARIANT`, plus protobuf compatibility.

## How is this tested?

- `cargo test -p databricks-zerobus-ingest-sdk schema::tests --features arrow-flight`
- `cargo clippy -p databricks-zerobus-ingest-sdk --features arrow-flight -- -D warnings`
- `make check`
- Local Shinkansen harness with a one-column `VARIANT` Delta table: pre-fix `LargeUtf8` Arrow schema fails stream setup; fixed SDK-derived Arrow schema succeeds.

### DIFF
--- a/rust/sdk/src/schema.rs
+++ b/rust/sdk/src/schema.rs
@@ -40,7 +40,8 @@
 //!
 //! | Unity Catalog type            | Proto type | Encoding contract                           |
 //! |-------------------------------|------------|---------------------------------------------|
-//! | `STRING`, `VARIANT`, `DECIMAL`| `string`   | `STRING`/`DECIMAL`: UTF-8 text; `VARIANT`: JSON text for protobuf ingestion |
+//! | `STRING`, `DECIMAL`           | `string`   | UTF-8 text                                  |
+//! | `VARIANT`                     | `string`   | JSON text for protobuf ingestion            |
 //! | `INT`, `INTEGER`              | `int32`    |                                             |
 //! | `LONG`, `BIGINT`              | `int64`    |                                             |
 //! | `SHORT`, `SMALLINT`, `BYTE`, `TINYINT` | `int32` | zero-extended; range-checked by the server |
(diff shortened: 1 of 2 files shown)


rust/sdk/src/schema.rs
```diff
@@ -40,7 +40,7 @@
 //!
 //! | Unity Catalog type            | Proto type | Encoding contract                           |
 //! |-------------------------------|------------|---------------------------------------------|
-//! | `STRING`, `VARIANT`, `DECIMAL`| `string`   | UTF-8 text                                  |
+//! | `STRING`, `VARIANT`, `DECIMAL`| `string`   | `STRING`/`DECIMAL`: UTF-8 text; `VARIANT`: JSON text for protobuf ingestion |
```
andrijast-db: Why not just separate table row for Variant?

teodordelibasic-db: Fair enough, fixed. 🙂

danilonajkov-db (approved):
lgtm, but it is a little bit weird that we require clients to do the variant conversion (json string -> metadata and value) themselves instead of doing it in our server like for proto cc @nikolaobradovic-db

andrijast-db (approved):

nikolaobradovic-db:
> lgtm, but it is a little bit weird that we require clients to do the variant conversion (json string -> metadata and value) themselves instead of doing it in our server like for proto cc @nikolaobradovic-db

If they are already using Arrow to ingest, they can leverage the Arrow ecosystem to generate it properly. One problem that I see is validation; it's up to them to validate it/correctly generate the variant type.

PR #529, opened 2026-07-17
[FFI] Fix headers-provider use-after-free via ownership transfer

### LINKED ISSUE #436: [FFI] HeadersProvider can be freed while an in-flight recovery get_headers() call runs (UAF)
## Summary

A custom `HeadersProvider` can be **freed by the wrapper while a Rust worker thread is still inside a synchronous `get_headers()` callback into it** — a use-after-free. The dangerous call happens during stream **recovery** (re-establishing the gRPC connection), and stream teardown frees the provider without waiting for that in-flight callback to finish. Cancellation is attempted but cannot preempt a synchronous FFI call.

> **This issue was originally filed as an `_nowait`-specific bug. That framing was wrong** — see "Correction" below. The real bug is a recovery-vs-teardown race that is **not** specific to `_nowait` and also affects the blocking ingest APIs.

## Correction: `_nowait` is not the trigger

The original report (and the C++ `headers_provider.hpp` WARNING / the C++ `_nowait` guard) claimed a detached `_nowait` task calls back into a freed provider. Tracing the code shows that's not the mechanism:

- The `_nowait` task only calls `ingest_record_offset` → `ingest_internal_v2` (`rust/sdk/src/stream/grpc/ingest.rs`), which **purely enqueues** into the landing zone. It checks `is_closed` (returns `StreamClosedError` if the stream is closing), takes a lock, assigns an offset, appends. **It never touches the headers provider, never connects, never sends.**
- `get_headers()` is called **only by the supervisor task** while (re)establishing a connection (`supervisor.rs:88` → `connection.rs:74`).
- Both `close()` (`close.rs`) and `Drop` (`mod.rs:216`) cancel the `CancellationToken` and `abort()` the supervisor.

So a late-running `_nowait` task just appends to a landing zone whose supervisor is already cancelled/aborted — no reconnect, no `get_headers`, no UAF. The `_nowait`-vs-provider guard in C++ appears **mis-scoped** (harmless, but not the actual fix).

## The actual reachable race

`get_headers` is reached only during a **recovery** that is in flight at the moment of teardown. Recovery is driven by the **server/network** (a server-initiated pause/close or a dropped connection — `receiver.rs:139-157`), not by the user, so it can run concurrently with the application dropping the stream. The pieces that turn that into a UAF:

1. **Synchronous callback.** The FFI provider's `get_headers` calls the C function pointer directly:
   ```rust
   // rust/ffi/src/common.rs:211
   let c_headers = (self.callback)(self.user_data);   // synchronous; derefs the C/C++ object
   ```
   It is not `spawn_blocking` and does not `.await` during the call, so tokio's `abort()` (cooperative, effective only at `.await` points) **cannot interrupt a worker thread executing it**.

2. **Short, non-re-awaited abort.** `close()` waits only `SHUTDOWN_TIMEOUT_SECS = 1` (`close.rs:17`) for the supervisor to exit, then calls `supervisor_task.abort()` and **does not re-await it** (`close.rs:73-84`). If `get_headers` takes >1s — typical for a custom provider that fetches/rotates a token over the network — `zerobus_stream_close` returns while the worker is still inside the C++ callback.

3. **Provider freed with no handshake.** The wrapper frees the provider immediately after `close`. In C++, `~Stream` runs `zerobus_stream_close` then implicitly destroys the `shared_ptr<HeadersProvider>` member — freeing the C++ object while a worker thread may still be in (or returning from) the callback. The Rust side only ever held a raw `user_data` pointer (ownership never crossed the FFI), so nothing keeps the C++ object alive.

### Interleaving

```
Worker thread (supervisor)              App thread
──────────────────────────             ──────────
server pause → recovery
create_stream_connection()
  get_headers().await
    (self.callback)(user_data)  ← slow  drop Stream → ~Stream / close()
    [token fetch > 1s]                    flush, cancel, wait 1s, abort()
                                          (abort can't preempt the sync call)
                                          zerobus_stream_close returns
                                          free provider object
    callback still derefs object ──────►  USE-AFTER-FREE
```

This requires a custom provider whose `get_headers` blocks longer than the 1s shutdown timeout during a recovery that coincides with teardown. Narrow and timing/network-dependent — a latent, hard-to-reproduce UAF, not a deterministic crash.

### Not `_nowait`-specific

The same race applies to a stream created with a custom provider and torn down (drop or `close`) while a recovery `get_headers` is in flight — regardless of whether `_nowait` was ever used. `_nowait` is irrelevant to it.

## Proposed fix (FFI/core)

The root cause is that the **provider object's lifetime is not coupled to the Rust-side code that can call it.** Fix it where the lifetime contract lives — the C FFI:

1. **Take ownership of the provider at the FFI boundary.** Have `zerobus_sdk_create_stream_with_headers_provider` accept a `user_data` plus a `free`/`destroy` callback, and make the Rust `CallbackHeadersProvider` own that `user_data` and invoke the destroy callback **only when the last `Arc` referencing it drops** — i.e. after every task that could call `get_headers` is gone. The wrapper then hands ownership across instead of keeping its own `shared_ptr`, so the object cannot be freed while a callback can still run.
2. **And/or fence in-flight callbacks before declaring teardown complete** — ensure no worker thread is inside `(self.callback)(user_data)` before the provider can be destroyed (the cooperative `abort()` + 1s timeout in `close.rs` is insufficient against a synchronous FFI call).

Once the FFI guarantees this, the C++ `_nowait`-vs-provider guard can be revisited (it does not address this race).

## Acceptance criteria

- A custom `HeadersProvider` cannot be freed while any task can still invoke `get_headers` (covered by an ASan/TSan stress test: custom provider with a slow `get_headers`, force a server-driven recovery, drop/close the stream mid-callback).
- Holds for both blocking and `_nowait` ingest.
- FFI header (`zerobus.h`) documents the ownership/free contract for the provider `user_data`.
- Re-evaluate the C++ `_nowait`-vs-provider guard (`cpp/src/stream.cpp`) and the `headers_provider.hpp` WARNING wording against the corrected mechanism.

## References

- Synchronous callback: `rust/ffi/src/common.rs:211` (and `CallbackHeadersProvider`, `common.rs:179-197`)
- `get_headers` call site (recovery/connect): `rust/sdk/src/stream/grpc/connection.rs:74`; supervisor `supervisor.rs:88`; recovery triggers `receiver.rs:139-157`
- Teardown: `rust/sdk/src/stream/grpc/close.rs:17,59-84`; `mod.rs:216` (Drop); `zerobus_stream_free` `rust/ffi/src/stream.rs:200`
- Enqueue-only ingest (exonerates `_nowait`): `rust/sdk/src/stream/grpc/ingest.rs` (`ingest_internal_v2`)
- C++ side: `cpp/src/stream.cpp` (`~Stream`, `_nowait` guard), `cpp/include/zerobus/headers_provider.hpp` (WARNING)
- Origin: PR #420 review (@irinatomic-db), comment on `cpp/include/zerobus/stream.hpp:70`

zlata-stefanovic-db:
## Summary

Fixes a use-after-free where a custom `HeadersProvider` could be freed by the wrapper (Go/C++/.NET) while a Rust worker thread was still inside a **synchronous** `get_headers()` callback during connection recovery. Cooperative `abort()` cannot preempt a synchronous FFI call, and `close()` waited only `SHUTDOWN_TIMEOUT_SECS = 1` before aborting, so a slow provider (e.g. one fetching/rotating a token) racing teardown could be invoked on freed memory.

## Approach: ownership transfer at the FFI boundary

The C FFI now **takes ownership** of the provider `user_data`. `zerobus_sdk_create_stream_with_headers_provider`, `zerobus_sdk_create_arrow_stream_with_headers_provider`, and `zerobus_sdk_create_stream_with_headers_provider_async` gain a nullable `free_user_data` destroy callback. `CallbackHeadersProvider` invokes it from `Drop`, which runs only after the last `Arc` referencing the provider drops — and the supervisor task holds its own `Arc` clone across the in-flight `get_headers()` call, so the destroy is guaranteed to run only after any synchronous callback has returned. The provider `Arc` is constructed before any fallible work, so `free_user_data` fires **exactly once on every path** (success after the last reference drops; failed create before returning). Wrappers hand ownership across and never free `user_data` themselves.

## Changes by layer

- **Rust FFI** (`rust/ffi`): `free_user_data` param + `Drop`-based destroy (panic-contained via `catch_unwind`) on both the sync and async create-with-provider entry points; regenerated `zerobus.h`; ownership/thread-safety docs.
- **Go** (`go/ffi.go`, `go/arrow_ffi.go`): hands the `cgo.Handle` to the FFI via `goFreeHeadersProvider`; removes the per-stream handle registry. `goFreeHeadersProvider` nil-guards the handle before `Delete()`.
- **C++** (`cpp/`): passes a heap `std::shared_ptr<HeadersProvider>` + `zerobus_cpp_headers_free` destroy trampoline; `Stream`/`ArrowStream` no longer keep the provider.
- **.NET** (`dotnet/`): adds `HeadersProviderFreeCallback` + the `free_user_data` param to both the sync and async P/Invoke bindings, and hands the provider's `GCHandle` ownership to the FFI (released via `HeadersProviderBridge.NativeFree` after any in-flight `GetHeaders` returns) so the stream no longer frees it on dispose. Removes the now-dead borrow-and-free-on-dispose machinery on the async path. See the scope note below.

## Scope: which SDKs are affected

The bug class is specific to passing a **borrowed** `user_data` across the C FFI while the core holds a raw pointer.

| SDK | Provider bridge | Affected? |
|-----|-----------------|-----------|
| Go, C++ | borrowed raw pointer / handle freed on close | **Yes — fixed here** |
| .NET (sync + async paths) | borrowed `GCHandle` freed on dispose | **Yes — fixed here** |
| Python | `HeadersProviderWrapper` owns a `PyObject`, handed as an owned `Arc` | No — Rust's Arc controls lifetime |
| TypeScript | headers resolved once at create → `StaticHeadersProvider` | No — no live JS object across FFI |
| Java | no custom headers-provider feature (OAuth only) | No — feature absent |

> **Note:** the original scope table listed .NET as unaffected because the .NET SDK did not exist when this PR was opened. It landed on `main` afterward with a `HeadersProvider` feature that consumed the old FFI signature, so this branch's FFI change broke `cross-sdk-dotnet` (a native test-host crash from the stale P/Invoke arity) until .NET was updated in lockstep — now done here.

## Async create-with-provider now fixed

The `_async` create-with-provider entry point originally kept the borrowed-ownership model. **This is now closed:** `zerobus_sdk_create_stream_with_headers_provider_async` takes a `free_user_data` parameter and constructs the provider `Arc` before any fallible work, so it frees `user_data` exactly once on every path — synchronously for scheduling failures (before returning), in the spawned task for asynchronous creation failures, and after the last reference drops on success. The .NET async path (`CreateStreamWithHeadersProviderAsync`, the SDK's default) now transfers ownership like the sync path and no longer frees the `GCHandle` on dispose. This matters because the .NET SDK's default headers-provider surface is the async one, so without this the recovery-vs-teardown UAF was still reachable there.

## ⚠️ Verification caveat: .NET not compiled locally

The Rust changes are built and tested (`cargo build`, `cargo clippy`, 87 FFI tests pass, including a new async sync-failure free-once regression test). **The .NET changes were not compiled** — there is no `dotnet` SDK in the authoring environment. They were verified by reading and cross-checking every call site (P/Invoke arity, delegate rooting, no double-free on the failure path, dead-code removal), but the .NET build and tests must be confirmed in CI before merge.

## Tests

- Rust: `free_user_data` fires once on `Drop` and on failed create; null destroy is a no-op; a teardown test reproducing the recovery-vs-teardown race (a blocking in-flight `get_headers` on one `Arc` clone while another drops) asserts the free is deferred until the callback returns; and the async variant frees `user_data` once on a synchronous scheduling failure.
- Go: `goFreeHeadersProvider` releases the `cgo.Handle`.
- C++: trampoline dispatch + free destroys the provider; null tolerance. Passes under ASan.
- .NET: `NativeFree` releases the provider `GCHandle` (observed via a `WeakReference` — `GCHandle` is a struct, so the underlying handle can only be observed indirectly) and tolerates a null pointer. The pre-existing `HeadersProviderIntegrationTests` exercise both create paths against the updated bindings (run in CI — see caveat above).

## Breaking change (FFI)

`zerobus_sdk_create_stream_with_headers_provider`, `zerobus_sdk_create_arrow_stream_with_headers_provider`, and `zerobus_sdk_create_stream_with_headers_provider_async` gain a `free_user_data` parameter — a non-additive `zerobus.h` change. Filed under Breaking Changes in `rust/ffi/NEXT_CHANGELOG.md`. All in-repo C FFI consumers (Go, C++, .NET) are updated in this PR.

## Follow-up required before Go picks this up

This changes the FFI signature, so an FFI release (`ffi/v*`) must land and regenerate **all five** platform `libzerobus_ffi.a` archives in `go/lib/` before the Go SDK can consume it. The committed archives are intentionally left at their pre-PR builds (the locally-rebuilt one exceeded GitHub's 100 MB limit and would have been an inconsistent single-platform artifact). Go CI's cross-SDK job builds the FFI from local Rust source, so it validates this PR without the prebuilt libs.


rust/ffi/src/tests.rs
```diff
@@ -916,6 +918,133 @@ fn test_callback_headers_provider_returns_headers() {
     assert!(headers.contains_key("Authorization"));
 }
 
+// The provider owns `user_data`: dropping it must invoke `free_user_data`
```
irinatomic-db, line 921: Note for myself: 

I'll be working on building the SDK testing infra with a gRPC mock server - to add the true full-path test there: real recovering stream, `get_headers` blocking on a worker thread during recovery while we tear the stream down, assert free fires once and only after the callback returns. Needs the mock to inject a disconnect + a blockable/counting provider + client-side free observation.

rust/ffi/src/stream.rs
```diff
@@ -492,10 +503,27 @@ pub extern "C" fn zerobus_sdk_create_stream_with_headers_provider(
     descriptor_proto_len: usize,
     headers_callback: HeadersProviderCallback,
     user_data: *mut std::ffi::c_void,
+    // Written inline (not via HeadersProviderFreeCallback) so cbindgen emits a
+    // nullable C function pointer instead of an opaque struct.
+    free_user_data: Option<extern "C" fn(user_data: *mut std::ffi::c_void)>,
```
irinatomic-db, line 508: Nit: These 3 spots (here, :624, arrow.rs:226) must stay spelled `Option<extern "C" fn(..)>` inline - using the `HeadersProviderFreeCallback` alias makes cbindgen emit an opaque struct instead of a nullable `fn ptr`, silently breaking the ABI. Right now only a comment stops someone "tidying" it to the alias. Can we add a CI check that diffs the committed zerobus.h against a fresh cbindgen run? That'd catch it instead of relying on reviewers.

dotnet/src/Zerobus/ZerobusSdk.cs
```diff
@@ -431,17 +436,21 @@ private ZerobusStream CreateStreamWithHeadersProviderCore(
...
                 GCHandle.ToIntPtr(handle),
+                bridge.FreeCallback,
                 ref nativeOpts);
         }
         catch
         {
-            handle.Free();
+            // On a thrown failure the FFI already invoked the free callback (the
+            // provider is built before any fallible work), so the handle is
+            // already freed — do NOT free it again here (double-free).
```
irinatomic-db, line 451: From Claude:

"Don't double-free" only holds because the P/Invoke is the sole thing that can throw in this try - and it always frees on failure. If anyone later adds a throwing line before the native call, the handle leaks. Can you shrink the try to wrap only the P/Invoke call, so this stays true by construction rather than by comment?

cpp/src/headers_callback.cpp
```diff
@@ -115,5 +123,14 @@ extern "C" CHeaders zerobus_cpp_headers_trampoline(void* user_data) {
   return result;
 }
 
+extern "C" void zerobus_cpp_headers_free(void* user_data) noexcept {
```
irinatomic-db, line 126: From Claude:

This now runs the provider's destructor on a tokio worker thread, not the thread that created the stream. Go/.NET call this out to users - can you add the same "cleanup may run on any thread" warning to the public HeadersProvider header so C++ users don't get surprised?

rust/ffi/src/stream.rs
zlata-stefanovic-db: Implemented — added a CI guard in Rust CI to ensure zerobus.h stays in sync with cbindgen output.

dotnet/src/Zerobus/ZerobusSdk.cs
zlata-stefanovic-db: Updated — I narrowed the ownership-transfer section to be safer by construction.

cpp/src/headers_callback.cpp
zlata-stefanovic-db: Done — this warning is now documented in the public C++ HeadersProvider header (cpp/include/zerobus/headers_provider.hpp).

dotnet/src/Zerobus/ZerobusSdk.cs
```diff
@@ -454,30 +466,39 @@ private async Task<ZerobusStream> CreateStreamWithHeadersProviderCoreAsync(
...
-                    GCHandle.ToIntPtr(handle),
+                    tableName,
+                    descriptorProto,
+                    bridge.Callback,
+                    handlePtr,
+                    bridge.FreeCallback,
                     nativeOpts)
                 .ConfigureAwait(false);
         }
         catch
```
irinatomic-db: Nit: This catch just rethrows with no cleanup so the try/catch is a no-op. 

Let's remove it then. And move the ownership transfer comment to sit above the `SdkCreateStreamWithHeadersProviderAsync` call

irinatomic-db (dismissed):

dotnet/src/Zerobus/ZerobusSdk.cs
zlata-stefanovic-db: removed

irinatomic-db (approved):

PR #530, opened 2026-07-17
[C++] Instrument the Rust FFI under ASan/TSan sanitizer builds

zlata-stefanovic-db:
## Summary

Follow-up to the TSan CI job added in the C++ test-coverage work (#499). That job ran ThreadSanitizer over only the C++ side — the FFI was a plain `cargo build --release`, so the Rust core (where `descriptor_bytes()` / `encode_json()` actually run) was uninstrumented and TSan could not see races inside it. This PR instruments the Rust FFI too, so `address`/`thread` sanitizer builds genuinely cover the core, not just the wrapper.

This is option (a) from @irinatomic-db's review comment on the TSan job.

## How

- **`cpp/cmake/BuildRustFfi.cmake`** — when `ZEROBUS_SANITIZE` is `address`/`thread`, build the FFI with nightly `-Zsanitizer=<san>` + `-Zbuild-std --target <host>`. Each sanitizer gets an **isolated cargo target dir** (`target/sanitize-<san>/`), because the `--target` output path is identical across sanitizers and changing only `RUSTFLAGS` doesn't invalidate cargo's cache — a shared dir would link a stale archive built for the other sanitizer.
- **`cpp/Makefile`** — for `address`/`thread`, build the C++ side with clang (`CLANG_CC`/`CLANG_CXX`, default `clang`/`clang++`). The instrumented Rust archive emits LLVM sanitizer runtime calls (`__tsan_*` / `__asan_*`); GCC's sanitizer runtime can't resolve them, so the C++ link must use clang to share LLVM's compiler-rt runtime. The default (non-sanitized) build is unchanged (GCC, plain FFI).
- **`.github/workflows/ci-cpp.yml`** — both sanitizer jobs switch to nightly + `rust-src` and install clang.
- Docs: `concurrency_test.cpp` header, `cpp/CLAUDE.md`, `cpp/NEXT_CHANGELOG.md` updated to reflect that the core is now instrumented.

## Validation

Verified locally with clang-18:

- **TSan:** FFI archive carries 8991 `__tsan` symbols; the test binary links the LLVM TSan runtime and runs under "ThreadSanitizer v3"; all 13 tests pass.
- **ASan:** FFI archive carries 27828 `__asan` symbols and 0 `__tsan` (isolation confirmed); all 13 tests pass.
- **Default GCC build (no sanitizer):** unchanged, plain `target/release` archive, 13/13 pass.

## Notes / trade-offs

- Sanitizer builds now require a nightly toolchain with `rust-src` and rebuild `std` (`-Zbuild-std`), so those CI jobs are slower (~1 min extra for the std rebuild). The normal build/test path is untouched.
- No `ProtoSchema` race is expected (it's immutable with read-only borrows); the value is that a future regression introducing one would now be caught inside the core.

### DIFF
--- a/cpp/CLAUDE.md
+++ b/cpp/CLAUDE.md
@@ -53,7 +53,9 @@ Run from `cpp/`:
   (`address`, `thread`, or `undefined`; CMake option `-DZEROBUS_SANITIZE=`). Off
   by default. Targets the memory/lifetime bugs this FFI wrapper is prone to
   (use-after-free, double-free). `address`/`thread` also instrument the Rust FFI
-  (nightly `-Zsanitizer` + `-Zbuild-std`), so they need a nightly toolchain with
+  (nightly `-Zsanitizer` + `-Zbuild-std`) — a sanitizer only observes code
+  compiled with it, so leaving the core uninstrumented would miss the races/UAF
+  that live inside the FFI. This needs a nightly toolchain with
   `rust-src` and build the C++ side with clang to match the LLVM sanitizer
   runtime (`CLANG_CC`/`CLANG_CXX`, default `clang`/`clang++`); GCC can't resolve
   the instrumented archive's runtime symbols. `undefined` uses the default
@@ -151,6 +153,10 @@ callback](#ack-callback)).
   be thread-safe with respect to their own state.
 - A single `ProtoSchema` may be used by concurrent readers
   (`descriptor_bytes`, `encode_json`); destruction must not race those calls.
+  `tests/concurrency_test.cpp` exercises this under ThreadSanitizer (which
+  instruments the Rust core too), so a race in the read path is caught in CI.
+  The destruction-vs-read contract, though, is not tested — it is the caller's
+  responsibility, not something the TSan job enforces.
 
 ## Breaking change rules
 
--- a/cpp/CLAUDE.md
+++ b/cpp/CLAUDE.md
@@ -52,7 +52,12 @@ Run from `cpp/`:
 - `make test SANITIZE=address` — Build + run the suite under a sanitizer
   (`address`, `thread`, or `undefined`; CMake option `-DZEROBUS_SANITIZE=`). Off
   by default. Targets the memory/lifetime bugs this FFI wrapper is prone to
-  (use-after-free, double-free) with no added dependency.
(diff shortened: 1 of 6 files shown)


cpp/cmake/BuildRustFfi.cmake
```diff
@@ -32,7 +32,54 @@ else()
   find_program(CARGO_EXECUTABLE cargo REQUIRED
       DOC "Path to the cargo build tool")
 
+  # When a sanitizer is requested, instrument the Rust core too — otherwise TSan
+  # / ASan only watch the thin C++ wrapper and miss races/UAF inside the FFI,
+  # where the real work runs. -Zsanitizer + -Zbuild-std are nightly-only and
```
irinatomic-db: Nit: duplicate comment on line :38 and :56: `-Zsanitizer + -Zbuild-std are nightly-only and need a --target...`

zlata-stefanovic-db: Done now

cpp/CLAUDE.md
```diff
@@ -52,7 +52,14 @@ Run from `cpp/`:
 - `make test SANITIZE=address` — Build + run the suite under a sanitizer
   (`address`, `thread`, or `undefined`; CMake option `-DZEROBUS_SANITIZE=`). Off
   by default. Targets the memory/lifetime bugs this FFI wrapper is prone to
-  (use-after-free, double-free) with no added dependency.
+  (use-after-free, double-free). `address`/`thread` also instrument the Rust FFI
```
irinatomic-db, line 55: Let's also add a small sentence why we need this. Something like "Instrumenting the FFI (core as well) because a sanitizer can only observer instrumented code"

cpp/CLAUDE.md
irinatomic-db, line 154: Let's also add here that the concurrent-readers contract is not actively enforced by the thread sanitizer job?

irinatomic-db (dismissed):

cpp/CLAUDE.md
zlata-stefanovic-db: Added

cpp/CLAUDE.md
zlata-stefanovic-db: Added

irinatomic-db (approved):

PR #538, opened 2026-07-20
[ArrowFlight] Bound in-flight batches by ack lifetime; make recovery replay failure-safe

nikolaobradovic-db:

- Add an inflight semaphore (capacity = max_inflight_batches) so ingest_batch bounds batches awaiting ack, not just pre-encode buffering — fixes unbounded pending_batches growth and applies real backpressure. Permit held on PendingBatch until ack/terminal/recovery-drop.
- Reject max_inflight_batches == 0 in build_arrow (+ guard in new()); previously panicked the zero-capacity channel.
- Recovery replay now rebuilds/installs pending_batches + counter under the lock before sending; a replay-send failure retains pending (with permits) for the next attempt instead of dropping it.

## How is this tested?

New backpressure_tests (blocks-until-ack, zero rejected) + replay_* unit tests (send-failure retention, in-order replay, partial-slice/drop).

### DIFF
--- a/rust/sdk/src/arrow_stream.rs
+++ b/rust/sdk/src/arrow_stream.rs
@@ -1309,6 +1309,10 @@ impl ZerobusArrowStream {
 
         // Acquire the backpressure permit BEFORE ingest_mutex: reconnect() holds that
         // mutex, so blocking on a permit while holding it could stall recovery.
+        // `inflight` is never closed, so the map_err is unreachable defensive code; a
+        // close during the wait is handled by the is_closed re-check below (not by
+        // semaphore closure), and permit-blocked ingests wake when acks or
+        // move_pending_to_failed free permits.
         let permit = Arc::clone(&self.inflight)
             .acquire_owned()
             .await
@@ -1359,6 +1363,11 @@ impl ZerobusArrowStream {
         let sender = match sender {
             Some(s) => s,
             None => {
+                // Narrow recovery-handoff window (sender nulled before reconnect). The
+                // batch intentionally stays in pending_batches (holding its permit) so
+                // it remains recoverable/replayable, even though this attempt returns
+                // an error. Permit is freed when the batch is later acked, moved to the
+                // failed set, or the stream closes.
                 if let Some(server_error) = self.server_error_rx.borrow().clone() {
                     return Err(server_error);
                 }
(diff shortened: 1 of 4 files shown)


rust/sdk/src/arrow_stream.rs
```diff
@@ -56,7 +56,9 @@ pub(crate) struct ArrowTableProperties {
 }
 
 /// A pending batch waiting for acknowledgment.
-#[derive(Clone)]
+///
+/// Not `Clone`: owns an `inflight` permit released (RAII) when the batch leaves
```
elenagaljak-db: nit: Do we need this comment

rust/sdk/src/arrow_stream.rs
teodordelibasic-db: We should probably update this since it doesn't return immediately.

teodordelibasic-db, line 920: Preexisting and narrow window, but seems plausible.

Since `last_acked_records.store(0)` and `cumulative_records_sent.store(0)` happen before `ingest_mutex.lock().await`, a concurrent `ingest_batch` (which still runs its `fetch_add` + pushes to pending even while `is_paused`) can win the mutex race after the reset, get `start_record = 0`, and create a batch whose fabricated low range falls under `acked_before_disconnect` - `slice_batch_for_recovery` drops it as "fully acked" - silent data loss.

We can fix by moving the two `store(0)` calls under `ingest_mutex` (ideally into `replay_pending_batches`).

nikolaobradovic-db: Aware of this one, I have a branch with the fix, which will go after this PR as a second follow-up.

nikolaobradovic-db: Fixed.

rust/sdk/src/arrow_stream.rs
nikolaobradovic-db: Removed.

rust/sdk/src/arrow_stream.rs
nikolaobradovic-db: Actually Ill send this one first.

elenagaljak-db (approved):

rust/sdk/src/arrow_stream.rs
nikolaobradovic-db: https://github.com/databricks/zerobus-sdk/pull/558

PR #540, opened 2026-07-20
[RUST] Introduce InvalidSchema error and return more details more from the server

flaviofcruz:

Zerobus server can return extra error details when creating a schema and the table schema is incompatible. Parse it and return it to the SDK caller with the different reasons for why the schema failed.

Users of the SDK can then decide what to do when the schema differs slightly.

## How is this tested?

Added unit tests.

### DIFF
--- a/rust/NEXT_CHANGELOG.md
+++ b/rust/NEXT_CHANGELOG.md
@@ -6,13 +6,17 @@
 
 ### New Features and Improvements
 
-- Arrow Flight: schema-validation rejections during stream setup now surface as
-  the new `ZerobusError::InvalidSchema` variant (carrying the server-reported
-  `causes` as typed `SchemaValidationCause` values) instead of a generic
-  `CreateStreamError`. This lets callers detect a table/stream schema mismatch —
-  e.g. a column added to or dropped from the target table — and re-resolve their
-  schema rather than treating it as an opaque invalid-argument failure. The
-  variant is not SDK-retryable.
+- Arrow Flight: schema-validation rejections now surface as the new
+  `ZerobusError::InvalidSchema` variant (carrying the server-reported `causes` as
+  typed `SchemaValidationCause` values) instead of a generic `CreateStreamError`.
+  This lets callers detect a table/stream schema mismatch — e.g. a column added
+  to or dropped from the target table — and re-resolve their schema rather than
+  treating it as an opaque invalid-argument failure. The variant is not
+  SDK-retryable. This applies both to initial stream setup and to mid-stream
+  reconnects: previously a schema change detected during recovery was retried
+  until the recovery budget drained and then reported as a generic failure;
+  now the non-retriable `InvalidSchema` is surfaced to a blocked `wait_for_offset`
+  (or `flush`) immediately so callers can rebuild the stream without downtime.
 
 ### Bug Fixes
 
--- a/rust/NEXT_CHANGELOG.md
+++ b/rust/NEXT_CHANGELOG.md
@@ -6,6 +6,14 @@
 
 ### New Features and Improvements
 
+- Arrow Flight: schema-validation rejections during stream setup now surface as
+  the new `ZerobusError::InvalidSchema` variant (carrying the server-reported
+  `causes` as typed `SchemaValidationCause` values) instead of a generic
+  `CreateStreamError`. This lets callers detect a table/stream schema mismatch —
+  e.g. a column added to or dropped from the target table — and re-resolve their
+  schema rather than treating it as an opaque invalid-argument failure. The
+  variant is not SDK-retryable.
+
 ### Bug Fixes
 
 - **Arrow Flight — `close()` now propagates flush errors** (Beta): `ZerobusArrowStream::close()` previously swallowed a failed final `flush()` and always returned `Ok(())`, contradicting its documentation and diverging from the proto stream's `close()`. It now returns the flush error after still tearing down the stream and moving pending batches to the failed set (retrievable via `get_unacked_batches()`).
--- a/rust/sdk/Cargo.toml
+++ b/rust/sdk/Cargo.toml
@@ -32,6 +32,7 @@ tokio-stream.workspace = true
 tokio-util = { workspace = true, features = ["rt"] }
 tonic = { workspace = true, features = ["tls-native-roots", "transport"] }
 tonic-prost.workspace = true
+tonic-types.workspace = true
 tracing.workspace = true
 hyper-http-proxy.workspace = true
 hyper-util.workspace = true
--- a/rust/sdk/Cargo.toml
+++ b/rust/sdk/Cargo.toml
@@ -32,7 +32,7 @@ tokio-stream.workspace = true
 tokio-util = { workspace = true, features = ["rt"] }
 tonic = { workspace = true, features = ["tls-native-roots", "transport"] }
 tonic-prost.workspace = true
-tonic-types.workspace = true
+tonic-types = { workspace = true, optional = true }
 tracing.workspace = true
 hyper-http-proxy.workspace = true
 hyper-util.workspace = true
@@ -64,7 +64,7 @@ prost-build = { version = "0.14", optional = true }
 [features]
 default = []
 # Arrow Flight is in Beta
-arrow-flight = ["dep:arrow-flight", "dep:arrow-array", "dep:arrow-schema", "dep:arrow-ipc", "dep:futures"]
+arrow-flight = ["dep:arrow-flight", "dep:arrow-array", "dep:arrow-schema", "dep:arrow-ipc", "dep:futures", "dep:tonic-types"]
 # Zero-copy protobuf parser.
 zeroparser = ["dep:self_cell", "dep:prost-build"]
 testing = ["dep:futures"]
--- a/rust/sdk/src/arrow_stream.rs
+++ b/rust/sdk/src/arrow_stream.rs
@@ -731,10 +731,34 @@ impl ZerobusArrowStream {
                                 // Loop continues with new stream.
                             }
                             Ok(Err(e)) => {
-                                // Mirror the initial-connect path: drop the cached
-                                // token on auth rejection so recovery re-mints.
                                 if e.is_auth_rejection() {
+                                    // Mirror the initial-connect path: drop the cached
+                                    // token on auth rejection so recovery re-mints, then
+                                    // fall through to retry with the fresh token.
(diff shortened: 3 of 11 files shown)


rust/sdk/src/arrow_stream.rs
teodordelibasic-db, line 733: This is preexisting, so we can fix in a separate PR as well. cc: @nikolaobradovic-db 

Reconnect setup classifies a schema rejection into the non-retryable `InvalidSchema` (at [`arrow_stream.rs:880`](https://github.com/databricks/zerobus-sdk/blob/890ffc37566f77aa545c664c0acd89892e200640/rust/sdk/src/arrow_stream.rs#L880)), but the supervisor's `Ok(Err(e))` arm then throws `e` away and swaps in a synthetic `Status::unavailable("Reconnection failed")`. `Unavailable` is retryable, so recovery loops until `recovery_retries` runs out, and callers waiting on `flush()` or an offset get a generic failure instead of `InvalidSchema { message, causes }`.

We could branch on `e.is_retryable()` in this arm, matching the primary supervisor arm: for a non-retryable error, set `is_closed`, publish `e` to waiters via `server_error_tx`, move pending batches to the failed set, and return `e` instead of injecting `Unavailable`?

A regression test would help: drop a stream with an unacked batch, have reconnect return a schema status, and assert the waiter gets `InvalidSchema`, recovery stops, and the batch is retrievable.

rust/sdk/src/errors.rs
```diff
@@ -56,6 +126,36 @@ const UNRETRIABLE_STATUS_CODES: &[tonic::Code] = &[
...
 
 impl ZerobusError {
+    /// Classify a `tonic::Status` returned by the server during stream setup.
+    ///
+    /// If the server attached a schema-validation `ErrorInfo` detail (reason
+    /// `SCHEMA_VALIDATION_FAILED`), returns [`ZerobusError::InvalidSchema`]
+    /// carrying the structured `causes` the server reported.
+    /// Otherwise falls back to [`ZerobusError::CreateStreamError`], preserving
+    /// the original status (and its gRPC code) for retry/auth classification.
+    pub(crate) fn from_setup_status(status: tonic::Status) -> Self {
```
teodordelibasic-db, line 156: The feature relies on the `ErrorInfo` details surviving `FlightError::Tonic(status).into()`, since production calls `from_setup_status(flight_error.into())` at [`arrow_stream.rs:579`](https://github.com/databricks/zerobus-sdk/blob/890ffc37566f77aa545c664c0acd89892e200640/rust/sdk/src/arrow_stream.rs#L579).

The new tests build a `tonic::Status` directly, so they skip that conversion.

Could we add a test under `#[cfg(feature = "arrow-flight")]` that wraps a schema-validation `Status` in `FlightError::Tonic(Box::new(status)).into()` and asserts `from_setup_status` still yields `InvalidSchema` with the expected `causes`? The mock Flight server already has an error path (`ServerBehavior::Error`), so a setup-time and reconnect-time test through it would also cover waiter notification and retry counts.

rust/sdk/src/errors.rs
```diff
@@ -56,6 +126,36 @@ const UNRETRIABLE_STATUS_CODES: &[tonic::Code] = &[
...
+        if let Some(info) = status.get_details_error_info() {
+            if info.reason == SCHEMA_VALIDATION_REASON {
```
teodordelibasic-db: `from_setup_status` specializes to `InvalidSchema` whenever `info.reason == "SCHEMA_VALIDATION_FAILED"`, ignoring the gRPC code and `ErrorInfo.domain`. The server only ever attaches this detail with `Code::InvalidArgument` and domain `zerobus.databricks.com`, so the reason token is unique in practice

Since the check is cheap, could we also require the domain to equal `zerobus.databricks.com` (accepting empty for forward-compat) before specializing, so an unrelated status carrying the same reason stays a `CreateStreamError`? A negative test for a wrong domain would document the intent.

rust/sdk/src/errors.rs
```diff
@@ -31,6 +82,25 @@ pub enum ZerobusError {
...
+    /// `causes` carries the raw, machine-readable causes the server reported
+    /// (e.g. [`SchemaValidationCause::FieldNotInTable`]); a single rejection can
+    /// list several. The SDK deliberately does not interpret them: whether a
+    /// mismatch is recoverable — e.g. by re-resolving the table schema and
+    /// rebuilding the stream — is a caller-side policy decision. Per-field
+    /// detail (offending column names) is carried in `message` for diagnostics,
+    /// not structurally. This error is not SDK-retryable, since the SDK holds a
+    /// fixed schema and its recovery loop would re-send the same rejected schema.
+    #[error("Arrow schema does not match the target table: {message}.")]
+    InvalidSchema {
```
teodordelibasic-db, line 117: `InvalidSchema { message, causes }` keeps only the message and parsed causes, dropping the gRPC code, the domain, and the `error_code` metadata (the server always sets `error_code = 8001`, useful for telemetry correlation). The variant is also not `#[non_exhaustive]`, so adding a field later would be a breaking change even though the enum is. 

Could we either carry `error_code: Option<String>` now, or mark the variant `#[non_exhaustive]` so fields can be added later without a major bump?

rust/sdk/Cargo.toml
```diff
@@ -32,6 +32,7 @@ tokio-stream.workspace = true
 tokio-util = { workspace = true, features = ["rt"] }
 tonic = { workspace = true, features = ["tls-native-roots", "transport"] }
 tonic-prost.workspace = true
+tonic-types.workspace = true
```
teodordelibasic-db: Even though it will probably be used in the regular gRPC path as well, `tonic-types` is currently used only by Arrow, but still added unconditionally.

Could we make `tonic-types` optional, add it to the `arrow-flight` feature, and gate `from_setup_status` and its tests behind that feature?

rust/NEXT_CHANGELOG.md
```diff
@@ -6,6 +6,14 @@
 
 ### New Features and Improvements
 
+- Arrow Flight: schema-validation rejections during stream setup now surface as
```
teodordelibasic-db: Could we make the relevant update in the `rust/README.md` file as well.

rust/sdk/src/errors.rs
flaviofcruz: Done. Added test.

rust/sdk/src/arrow_stream.rs
flaviofcruz: I made the change for InvalidSchema only. Happy to remove if you guys are doing a fix on a separate PR, can remove it if necessary, just wanted to make it complete.

rust/sdk/src/errors.rs
flaviofcruz: Added more gates.

rust/sdk/Cargo.toml
flaviofcruz: Done.

rust/NEXT_CHANGELOG.md
flaviofcruz: Done.

rust/sdk/src/errors.rs
flaviofcruz: Added error_code and non_exhaustive.

rust/sdk/src/arrow_stream.rs
flaviofcruz: https://github.com/databricks/zerobus-sdk/pull/561

teodordelibasic-db: Sounds good, Nikola is handling it in his PR.

teodordelibasic-db (commented):
LGTM, but you will need to open a PR directly on main, not via fork, CI is broken for forks currently due to security.

Also, there are conflicts with Nikola's large change, so here is a compiled summary to hopefully ease rebasing:

#562 rewrites the same supervisor arm, `flush()`, and `wait_for_offset_internal` that this PR touches, so after it lands a few things here become redundant and should be dropped rather than merged. Rebasing this small PR onto that large one is the easier direction.

Once #562 is in `main`, here is what changes on this branch.

Keep as-is (the actual payload, no conflict expected):

- All of `errors.rs`: the `InvalidSchema` variant, `SchemaValidationCause`, `from_setup_status`, and the unit tests. #562 does not touch `errors.rs`.
- The two classification lines in `arrow_stream.rs` that upgrade a schema rejection from `CreateStreamError` to the typed `InvalidSchema`:
  - setup path (currently `arrow_stream.rs:579`): `Err(ZerobusError::from_setup_status(flight_error.into()))`
  - reconnect-setup path (currently `arrow_stream.rs:904`): same call
  These stay; they are what makes the error typed. Their line numbers will move after the rebase.
- README section, `NEXT_CHANGELOG.md` entry, `Cargo.toml` / `lib.rs` / lockfile changes.

Drop (now redundant under #562):

- The `else if matches!(e, ZerobusError::InvalidSchema { .. })` branch this PR added to the supervisor's `Ok(Err(e))` arm (currently around `arrow_stream.rs:739`), including its manual `is_closed.store`, `server_error_tx.send`, and `move_pending_to_failed`. #562 reworks that arm to carry any failed reconnect's real error into the next iteration and route it through a single terminal path guarded by `error.is_retryable()`. Since `InvalidSchema::is_retryable()` is `false`, a schema rejection on reconnect already terminates with its real cause under #562, publishing the error before and after finalization. The special-case branch would be dead code, so remove it and let the generic path handle it.

Reconcile (both PRs added similar things):

- Mock Flight variant: this PR added `MockFlightResponse::SetupError { status }`; #562 added `MockFlightResponse::FailSetup { status }` for the same purpose (reject during setup on the schema message). Collapse to one variant after the rebase and update whichever tests reference the removed name.
- `NEXT_CHANGELOG.md`: both add entries under the same release; keep both, textual only.
- `arrow_tests.rs`: both add tests; keep both, textual only.

Worth adding after the rebase (currently untested by either PR):

- A regression test for a `flush()` (or `wait_for_offset`) that starts after the schema-reconnect has already closed the stream. The existing reconnect test here only covers a waiter already parked in the `select!`; #562's precedence tests use generic terminal errors. A combined test (reject the reconnect with the schema `ErrorInfo`, let the supervisor close, then call `flush()`, assert `InvalidSchema` with the expected `causes`) proves the typed error survives #562's rewritten close path, which is the whole point of integrating the two.

Net: after #562, this PR is `errors.rs` plus two one-line classification calls plus docs and tests. The supervisor special-case comes out.

PR #557, opened 2026-07-21
[C++] Add release workflow and finalize v0.1.0 changelog

zlata-stefanovic-db:
## Summary

Adds the C++ SDK release pipeline (builder half) and finalizes the `v0.1.0` changelog, so the C++ SDK can cut its first release.

## What this does

- **`.github/workflows/release-cpp.yml`** — the builder workflow. Manually dispatched; for each of the 7 target platforms it cross-builds the Rust C FFI archive (`cargo-zigbuild`, mirroring `release-ffi.yml`) and bundles it with the platform-neutral C++ source tree into one artifact. Produces artifacts only — no GitHub Release.
- **`cpp/CHANGELOG.md`** — promoted the accumulated `v0.1.0` entry from `NEXT_CHANGELOG.md`.
- **`cpp/NEXT_CHANGELOG.md`** — reset to an empty `v0.2.0` template.
- **`cpp/CLAUDE.md`** — rewrote the Release section to describe the two-repo split and the FFI-prebuilt + C++-from-source distribution model.

## Distribution model

The C++ wrapper is shipped as **source** (its C++ ABI is not stable across compiler/stdlib), while the Rust C FFI archive (`libzerobus_ffi.a`, a stable C ABI, and the one piece a consumer can't build without a Rust toolchain) is shipped **prebuilt per platform**. A consumer builds only the thin C++ wrapper with their own compiler, pointing CMake at the bundled FFI archive via `-DZEROBUS_FFI_LIBRARY`. This matches the source-build model already documented in `cpp/CLAUDE.md` and mirrors how `release-go.yml` / `release-jni.yml` build from source.

## Release flow (two repos)

This is the **builder** half, in `zerobus-sdk`. The **orchestrator** half (`release-zerobus-sdk-cpp.yml`) lands separately in `databricks/secure-public-registry-releases-eng`: it validates the `cpp/v*` tag against `ZEROBUS_CPP_VERSION`, triggers this builder, downloads the artifacts, runs the mandatory security scan, and creates the GitHub Release on `zerobus-sdk`.

## Target platforms (7)

`linux-x86_64` (gnu), `linux-aarch64` (gnu, cross), `linux-musl-x86_64`, `linux-musl-aarch64`, `darwin-x86_64`, `darwin-aarch64`, and `windows-x86_64` (MSVC). The `windows-gnu-x86_64` (MinGW) leg was dropped during review — its cross-build fails on the missing mingw `dlltool`, and the MSVC leg already covers Windows.

## Testing

- Native install/build path verified locally: built `libzerobus_ffi.a`, then built `libzerobus_cpp.a` against it via `-DZEROBUS_FFI_LIBRARY=` — CMake logged "using prebuilt FFI library" (no cargo invoked; no Rust toolchain needed for the C++ half).
- Every FFI build leg exercised locally with each platform's exact workflow command (Zig 0.13.0, cross-compiling from Linux x86_64): all 4 Linux legs and both darwin legs produce a valid `libzerobus_ffi.a`. The darwin legs build only the staticlib (`cargo-zigbuild rustc … --crate-type staticlib`) because a plain `zigbuild` also links the manifest's `cdylib`, whose macOS `Security`/`CoreFoundation` framework link can't be satisfied when cross-compiling from Linux.
- `windows-x86_64` (MSVC) builds natively on a Windows runner, so it is validated by the workflow dispatch rather than locally.
- Both workflow YAML files parse.
- Changelog note extraction verified against the orchestrator's `awk` logic.

## Before the first real release

Dispatch this workflow once (or run the `dry-run: true` orchestrator) to green the full matrix on the real runners — in particular the `windows-x86_64` (MSVC) leg, which can't be exercised locally — before pushing the first `cpp/v0.1.0` tag.

### DIFF
--- a/.github/workflows/release-cpp.yml
+++ b/.github/workflows/release-cpp.yml
@@ -1,0 +1,196 @@
+name: Build C++
+run-name: "Build C++ SDK ${{ github.ref_name }}"
+
+# Builds one self-contained artifact per target platform for the C++ SDK. Each
+# artifact bundles:
+#   - the C++ SDK source tree (headers, sources, CMake build) — platform-neutral
+#   - the matching prebuilt Rust C FFI archive (libzerobus_ffi.a / zerobus_ffi.lib)
+#     plus zerobus.h, cross-built for that platform
+#
+# Rationale: the FFI archive is a C ABI (stable across C++ compilers) and is the
+# one piece a consumer can't build without a Rust toolchain, so we ship it
+# prebuilt per platform. The thin C++ wrapper is ABI-fragile across
+# compiler/stdlib, so we ship it as source and let the consumer's own compiler
+# build it, pointing CMake at the bundled FFI archive:
+#
+#   cmake -S <artifact>/cpp -B build \
+#     -DZEROBUS_FFI_LIBRARY=<artifact>/lib/libzerobus_ffi.a \
+#     -DZEROBUS_FFI_HEADER_DIR=<artifact>/lib
+#
+# The FFI build steps mirror release-ffi.yml (cargo-zigbuild for the cross
+# targets); no C++ cross-compilation is performed here.
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build-cpp:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - runner: { group: databricks-protected-runner-group, labels: linux-ubuntu-latest }
+            target: x86_64-unknown-linux-gnu
+            static_lib: libzerobus_ffi.a
+            artifact_dir: linux-x86_64
+            cross: false
+          - runner: { group: databricks-protected-runner-group, labels: linux-ubuntu-latest }
+            target: aarch64-unknown-linux-gnu
+            static_lib: libzerobus_ffi.a
+            artifact_dir: linux-aarch64
+            cross: true
+          - runner: { group: databricks-protected-runner-group, labels: linux-ubuntu-latest }
+            target: x86_64-unknown-linux-musl
+            static_lib: libzerobus_ffi.a
+            artifact_dir: linux-musl-x86_64
+            cross: false
+            zigbuild: true
+            rustflags: "-C target-feature=-crt-static"
+          - runner: { group: databricks-protected-runner-group, labels: linux-ubuntu-latest }
+            target: aarch64-unknown-linux-musl
+            static_lib: libzerobus_ffi.a
+            artifact_dir: linux-musl-aarch64
+            cross: false
+            zigbuild: true
+            rustflags: "-C target-feature=-crt-static"
+          - runner: { group: databricks-protected-runner-group, labels: linux-ubuntu-latest }
+            target: x86_64-apple-darwin
+            static_lib: libzerobus_ffi.a
+            artifact_dir: darwin-x86_64
+            cross: false
+            zigbuild: true
+            static_only: true
+          - runner: { group: databricks-protected-runner-group, labels: linux-ubuntu-latest }
+            target: aarch64-apple-darwin
+            static_lib: libzerobus_ffi.a
+            artifact_dir: darwin-aarch64
+            cross: false
+            zigbuild: true
+            static_only: true
+          - runner: { group: databricks-protected-runner-group, labels: windows-server-latest }
+            target: x86_64-pc-windows-msvc
+            static_lib: zerobus_ffi.lib
+            artifact_dir: windows-x86_64
+            cross: false
+          - runner: { group: databricks-protected-runner-group, labels: linux-ubuntu-latest }
+            target: x86_64-pc-windows-gnu
+            static_lib: libzerobus_ffi.a
+            artifact_dir: windows-gnu-x86_64
+            cross: false
+            zigbuild: true
+
+    name: Build C++ - ${{ matrix.artifact_dir }}
+    runs-on: ${{ matrix.runner }}
+    permissions:
+      contents: read
+      id-token: write
+
+    defaults:
+      run:
+        working-directory: rust
+
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - name: Get JFrog OIDC token
+        shell: bash
+        run: bash "$GITHUB_WORKSPACE/.github/scripts/jfrog-oidc-token.sh"
+
+      - name: Configure cargo for JFrog
+        shell: bash
+        run: bash "$GITHUB_WORKSPACE/.github/scripts/configure-cargo.sh"
+
+      - name: Install protoc
+        uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b # v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Install cross-compilation toolchain (Linux ARM64)
+        if: matrix.cross
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -y -qq gcc-aarch64-linux-gnu
+          mkdir -p .cargo
+          echo '[target.aarch64-unknown-linux-gnu]' > .cargo/config.toml
+          echo 'linker = "aarch64-linux-gnu-gcc"' >> .cargo/config.toml
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
+        with:
+          targets: ${{ matrix.target }}
+
+      - name: Install Zig (zigbuild targets)
+        if: matrix.zigbuild
+        uses: mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29 # v2.2.1
+        with:
+          version: 0.13.0
+
+      - name: Install cargo-zigbuild (zigbuild targets)
+        if: matrix.zigbuild
+        run: cargo install cargo-zigbuild --locked --version 0.19.8
+
+      - name: Build FFI library
+        if: '!matrix.zigbuild'
+        run: cargo build --release -p zerobus-ffi --target ${{ matrix.target }}
+
+      - name: Build FFI library (zigbuild)
+        if: matrix.zigbuild && !matrix.static_only
+        env:
+          RUSTFLAGS: ${{ matrix.rustflags }}
+        run: cargo zigbuild --release -p zerobus-ffi --target ${{ matrix.target }}
+
+      - name: Build FFI static library (zigbuild, static only)
+        if: matrix.zigbuild && matrix.static_only
+        env:
+          RUSTFLAGS: ${{ matrix.rustflags }}
+        run: cargo zigbuild rustc --release -p zerobus-ffi --target ${{ matrix.target }} --crate-type staticlib
+
+      - name: Assemble C++ SDK bundle
+        shell: bash
+        run: |
+          root="$GITHUB_WORKSPACE"
+          out="$root/artifacts/${{ matrix.artifact_dir }}"
+          mkdir -p "$out/cpp" "$out/lib"
+
+          # Platform-neutral C++ SDK source (everything CMake needs to build the
+          # wrapper: headers, sources, the CMake build, and top-level docs).
+          cp -R "$root/cpp/include"        "$out/cpp/"
+          cp -R "$root/cpp/src"            "$out/cpp/"
+          cp -R "$root/cpp/cmake"          "$out/cpp/"
+          cp -R "$root/cpp/examples"       "$out/cpp/"
+          cp "$root/cpp/CMakeLists.txt"    "$out/cpp/"
+          cp "$root/cpp/README.md"         "$out/cpp/"
+          cp "$root/cpp/CHANGELOG.md"      "$out/cpp/"
+          cp "$root/cpp/LICENSE"           "$out/cpp/"
+          cp "$root/cpp/NOTICE"            "$out/cpp/"
+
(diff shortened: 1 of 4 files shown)


.github/workflows/release-cpp.yml
```diff
@@ -0,0 +1,196 @@
...
+        if: matrix.zigbuild && matrix.static_only
+        env:
+          RUSTFLAGS: ${{ matrix.rustflags }}
+        run: cargo zigbuild rustc --release -p zerobus-ffi --target ${{ matrix.target }} --crate-type staticlib
+
+      - name: Assemble C++ SDK bundle
+        shell: bash
+        run: |
+          root="$GITHUB_WORKSPACE"
+          out="$root/artifacts/${{ matrix.artifact_dir }}"
```
irinatomic-db: Nit: consisenty with release-ffi.yml

This step writes the bundle to `$GITHUB_WORKSPACE/artifacts/...` (repo root) and the upload step uses `path: artifacts/...`, whereas the sibling `release-ffi.yml` writes relative to the `rust/` working directory (`artifacts/...` → `rust/artifacts`) and uploads `path: rust/artifacts/...`.

Both are correct — `upload-artifact`'s `path` resolves against `$GITHUB_WORKSPACE`, not the step's `working-directory`, so the repo-root form here lines up and I confirmed it produces a valid upload.

But the two release workflows now use opposite path conventions for the same operation. Consider matching `release-ffi.yml` (relative `rust/artifacts` + `path: rust/artifacts/...`), or dropping `defaults.run.working-directory: rust` and `cd rust` in the Cargo steps, so paths are uniform throughout the file.

irinatomic-db (dismissed):
Reviewed and exercised locally. The design (ship the C++ wrapper as source + a prebuilt per-platform Rust C FFI archive with a stable C ABI) is sound and matches the `ZEROBUS_FFI_LIBRARY` contract in `cpp/cmake/BuildRustFfi.cmake`.

I couldn't run the full matrix (no access to the protected runners, and the Zig cross-compile legs need Zig), but I replayed the native `linux-x86_64` leg end-to-end.

```sh
# 1. Build the FFI archive (exact workflow command for the linux-x86_64 leg)
cd rust
cargo build --release -p zerobus-ffi --target x86_64-unknown-linux-gnu
#    -> Finished `release` in ~40s; target/x86_64-unknown-linux-gnu/release/libzerobus_ffi.a

# 2. Replay the "Assemble C++ SDK bundle" step (matrix values inlined,
#    working-directory: rust, GITHUB_WORKSPACE=repo root)
#    -> 41 files, correct cpp/ + lib/ layout; libzerobus_ffi.a + zerobus.h under lib/

# 3. Build the C++ wrapper from the bundle exactly as the generated README instructs
cd artifacts/linux-x86_64
cmake -S cpp -B build \
  -DZEROBUS_FFI_LIBRARY="$PWD/lib/libzerobus_ffi.a" \
  -DZEROBUS_FFI_HEADER_DIR="$PWD/lib"
#    -> "-- Zerobus: using prebuilt FFI library …/lib/libzerobus_ffi.a" (no Cargo invoked)

cmake --build build -j
#    -> libzerobus_cpp.a + all 4 examples (json/proto single+batch) linked

# 4. Runtime sanity
./build/examples/json_single
#    -> "error: environment variable ZEROBUS_SERVER_ENDPOINT is not set" (ran; not a link error)
```

This validates the two parts that hadn't been exercised in workflow form — the assemble-bundle script and the prebuilt-FFI consumer path — and confirms the PR's claim that the C++ half needs no Rust toolchain (CMake took the `ZEROBUS_FFI_LIBRARY` branch; no Cargo ran).

I also verified the companion orchestrator: the four cross-repo contracts match.

Suggestion before the first real release: dispatch this workflow once on the branch to green the Zig cross-compile legs (`musl`, `darwin`, `windows-gnu`) that I couldn't run locally.

.github/workflows/release-cpp.yml
```diff
@@ -0,0 +1,196 @@
...
+        if: matrix.zigbuild && !matrix.static_only
+        env:
+          RUSTFLAGS: ${{ matrix.rustflags }}
+        run: cargo zigbuild --release -p zerobus-ffi --target ${{ matrix.target }}
+
+      - name: Build FFI static library (zigbuild, static only)
+        if: matrix.zigbuild && matrix.static_only
+        env:
+          RUSTFLAGS: ${{ matrix.rustflags }}
+        run: cargo zigbuild rustc --release -p zerobus-ffi --target ${{ matrix.target }} --crate-type staticlib
```
teodordelibasic-db: Both darwin matrix entries run `cargo zigbuild rustc ... --crate-type staticlib`, but `cargo-zigbuild` allegedly has no `rustc` subcommand, so this exits 2 before building anything. I think this identical step failed both darwin jobs in the FFI `v1.4.0` and `v1.5.0` builder runs for me with:

```text
error: unexpected argument 'rustc' found

Usage: cargo-zigbuild zigbuild [OPTIONS]
```

The working non-static-only step right above already does the right thing, and `rust/ffi/Cargo.toml` declares `crate-type = ["staticlib", "cdylib"]`, so a plain `zigbuild` still emits `libzerobus_ffi.a`:

```yaml
- name: Build FFI static library (zigbuild, static only)
  if: matrix.zigbuild && matrix.static_only
  env:
    RUSTFLAGS: ${{ matrix.rustflags }}
  run
```

.github/workflows/release-cpp.yml
```diff
@@ -0,0 +1,196 @@
...
+            cross: false
+            zigbuild: true
+            static_only: true
+          - runner: { group: databricks-protected-runner-group, labels: windows-server-latest }
+            target: x86_64-pc-windows-msvc
+            static_lib: zerobus_ffi.lib
+            artifact_dir: windows-x86_64
+            cross: false
+          - runner: { group: databricks-protected-runner-group, labels: linux-ubuntu-latest }
+            target: x86_64-pc-windows-gnu
```
teodordelibasic-db: This target cross-builds with `cargo zigbuild`, and that step failed for me in FFI runs because the mingw `dlltool` is not on the runner:

```text
error: error calling dlltool 'x86_64-w64-mingw32-dlltool': No such file or directory (os error 2)
error: could not compile `windows-sys` (lib) due to 1 previous error
```

So the simplest fix is to drop this row:

```yaml
          # remove the windows-gnu-x86_64 matrix entry; MSVC windows-x86_64 covers Windows
```

If a MinGW bundle is actually wanted, we would need to install `mingw-w64` (which provides `x86_64-w64-mingw32-dlltool`) on the runner and prove it green first.

cpp/CLAUDE.md
```diff
@@ -188,25 +188,30 @@ Public API is everything under `include/zerobus/`:
...
+Conan, vcpkg, etc.), mirroring the C FFI. The C++ wrapper is always built from
+the consumer's own compiler (its C++ ABI is not stable across compiler/stdlib),
+so we do **not** ship a prebuilt `libzerobus_cpp`. What we ship prebuilt is the
+Rust C FFI archive (`libzerobus_ffi.a`) — a stable C ABI, and the one piece a
+consumer can't build without a Rust toolchain. A consumer then either builds
+everything from source (needs Rust) via `add_subdirectory` / `FetchContent` /
+`find_package`, or points `-DZEROBUS_FFI_LIBRARY=` at the matching prebuilt FFI
+archive from a release and builds only the C++ (no Rust needed).
+
+- Tag: `cpp/v<semver>` → `release-cpp.yml` builds the per-platform bundles (the
```
teodordelibasic-db: Wording might be confusing and implicate a tag automatically does all of this. Maybe we should split it into two steps:

```text
- Create the cpp/v<semver> tag.
- Manually dispatch the release-zerobus-sdk-cpp.yml orchestrator (in
  secure-public-registry-releases-eng) with the tag as `ref`. It invokes
  release-cpp.yml in this repo, scans the artifacts, and cuts the GitHub
  Release with notes from cpp/CHANGELOG.md.
```

.github/workflows/release-cpp.yml
zlata-stefanovic-db: Thank you!

zlata-stefanovic-db: Good catch - you were right it was broken, but the fix went a different way than plain zigbuild. Tested all three locally (Zig 0.13.0, cross-compiling from Linux):

- plain build: cargo zigbuild --release -> unable to find framework 'Security'
- space form (original): cargo zigbuild rustc ... --crate-type staticlib -> unexpected argument 'rustc'
- hyphenated binary: cargo-zigbuild rustc ... --crate-type staticlib -> builds + links a valid .a

The unexpected argument 'rustc' you hit is just an invocation-form bug: cargo-zigbuild does have a rustc subcommand, but only when you call the binary directly. The cargo zigbuild rustc (space) form routes rustc as a positional into the zigbuild subcommand, which rejects it - so "cargo zigbuild" -> "cargo-zigbuild" is the actual fix. Separately, plain zigbuild can't work on darwin because the manifest's cdylib link needs the macOS Security/CoreFoundation frameworks (absent when cross-compiling from Linux); --crate-type staticlib skips that. Plain zigbuild is fine for musl since there are no frameworks.

So static_only was necessary, just with the wrong command. Fixed - and I verified the corrected command with a clean from-scratch build (cargo clean -p zerobus-ffi first) of both darwin arches: real compile + link, ~76 MB valid ar archive each, both green.

Heads up: release-ffi.yml still has the same broken cargo zigbuild rustc on its darwin legs - same one-char fix should land there too. This PR or separate?

zlata-stefanovic-db:
> Reviewed and exercised locally. The design (ship the C++ wrapper as source + a prebuilt per-platform Rust C FFI archive with a stable C ABI) is sound and matches the `ZEROBUS_FFI_LIBRARY` contract in `cpp/cmake/BuildRustFfi.cmake`.
> 
> I couldn't run the full matrix (no access to the protected runners, and the Zig cross-compile legs need Zig), but I replayed the native `linux-x86_64` leg end-to-end.
> 
> ```shell
> # 1. Build the FFI archive (exact workflow command for the linux-x86_64 leg)
> cd rust
> cargo build --release -p zerobus-ffi --target x86_64-unknown-linux-gnu
> #    -> Finished `release` in ~40s; target/x86_64-unknown-linux-gnu/release/libzerobus_ffi.a
> 
> # 2. Replay the "Assemble C++ SDK bundle" step (matrix values inlined,
> #    working-directory: rust, GITHUB_WORKSPACE=repo root)
> #    -> 41 files, correct cpp/ + lib/ layout; libzerobus_ffi.a + zerobus.h under lib/
> 
> # 3. Build the C++ wrapper from the bundle exactly as the generated README instructs
> cd artifacts/linux-x86_64
> cmake -S cpp -B build \
>   -DZEROBUS_FFI_LIBRARY="$PWD/lib/libzerobus_ffi.a" \
>   -DZEROBUS_FFI_HEADER_DIR="$PWD/lib"
> #    -> "-- Zerobus: using prebuilt FFI library …/lib/libzerobus_ffi.a" (no Cargo invoked)
> 
> cmake --build build -j
> #    -> libzerobus_cpp.a + all 4 examples (json/proto single+batch) linked
> 
> # 4. Runtime sanity
> ./build/examples/json_single
> #    -> "error: environment variable ZEROBUS_SERVER_ENDPOINT is not set" (ran; not a link error)
> ```
> 
> This validates the two parts that hadn't been exercised in workflow form — the assemble-bundle script and the prebuilt-FFI consumer path — and confirms the PR's claim that the C++ half needs no Rust toolchain (CMake took the `ZEROBUS_FFI_LIBRARY` branch; no Cargo ran).
> 
> I also verified the companion orchestrator: the four cross-repo contracts match.
> 
> Suggestion before the first real release: dispatch this workflow once on the branch to green the Zig cross-compile legs (`musl`, `darwin`, `windows-gnu`) that I couldn't run locally.

Thanks @irinatomic-db ! I picked up your suggestion about the Zig cross-compile legs and exercised them locally (Zig 0.13.0, cross-compiling from Linux x86_64, running each leg's exact workflow command from scratch). All the buildable legs produce a valid libzerobus_ffi.a:

- x86_64-unknown-linux-gnu - cargo build --release
- aarch64-unknown-linux-gnu - cargo build --release + aarch64-linux-gnu-gcc cross-linker
- x86_64-unknown-linux-musl - cargo zigbuild --release
- aarch64-unknown-linux-musl - cargo zigbuild --release
- x86_64-apple-darwin - cargo-zigbuild rustc ... --crate-type staticlib
- aarch64-apple-darwin - cargo-zigbuild rustc ... --crate-type staticlib

Two things came out of this:

- The darwin legs needed a fix - plain zigbuild fails there because the manifest's cdylib link needs the macOS Security/CoreFoundation frameworks (absent when cross-compiling from Linux). Building only the staticlib skips that. (More detail on Teodor's thread.)
- windows-gnu was dropped - its cross-build fails on the missing mingw dlltool; MSVC covers Windows.

The only leg I couldn't run here is x86_64-pc-windows-msvc, since it builds natively on a Windows runner. So your "dispatch once before the first real release" suggestion still stands as the way to green that last leg on the actual runners - I'll do a workflow dispatch (or the orchestrator dry-run) once this is on main.

.github/workflows/release-cpp.yml
zlata-stefanovic-db: Dropped this

cpp/CLAUDE.md
zlata-stefanovic-db: Changed the wording

irinatomic-db (approved):

.github/workflows/release-cpp.yml
zlata-stefanovic-db: Created a PR for this  #568 @teodordelibasic-db

PR #558, opened 2026-07-21
[ArrowFlight] Make recovery handoff atomic with ingest

nikolaobradovic-db:

Two recovery-handoff races let a concurrent ingest_batch mishandle a batch it had already queued into pending_batches:
* Spurious ingest error: the supervisor set is_paused = true and detached batch_tx as two steps without ingest_mutex, so an in-flight ingest could see is_paused == false, then batch_tx == None, and return "stream sender is closed" for a batch recovery would replay. Extracted pause_and_detach_sender, which does both under ingest_mutex; ingest now either sends before the transition or observes is_paused == true and buffers (Ok).
* Dropped-on-replay: reconnect reset the recovery counters before rebuilding the pending ranges, so a record ingested in that window got a stale range and was skipped by replay as already acked. The reset and range rebuild are now atomic (under the pending_batches lock); the watermark uses Acquire/Release.

## How is this tested?

* New deterministic tests (both verified red first): pause_and_detach_waits_for_in_flight_ingest (unit) and test_ingest_during_reconnect_window_is_not_dropped (integration, uses a test-only barrier to park reconnect in the rebuild window).
* Added a test-only test-hooks Cargo feature for the barriers (zero footprint; not enabled by FFI).

### DIFF
--- a/rust/sdk/src/arrow_stream.rs
+++ b/rust/sdk/src/arrow_stream.rs
@@ -20,6 +20,8 @@ use arrow_ipc::writer::IpcWriteOptions;
 use bytes::Bytes;
 use futures::{Stream, StreamExt};
 use tokio::sync::{mpsc, watch, Mutex, OwnedSemaphorePermit, Semaphore};
+#[cfg(feature = "test-hooks")]
+use tokio::sync::Notify;
 use tokio::time::{sleep, Duration};
 use tokio_retry::strategy::FixedInterval;
 use tokio_retry::RetryIf;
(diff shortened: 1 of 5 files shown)


rust/sdk/src/arrow_stream.rs
teodordelibasic-db, line 1: IMO this file could benefit from splitting into a few separate ones based on responsibility. I skipped it when splitting the gRPC one and meant to do it at some point.

rust/sdk/src/arrow_stream.rs
```diff
@@ -991,30 +1042,31 @@ impl ZerobusArrowStream {
...
-                             possible orphaned buffered batch from pause-gate handoff race",
-                            expected_start, pb.start_record, pb.offset_id,
-                        );
-                        expected_start = pb.end_record;
-                    }
+            // Reset counters together with the range install, before any send.
+            cumulative_records_sent.store(new_cumulative, Ordering::Relaxed);
+            last_acked_records.store(0, Ordering::Release);
+
+            // Debug-only: rebuilt ranges must be contiguous from 0. Guards a rebuild
```
teodordelibasic-db: nit for a slightly overstating comment:

The rebuild loop assigns each entry a contiguous range from `0`, so the `debug_assert_eq!` right after only re-checks what the loop just built. It can't catch the "orphaned buffered batch from pause-gate handoff race" the comment claims, since `pending` is fully drained and rebuilt under the same lock and any buffered batch is included in the rebuild. More accurate would be:

```rust
// Debug-only self-check: the rebuild loop above assigns contiguous ranges from 0,
// so this only trips if that loop is later changed to break contiguity.
#[cfg(debug_assertions)]
{
    let mut expected_start: u64 = 0;
    ...
}
```

rust/sdk/src/arrow_stream.rs
```diff
@@ -1745,6 +1824,34 @@ impl ZerobusArrowStream {
         self.is_closed.load(Ordering::Relaxed)
     }
 
+    /// Test-only: arms the reconnect rebuild barrier. The next `reconnect` pauses after
```
teodordelibasic-db, line 1810: The advertised use of driving a concurrent `close()` while reconnect is parked is not utilized by any test.

Two potential cases worth adding, since the seams are already built for them:

- An end-to-end test that runs a real `ingest_batch` concurrent with the pause/detach transition and asserts it returns `Ok` and the batch is later acked or replayed, not the spurious closed error.
- The `close()`-during-window test the barrier doc describes, asserting pending batches end up in the failed set and no panic or hang beyond the flush timeout.

teodordelibasic-db (dismissed):
LGTM

rust/sdk/src/arrow_stream.rs
nikolaobradovic-db: This assertion is obsolete and will be dropped.

rust/sdk/src/arrow_stream.rs
nikolaobradovic-db: Adding the second one. The first one is mostly covered by the unit test pause_and_detach_waits_for_in_flight_ingest.

rust/sdk/src/arrow_stream.rs
nikolaobradovic-db: For better times, maybe after SDK hardening is complete.

teodordelibasic-db (approved):
LGTM

PR #561, opened 2026-07-21
[ArrowFlight] On reconnect, fail immediately for non retryable errors

flaviofcruz:

When the supervisor needs to reconnect and the error is non-retryable, then close the stream immediately rather than re-trying hopelessly. This will be necessary for when we introduce schema errors.

Note that auth rejections just force a new token, so won't cause the stream to be closed immediately.

## How is this tested?

Added test.


nikolaobradovic-db:
Hey @flaviofcruz I am handling this in https://github.com/databricks/zerobus-sdk/pull/562 plus a bit more

PR #562, opened 2026-07-21
[ArrowFlight] Harden terminal/close/recovery handling

nikolaobradovic-db:

Makes the Arrow stream's terminal, close, waiter, and recovery paths correct under concurrency and surface accurate errors.

- Sliced, race-free retrieval: the terminal drain and get_unacked_batches() slice a partially-acked auto-chunked batch to its un-acknowledged suffix (no re-sending the durable prefix), and return a consistent, idempotent snapshot. finalize_closed publishes is_closed and drains under ingest_mutex, so an ingest_batch racing the terminal drain is included, never omitted-then-revealed.
- close() teardown: flush → reap supervisor (join in place) → detach sender → finalize. Detaching after the reap stops an in-flight reconnect from reinstalling the sender; is_closed is set only at the end, so a cancelled close() retry re-runs teardown instead of enabling retrieval prematurely.
- Real terminal errors to waiters: every terminal path (mid-stream error, stream-end, ack-timeout) publishes the error and the supervisor re-publishes after finalization, so flush()/wait_for_offset() return the actual cause, not a generic "timed out"/"stream is closed". Finalization runs before the user HeadersProvider::invalidate() so a slow callback can't leave a failed stream open.
- Ack precedence: the waiter checks the ack watermark before is_closed and re-reads it after observing closure, so a target acked just before closure resolves Ok(()). flush() defers to the waiter instead of an early closed check.
- Reconnect cause preserved: failed reconnects carry the real error instead of a synthetic "Reconnection failed". Non-retryable non-auth failures terminate immediately; retryable ones retry to exhaustion then surface the real error; an auth rejection invalidates credentials and retries (OAuth re-mint), surfacing the original error only if retries are exhausted (a scoped flag prevents that override leaking into a later non-auth failure).
- Docs: get_unacked_batches rustdoc rewritten; its example no longer treats close() == Ok(()) as "all acked" or double-ingests; removed a tautological replay debug-assert and stale comments.

## How is this tested?

New/updated tests (plus a FailSetup mock variant and a CountingHeadersProvider) covering: sliced/idempotent retrieval and the finalize_closed-vs-ingest race; real terminal errors for all three arms (mid-stream, stream-end, genuine ack-timeout); ack-vs-close precedence and closed empty-stream flush; and the full reconnect matrix — non-auth non-retryable (terminates), non-auth retryable (exhausts), auth (invalidate+retry, 1 invalidation), auth exhaustion (2 invalidations, original error), and an auth→non-auth flag-leak guard — asserting message, retry classification, and invalidation counts.

I will also hotpatch the SDK in the universe with changes and run E2E server-side tests + performance benchmarks.


rust/sdk/src/arrow_stream.rs
```diff
@@ -1708,67 +1813,79 @@ impl ZerobusArrowStream {
             "Closing Arrow Flight stream"
         );
 
-        // Flush pending batches. Capture the result so the stream is still torn
-        // down on failure, then propagate it to the caller.
+        // Flush pending batches while the supervisor is still alive to receive acks.
+        // Capture the result so teardown still runs on failure, then propagate it.
         let flush_result = self.flush().await;
         if let Err(e) = &flush_result {
             warn!(
```
teodordelibasic-db, line 1890: nit: Should we maybe log this at `error` level?

rust/sdk/src/arrow_stream.rs
```diff
@@ -1708,67 +1813,79 @@ impl ZerobusArrowStream {
...
         }
 
-        // Mark as closed.
-        self.is_closed.store(true, Ordering::Relaxed);
+        // Reap the supervisor (abort + await) BEFORE clearing the sender, so an in-flight
+        // reconnect can't reinstall batch_tx after we clear it, and no process_acks /
+        // reconnect mutates pending_batches or last_acked_records while we drain. Join in
+        // place and only clear receiver_task once the join completes, so a close()
+        // cancelled during the await doesn't drop the handle — a retry re-joins it.
+        {
```
teodordelibasic-db, line 1901: If `close()` is cancelled between the supervisor abort and `finalize_closed`, the stream is left falsely active: `is_closed` is still `false`, the supervisor is dead, and the sender's receive side has been dropped. A later `ingest_batch` then passes the `is_closed` check, takes a permit, appends to `pending_batches`, and with recovery enabled (the default) returns `Ok(offset)` when the channel `send` fails, so the batch is silently accepted but never sent, acked, or retrievable. 

`get_unacked_batches()` also refuses, because the stream still looks active. This is reachable whenever `close()` is wrapped in `tokio::time::timeout` or a `select!` and the stream is touched again afterward, since `close(&mut self)` hands `&mut self` back once the future is dropped.

Retrying `close()` does not promptly repair it either: it starts with `flush()`, which cannot make progress with no live supervisor and can block for the full `flush_timeout` (default five minutes) before teardown finally drains the batch.

Maybe we could set a synchronized teardown flag before the abort so new ingests are rejected once teardown starts, while still letting a retried `close()` resume teardown (skipping `flush()`) after the supervisor is gone? Running teardown in a cancellation-independent task would also work.

```rust
// At the very start of teardown, before the cancellable awaits:
{
    let _guard = self.ingest_mutex.lock().await;
    self.tearing_down.store(true, Ordering::Release);
}

// In ingest_batch, after taking ingest_mutex (alongside the is_closed re-check):
if self.tearing_down.load(Ordering::Acquire) {
    return Err(ZerobusError::StreamClosedError(
        tonic::Status::internal("Stream is closing"),
    ));
}
```

rust/sdk/src/arrow_stream.rs
```diff
@@ -780,46 +813,59 @@ impl ZerobusArrowStream {
...
-                                // token on auth rejection so recovery re-mints.
+                                warn!("Supervisor: Reconnection failed: {}", e);
+                                // On an auth rejection the cached token is likely expired:
+                                // invalidate it and retry so the next attempt mints a fresh
+                                // one (auth errors are otherwise non-retryable). Non-auth
+                                // errors are carried as-is and classified normally. Either
+                                // way the real error is preserved as the final cause if
+                                // retries are ultimately exhausted.
                                 if e.is_auth_rejection() {
                                     headers_provider.invalidate().await;
```
teodordelibasic-db: `is_closed` is still `false` here and `is_paused` is `true`. If `invalidate()` hangs it stalls the whole recovery cycle: ingests keep buffering under the pause, and waiters eventually surface a generic flush timeout instead of the real auth error.

Could this reconnect arm follow the same pattern as the terminal arm and move `invalidate()` to after `finalize_closed`?

rust/sdk/src/arrow_stream.rs
teodordelibasic-db, line 1569: Instead of holding `ingest_mutex` across `sender.send(Ok(batch)).await`, could we record the pending batch and clone the sender under the lock, drop the guard, then send?

```rust
let sender = {
    let _guard = self.ingest_mutex.lock().await;
    // is_closed re-check, offset/range assignment, push to pending_batches,
    // is_paused early-return ...
    self.batch_tx.lock().await.clone()
};
// ingest_mutex released; send without holding it
if let Some(sender) = sender {
    if let Err(e) = sender.send(Ok(batch)).await { /* recovery / error path */ }
}
```

rust/sdk/src/arrow_stream.rs
```diff
@@ -1700,6 +1798,13 @@ impl ZerobusArrowStream {
     #[instrument(level = "debug", skip_all, fields(table_name = %self.table_properties.table_name))]
     pub async fn close(&mut self) -> ZerobusResult<()> {
         if self.is_closed.load(Ordering::Relaxed) {
```
teodordelibasic-db: The two already-closed outcomes seem inconsistent, not sure if intentional.

After a background terminal failure, every later `close()` returns the stored `server_error`, which is correct.

But after a `close()` that failed on a flush timeout, teardown runs and `finalize_closed` sets `is_closed`, yet the timeout is never published to `server_error_rx`, so a second `close()` takes the early-return branch and returns `Ok(())` even though unacked batches are sitting in the failed set.

rust/sdk/src/arrow_stream.rs
nikolaobradovic-db: No, but I can set a timeout to invalidate.

rust/sdk/src/arrow_stream.rs
nikolaobradovic-db: Warning is fine, we propagate the error back from all paths now.

rust/sdk/src/arrow_stream.rs
teodordelibasic-db: nit: Also returns a stored terminal error when the stream was already closed by a background failure.

rust/sdk/src/arrow_stream.rs
nikolaobradovic-db: Cancellation is tricky to handle, but it should be good now with an additional flag.

rust/sdk/src/arrow_stream.rs
nikolaobradovic-db: Handled this one and also termination case.

rust/sdk/src/arrow_stream.rs
nikolaobradovic-db: No, it's unsafe, it could mess up batch ordering and open a whole new can of worms in concurrency with recovery.

rust/sdk/src/arrow_stream.rs
nikolaobradovic-db: Resolved this with cancellation safe close.

rust/sdk/src/arrow_stream.rs
nikolaobradovic-db: Ill make a comment update pass and push it, some of them remained outdated and inconsistent.

nikolaobradovic-db: Done.

rust/tests/src/arrow_tests.rs
```diff
@@ -258,68 +503,832 @@ mod arrow_flight_tests {
...
+
+            let mut stream = sdk
+                .stream_builder()
+                .table(TABLE_NAME)
+                .headers_provider(Arc::new(TestHeadersProvider::default()))
+                .arrow(schema.clone())
+                .build_arrow()
+                .await?;
+
+            let (reached, _proceed) = stream.arm_close_finalize_barrier().await;
```
teodordelibasic-db, line 566: The cancellation test reaches the finalization barrier with no pending batch, so its flush succeeds and `close_flush_error` remains empty. Separately, `test_close_propagates_flush_error` calls `close()` only once. A regression could therefore lose the stored timeout during cancelled teardown, or return `Ok(())` from a later close, while both tests remain green.

Could we cover the combined state transition with an unacknowledged batch, a short flush timeout, cancellation at this barrier, and these assertions after resuming?

```rust
let resumed = stream.close().await.expect_err("stored flush error");
assert!(resumed.to_string().contains("Flush timed out"));

let repeated = stream.close().await.expect_err("idempotent flush error");
assert!(repeated.to_string().contains("Flush timed out"));
assert_eq!(stream.get_unacked_batches().await?.len(), 1);
```

teodordelibasic-db (dismissed):
LGTM, just one potential test gap flagged by LLM.

rust/tests/src/arrow_tests.rs
nikolaobradovic-db: Fixed.

teodordelibasic-db (dismissed):

teodordelibasic-db (approved):

PR #566, opened 2026-07-22
[PureGo] Fix OAuth redirect leak, expired-on-arrival token, and cancel-cause classification

### LINKED ISSUE #563: PureGo: OAuth mint follows redirects, can leak client credentials over https→http
The OAuth token mint request in `purego/internal/auth/oauth.go` sends the client ID and secret to the server over `https` (the constructor enforces `https` for exactly this reason). However, the request uses the default HTTP client, which automatically follows redirects.

Go's redirect handling only checks that the redirect target has the same host before re-attaching credentials — it does **not** check that the target is still `https`. So a redirect from `https://host/...` to `http://host/...` causes Go to re-send the client ID and secret **in the clear**, where anyone on the network can read them. The `https` guarantee from the constructor is lost the moment a redirect is followed.

### Suggested fix

A token endpoint has no legitimate reason to redirect, so the safest fix is to not follow redirects on this request. Copy the client first to avoid mutating a caller-supplied client:

```go
client := *p.client
client.CheckRedirect = func(*http.Request, []*http.Request) error {
        return http.ErrUseLastResponse
}
resp, err := client.Do(req)
```

With `ErrUseLastResponse`, a redirect is returned as an ordinary non-success response and the mint fails through the existing error path instead of quietly following the redirect.

### Testing

Add tests with a same-host `https`-to-`http` redirect and a cross-host redirect to lock this in.

### Context

The Rust SDK is not exposed to this even though it also follows redirects with credentials attached — Rust's HTTP library drops the credentials on this kind of redirect, while Go's keeps them. This is a case where PureGo behaves less safely than Rust rather than matching it.

---
_Filed from review feedback on [PR #496](https://github.com/databricks/zerobus-sdk/pull/496) (comment on `purego/internal/auth/oauth.go:318`)._

### LINKED ISSUE #564: PureGo: getOrFetch returns nil error with an expired token on cold-miss expired-on-arrival
In `purego/internal/auth/token_cache.go`, when a freshly minted token is already past its receipt-anchored TTL and there is no still-valid cached entry to fall back on, `getOrFetch` falls into the `!keepExisting` case, clears `entry.cached`, and returns the expired `token` with a **nil error**.

`getOrFetch` documents that it returns a valid token, so the caller receives a nil error but an unusable token. The failure then surfaces later as a server-side auth rejection rather than here, where the cause is known.

### Suggested fix

Return an error for the cold-miss expired-on-arrival case, while keeping the existing behavior that serves a still-valid cached token when an early refresh produces an expired replacement:

```go
case alreadyExpired && keepExisting:
        token = entry.cached.value
case alreadyExpired: // expired on arrival, nothing valid to serve
        entry.cached = nil
        err := &TokenError{msg: "minted token already expired on arrival", retryable: true}
        flight.token, flight.err = "", err
        close(flight.done)
        entry.mu.Unlock()
        return "", err, false
case !keepExisting:
        entry.cached = nil
```

### Notes

This is a narrow branch in practice: it needs a very short `expires_in` plus slow post-receipt work to consume the whole TTL, so real UC tokens rarely hit it.

### Testing

`TestTokenCacheDoesNotCacheAlreadyExpiredToken` currently asserts the token is returned; it would need to assert the error instead.

---
_Filed from review feedback on [PR #496](https://github.com/databricks/zerobus-sdk/pull/496) (comment on `purego/internal/auth/token_cache.go:320`)._

### LINKED ISSUE #565: PureGo: isCallerCancellation misclassifies WithCancelCause cancellations as retryable
In `purego/internal/auth/token_cache.go`, `isCallerCancellation` needs to distinguish a real caller cancellation (which must be honored by failing) from our own internal open-budget timeout (which is transient, so we can fall back to a still-valid cached token).

It currently decides by matching `context.Cause(ctx)` against Go's two standard values. But a caller who cancels via `context.WithCancelCause` supplies a custom cause, so we fail to recognize their cancellation, treat it as retryable, and can serve a cached token instead of stopping as they asked.

### Impact

Mild: the caller still gets a valid token, not a hang or a wrong error, and nothing inside the SDK cancels this way, so it's only reachable when an external caller uses `WithCancelCause`. Still, cancelling with a reason should behave the same as cancelling without one.

### Suggested fix

The logic is currently "recognize the caller, treat everything else as our budget." It is more robust flipped: recognize only our own budget, and treat everything else as a caller cancellation.

```go
func isCallerCancellation(ctx context.Context, err error) bool {
        if !isContextError(err) || ctx.Err() == nil {
                return false
        }
        // Our own open budget is the one non-caller cause; anything else is the caller.
        return context.Cause(ctx) != errHeadersBudgetExceeded
}
```

### Snag

`auth` deliberately does not import `transport`, where `errHeadersBudgetExceeded` lives, so this needs that sentinel shared across the two packages rather than a one-line edit.

---
_Filed from review feedback on [PR #496](https://github.com/databricks/zerobus-sdk/pull/496) (comment on `purego/internal/auth/token_cache.go:350`)._

zlata-stefanovic-db:
Fixes three review-feedback issues in the PureGo auth layer.

### 563 — OAuth mint follows redirects, can leak client credentials over https→http
`fetchToken` used the default HTTP client, which follows redirects. Go re-attaches credentials on a same-host redirect without re-checking the scheme, so an `https`→`http` redirect would re-send the client ID and secret in the clear. The mint request now copies the client and sets `CheckRedirect` to `http.ErrUseLastResponse`, so a redirect is surfaced as a non-success response and fails through `classifyHTTPError` instead of being followed. Copying the client avoids mutating a caller-supplied one (`WithHTTPClient`).

### 564 — `getOrFetch` returned a nil error with an expired token on cold-miss expired-on-arrival
A freshly minted token whose receipt-anchored TTL has already elapsed, with no still-valid cached entry to fall back on, previously fell into the `!keepExisting` case and returned the expired token with a nil error — surfacing later as an opaque server-side auth rejection. A dedicated switch case now fails fast with a retryable `TokenError`, while the existing behavior of serving a still-valid cached token on an early refresh is preserved.

### 565 — `isCallerCancellation` misclassified `WithCancelCause` cancellations as retryable
The check recognized a caller cancellation only by matching `context.Cause` against Go's two standard values, so a caller cancelling via `context.WithCancelCause` (custom cause) was treated as retryable and could be served a cached token instead of stopping. The logic is flipped to recognize only the SDK's own header-resolution budget as the non-caller cause and treat everything else as a caller cancellation. The budget sentinel was moved into a new dependency-free leaf package `internal/authctx` so `auth` and `transport` can share the same value without importing each other.

### Testing
- Added same-host and cross-host redirect tests (563).
- Updated `TestTokenCacheDoesNotCacheAlreadyExpiredToken` to assert the retryable error; the sibling cached-fallback test is unchanged (564).
- Updated `TestIsCallerCancellationDistinguishesBudgetFromCaller` to use the shared sentinel and added a `WithCancelCause` case (565).
- `gofmt`, `go vet`, and the `auth` + `transport` suites pass, including under `-race`.


purego/internal/auth/oauth.go
teodordelibasic-db, line 558: The internal header budget is still classified as nonretryable on HTTP/1. When the `WithTimeoutCause` budget fires, Go returns a `url.Error` wrapping `authctx.ErrHeadersBudgetExceeded` directly. That chain does not match `context.Canceled` or `context.DeadlineExceeded`, `Timeout()` is false, and it contains no `net.OpError`. We therefore skip the `isContextError` branch and return `false`, so a refresh fails even when a still-valid cached token should be served.

Could we classify a completed request context before inspecting the transport error, while continuing to treat a client timeout as retryable when the passed context is still live? For example:

```go
func isRetryableTransportError(ctx context.Context, err error) bool {
        if ctx.Err() != nil {
                return context.Cause(ctx) == authctx.ErrHeadersBudgetExceeded
        }
        if isContextError(err) {
                return true
        }
        // Existing network error checks.
}
```

A focused HTTP/1 test could seed a refreshable cache entry, expire a `WithTimeoutCause` header budget against a hanging token endpoint, and assert that the cached token is returned.

purego/internal/auth/token_cache.go
teodordelibasic-db: A live waiter can inherit another caller's custom cancellation. On HTTP/1, a `WithCancelCause` cancellation reaches the flight as a `TokenError` wrapping a `url.Error` wrapping the custom cause. Since that chain does not match `context.Canceled` or `context.DeadlineExceeded`, `isContextError(flight.err)` is false and the waiter returns the leader's error instead of retrying with its own live context.

Could we publish the cancellation provenance on `tokenFlight` while the leader context is still available, then use that flag instead of reconstructing provenance from the final error shape? For example:

```go
type tokenFlight struct {
        done           chan struct{}
        token          string
        err            error
        leaderCanceled bool
}

flight.leaderCanceled = err != nil &&
        ctx.Err() != nil &&
        context.Cause(ctx) != authctx.ErrHeadersBudgetExceeded

if flight.leaderCanceled && ctx.Err() == nil {
        return "", nil, true
}
```

A concurrent HTTP/1 test with a custom-cause leader and a live waiter would lock in the isolation promised by `getOrFetch`.

teodordelibasic-db (commented):
LGTM.

purego/internal/auth/oauth.go
zlata-stefanovic-db: Confirmed over HTTP/1.1 - the transport wraps ErrHeadersBudgetExceeded into the url.Error, so the shape checks all miss it. Fixed as suggested: classify ctx first (context.Cause), inspect the error only when ctx is still live. Added an HTTP/1.1 test that trips the budget and asserts the cached token is served; verified red on the old logic.

purego/internal/auth/token_cache.go
zlata-stefanovic-db: Agreed. Went with publishing provenance on the flight: a leaderCanceled flag stamped from the live context, gated resErr != nil && !isRetryable(resErr) && isCallerCancellation(ctx) so a cached fallback or retryable fault is still shared. Added HTTP/1.1 tests for the custom-cause waiter case plus both gate conditions; all red before / green after.

purego/internal/auth/token_cache.go
```diff
@@ -290,9 +293,13 @@ func (c *tokenCache) tryGetOrFetch(
...
 		// Publish to waiters: writes before close(done) happen-before <-done.
-		// A caller-cancellation error published here lets a live-ctx waiter
-		// re-attempt instead of inheriting a cancellation it never requested.
+		// Stamp leaderCanceled only when the leader's own caller-cancel produced a
+		// non-retryable failure, so a live-ctx waiter re-attempts rather than
+		// inherits it. A successful cached fallback (resErr == nil) or a retryable
+		// outcome is still shared: those don't need re-minting, and the context can
+		// only be inspected for provenance, not blamed for the error itself.
 		flight.token, flight.err = token, resErr
+		flight.leaderCanceled = resErr != nil && !isRetryable(resErr) && isCallerCancellation(ctx)
```
teodordelibasic-db: Cancellation provenance is sampled too late here. A definitive `401` can already have been classified while `ctx` is live, then the synchronous logger can block and the caller can cancel before `mint` returns. This expression marks that completed server failure as `leaderCanceled`, so live waiters mint again instead of sharing the definitive response and the single-flight guarantee is broken.

Could we capture causality immediately after `fetchToken` returns, before logging, and carry that private bit on `TokenError`?

```go
type TokenError struct {
        msg              string
        retryable        bool
        cause            error
        callerCanceled   bool
        ResponseBody     string
}

var tokenErr *TokenError
if errors.As(err, &tokenErr) && isCallerCancellation(ctx) {
        cause := context.Cause(ctx)
        tokenErr.callerCanceled =
                errors.Is(err, cause) || errors.Is(err, ctx.Err())
}

// When publishing the flight:
var resolvedTokenErr *TokenError
flight.leaderCanceled =
        errors.As(resErr, &resolvedTokenErr) && resolvedTokenErr.callerCanceled
```

The existing `WithLogger` seam supports a deterministic regression test. A blocking `slog.Handler` can pause after a real `401` is classified, while a waiter context signals through its `Done` method that it entered the flight wait. Cancel the leader, release the logger, and assert that both callers receive the `401` with one HTTP request.

purego/internal/auth/token_cache.go
teodordelibasic-db, line 216: Expired-on-arrival validation is still bypassed when caching is disabled and by `FetchToken`. Both paths return `fetched.token` directly after the synchronous logger runs, so logging can consume a short receipt-anchored TTL and the caller receives an already-expired token with a nil error. This contradicts the valid-token contract on `Token` and the guaranteed-fresh contract on `FetchToken`.

Could we share the receipt-anchored expiry calculation and apply the same retryable error after `mint` in both uncached paths, while retaining cached fallback only in `tryGetOrFetch`?

```go
func (f fetchedToken) expiredAt(now time.Time) bool {
        if f.expiresIn == nil || *f.expiresIn <= 0 {
                return false
        }
        receivedAt := f.receivedAt
        if receivedAt.IsZero() {
                receivedAt = now
        }
        return !now.Before(receivedAt.Add(*f.expiresIn))
}

if fetched.expiredAt(time.Now()) {
        return "", &TokenError{
                msg:       "minted token already expired on arrival",
                retryable: true,
        }
}
```

The cache-disabled unit test can return a `fetchedToken` with a past `receivedAt`. The direct-fetch test can use a one-second `expires_in` and a blocking `WithLogger` handler that releases after the TTL, then assert an empty token and retryable `TokenError` from both paths.

purego/internal/auth/token_cache_test.go
teodordelibasic-db: Could we replace the scheduling delay with a waiter context whose `Done` method signals when `tryGetOrFetch` evaluates the flight-wait `select`?

```go
type observedDoneContext struct {
        context.Context
        observed chan struct{}
        once     sync.Once
}

func (c *observedDoneContext) Done() <-chan struct{} {
        c.once.Do(func() { close(c.observed) })
        return c.Context.Done()
}

waiterCtx := &observedDoneContext{
        Context:  context.Background(),
        observed: make(chan struct{}),
}
go callToken(waiterCtx)
<-waiterCtx.observed
cancelLeader()
```

purego/internal/auth/token_cache.go
zlata-stefanovic-db: Thank you, fixed as suggested - we capture causality immediately after fetchToken returns

purego/internal/auth/token_cache.go
zlata-stefanovic-db: Done, added a shared receipt-anchored expiry calculation and now apply it to the cached, cache-disabled, and direct FetchToken paths.

purego/internal/auth/token_cache_test.go
zlata-stefanovic-db: Yes, done as suggested - applied to the single-flight tests too

purego/internal/auth/oauth.go
teodordelibasic-db, line 365: Response-body timeouts are hard-coded as non-retryable here. `http.Client.Timeout` and the SDK header budget both cover reading the response body, not only receiving its headers. If UC sends a `200` response with partial JSON and then stalls, `Decode` returns a timeout and a proactive refresh fails even when a still-valid cached token is available.

Could we classify the decode error with the same context-aware transport logic used for `client.Do`? A malformed JSON response stays non-retryable because it does not match any timeout or network error.

```go
return fetchedToken{}, &TokenError{
        msg:       fmt.Sprintf("parse token response: %v", err),
        retryable: isRetryableTransportError(ctx, err),
        cause:     err,
}
```

A focused table test can flush a `200` response with partial JSON, stall the body, and cover both `http.Client.Timeout` and `authctx.ErrHeadersBudgetExceeded`. Each case should serve the refreshable cached token.

purego/internal/auth/oauth.go
```diff
@@ -508,13 +537,20 @@ func isHTTPSuccess(code int) bool { return code >= 200 && code < 300 }
...
 // failures, and request/budget timeouts qualify; a caller-owned cancel or
-// deadline does not — that is the caller's signal to stop, not a server fault.
-// isCallerCancellation tells the two apart when the error is a context error.
+// deadline does not.
+//
+// Provenance comes from the context, not the error's shape: on HTTP/1 the
+// transport wraps a WithTimeoutCause/WithCancelCause cause into the error, so
+// isContextError alone would miss the SDK's own budget. Check the context first.
 func isRetryableTransportError(ctx context.Context, err error) bool {
+	if ctx.Err() != nil {
```
teodordelibasic-db, line 546: An expired SDK budget can hide an unrelated permanent transport failure. Once `ctx.Err()` is non-nil, this branch ignores `err` and marks the result retryable solely from `context.Cause(ctx)`. A caller-provided `RoundTripper` that returns a permanent configuration error as the budget expires is therefore treated as transient. During proactive refresh, `Token` returns the cached token with no error even though cache fallback is documented for transient failures.

Could we attribute the failure to the SDK budget only when `err` wraps either the custom budget cause used by HTTP/1 or the standard context error used by HTTP/2, then classify unrelated errors normally?

```go
if ctx.Err() != nil {
        cause := context.Cause(ctx)
        if cause != authctx.ErrHeadersBudgetExceeded {
                return false
        }
        if errors.Is(err, cause) || errors.Is(err, ctx.Err()) {
                return true
        }
}
```

`WithHTTPClient` gives us a deterministic regression seam. A custom `RoundTripper` can wait for `req.Context().Done()`, return a permanent sentinel, and verify that a refreshable cache entry does not suppress that error.

purego/internal/auth/oauth.go
zlata-stefanovic-db: fixed decode errors to use the transport retry classifier instead of always non-retryable. So decode timeouts now count as transient and we can fall back to cached token.

purego/internal/auth/oauth.go
zlata-stefanovic-db: Fixed the ctx.Err()!=nil issue: we only mark retryable for the SDK budget timeout when the actual error matches that cause/context error

teodordelibasic-db (approved):

PR #569, opened 2026-07-22
[Rust] Honor proxy settings in Arrow Flight

### LINKED ISSUE #388: [Rust] Proxied gRPC channel ignores ClientTlsConfig → InvalidContentType on every TLS endpoint
### SDK

Rust

### SDK Version

rust v2.2.2 and Go v1.2.0

### Operating System

Linux 6.17.0-29-generic (Ubuntu)

### Description

When the `zerobus-sdk` Rust core builds a gRPC channel through a proxy (`HTTPS_PROXY` or `GRPC_PROXY`), the endpoint's `ClientTlsConfig` is **not** applied on top of the tunneled stream. The HTTP `CONNECT` to the Zerobus host succeeds, but the client then speaks **plaintext HTTP/2** to a TLS-only server through the tunnel, and the server's response is rejected as:

```
ZerobusError: Failed to create stream: code: 'The service is currently unavailable',
message: "received corrupt message of type InvalidContentType",
source: tonic::transport::Error(Transport, ConnectError(Custom { kind: InvalidData,
        error: InvalidMessage(InvalidContentType) }))
```

A **direct** (non-proxied) connection with the same credentials connects fine. Proxying — the whole point of the feature — is non-functional for real (TLS) Databricks endpoints. Because every language SDK wraps this Rust core, all of them are affected.

### Reproduction

1. Create a Zerobus Ingest client with valid creds (`workspaceUrl`, `zerobusServerUrl`, catalog, schema, table, `clientId`, `clientSecret`).
2. Set `GRPC_PROXY` (or `HTTPS_PROXY`) to `http://<any-working-http-proxy>:<port>` — a corporate Squid, or a trivial CONNECT proxy that tunnels to the internet.
3. Attempt to open the stream.
4. Stream creation fails with `InvalidContentType`. The proxy logs show `CONNECT <zerobusServerUrl>:443` was received — i.e. routing works; the TLS-over-tunnel layer is what fails.

### Matrix (all proxied combinations fail identically; only direct works)

| Connection | Result |
|---|---|
| Direct (no proxy) | ✅ connects |
| `GRPC_PROXY` via mock CONNECT proxy | ❌ `InvalidContentType` |
| `GRPC_PROXY` via real Squid | ❌ `InvalidContentType` |
| `HTTPS_PROXY` via mock CONNECT proxy | ❌ `InvalidContentType` |
| `HTTPS_PROXY` via real Squid | ❌ `InvalidContentType` |
| `no_proxy` matches host (→ bypass → direct) | ✅ connects |

### Expected behavior

When a proxy is configured (HTTPS_PROXY or GRPC_PROXY), the SDK should establish a gRPC channel that is functionally equivalent to the direct path:

  1. CONNECT tunnel is opened through the proxy to <zerobusServerUrl>:443 — (already works).
  2. TLS handshake runs on top of the tunneled stream, using the endpoint's ClientTlsConfig (SecureTlsConfig / native roots), with SNI and certificate verification against the Zerobus host.
  3. HTTP/2 + gRPC are spoken over the TLS session, not over the raw tunnel.
  4. Stream creation succeeds and the client reports credentials verified / Connected, identically to the no-proxy case.
  5. No InvalidContentType (or any other TLS/plaintext-mismatch error) is produced when the server is TLS-only.

  Equivalently: the connect_with_connector_lazy(proxy_connector) path must honor ClientTlsConfig the same way connect_lazy() does today. Proxied and direct connections should be indistinguishable to the
  Zerobus server.

### Is it a regression?

No

### Logs

```shell

```

## Impact

- Any user behind a forward proxy (the common enterprise case, e.g. corporate Squid via `HTTPS_PROXY`) **cannot use** the Zerobus SDK. Only direct egress works.
- Affects every proxy type and every language SDK that depends on the shared Rust core.

## Suggested fix

Apply the endpoint's TLS to the proxied connection — e.g. wrap the proxy connector so tonic's TLS layer runs on top of the tunneled stream, or build the channel so `ClientTlsConfig` is honored on the `connect_with_connector` path.

Add a regression test that uses `SecureTlsConfig` against a **real TLS server through a CONNECT proxy** (not `NoTlsConfig` / plaintext), so this path stops being silently untested.

## Workaround

None for proxied egress. Only direct connectivity, or a `no_proxy` bypass for the Zerobus host, works.

teodordelibasic-db:

Arrow Flight now uses the SDK's shared proxy resolution, so environment proxies, no-proxy bypasses, and custom `connector_factory` policies apply consistently to initial connections, recovery, and explicit stream recreation.

The CONNECT tunnel remains unwrapped so tonic applies target TLS once, while HTTPS proxy hops use native roots. The public API is unchanged; connector documentation and the upcoming changelog are updated.

## How is this tested?

Added Rust integration tests for Arrow proxy routing and no-proxy bypass, target TLS through CONNECT, environment/custom proxy reuse during recovery, and custom factory reuse during explicit stream recreation.


rust/sdk/src/proxy.rs
```diff
@@ -96,9 +123,9 @@ fn read_first_env(names: &[&str]) -> Option<String> {
...
 ///
-/// Uses `from_proxy` so `https://` proxy URLs work (TLS handshake on the
-/// client→proxy hop using the system trust store). The CONNECT tunnel still
-/// carries raw TCP; tonic applies TLS to the target on top.
+/// The underlying connector handles TLS for `https://` proxy URLs using the
+/// system trust store. The CONNECT tunnel remains raw so tonic applies any
+/// target TLS exactly once.
 pub(crate) fn create_proxy_connector() -> Option<ProxiedConnector> {
     let proxy_url = read_first_env(PROXY_ENV_VARS)?;
     info!("Using HTTP proxy: {}", proxy_url);
```
nikolaobradovic-db: Credential leak: Logs the full proxy URL. Arrow recovery now repeats this, exposing user:password@.... Redact credentials.

rust/sdk/src/proxy.rs
nikolaobradovic-db: Proxy fails open: proxy.rs:132-136 converts malformed proxy/TLS setup into None; arrow_stream.rs:536-540 then connects directly. Propagate configuration errors instead.

rust/tests/src/proxy_tests.rs
```diff
@@ -176,5 +246,210 @@ async fn test_proxy_and_no_proxy() -> Result<(), Box<dyn std::error::Error>> {
...
+
+    // === Part 5: Verify target TLS is preserved through the CONNECT tunnel ===
+    info!("=== Testing Arrow Flight TLS through proxy ===");
+    {
+        let (mock_server, server_url, ca_pem) = start_mock_tls_flight_server().await?;
+        mock_server.inject_responses(TABLE_NAME, vec![]).await;
+
+        let (proxy_url, connect_count) = start_mock_proxy().await;
+        std::env::set_var("grpc_proxy", &proxy_url);
+
```
nikolaobradovic-db, line 378: Missing coverage: proxy_tests.rs:292-308 tests an HTTPS target through an HTTP proxy, not an HTTPS proxy.

nikolaobradovic-db (dismissed):

rust/tests/src/proxy_tests.rs
teodordelibasic-db: Done.

rust/sdk/src/proxy.rs
teodordelibasic-db: Done.

rust/sdk/src/proxy.rs
teodordelibasic-db: Done.

nikolaobradovic-db (approved):

PR #570, opened 2026-07-22
[RUST] Introduce InvalidSchema error and return more details more fro…

flaviofcruz:

Zerobus server can return extra error details when creating a schema and the table schema is incompatible. Parse it and return it to the SDK caller with the different reasons for why the schema failed.

Users of the SDK can then decide what to do when the schema differs slightly.know how to interact and update your
code.

Note: forked from https://github.com/databricks/zerobus-sdk/pull/540?timeline_per_page=5

## How is this tested?

Added unit tests.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -788,6 +788,7 @@ dependencies = [
  "tonic",
  "tonic-prost",
  "tonic-prost-build",
+ "tonic-types",
  "tracing",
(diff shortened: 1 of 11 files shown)


teodordelibasic-db (approved):
LGTM

teodordelibasic-db:
Transient Windows issue for the failed merge queue.

PR #571, opened 2026-07-22
[PureGo] Add ingestion core skeleton (1/3)

zlata-stefanovic-db:
## Summary

First of a three-PR split of #531 (ingestion core). Adds `internal/stream` — the generic protocol-agnostic ingestion core that sits between the auth/transport layers and the (not yet built) public API. The core is written once and instantiated per wire protocol through three seams (`encoder`, `ackModel`, `wireStream`) so the Arrow Flight path can land later without rewriting sender / receiver / supervisor / buffer.

## Design

- **`buffer.go`** — bounded, offset-assigning queue with semaphore backpressure. Split into `reserve` + `append` so the ctx-cancellable semaphore wait happens outside the core's offset-assignment mutex; many-goroutine `Ingest` scales, and each blocked caller observes its own ctx cancellation.

- **`encoder.go`** — proto and JSON encoders build EphemeralStream payloads eagerly at `Ingest` time so the buffer holds `[]byte`, never live user objects.

- **`ackmodel.go`** — `offsetAckModel` for proto/JSON; distinguishes `ack`, `pause`, and other response kinds. Arrow's record-count ack model will slot in without touching the core.

- **`wirestream.go`** — generic `opener` / `wireStream` seam so the transport is injected, not named.

- **`core.go`** — three goroutines (sender, receiver, supervisor), monotonic ack watermark, `Flush` / `WaitForOffset` / `Close` / `GetUnacked`. Sender uses a per-`runOnce` derived context so a stream teardown unblocks `buf.next` without cancelling the supervisor context (which would signal `Close`). Offset assignment is serialized under `offsetMu` so `nextOffset` only advances for records that actually make it into the buffer; `Flush` targets the highest successfully-enqueued offset. Server pause signal is handled by aborting the current stream and reconnecting after `waitPause`.

- **`supervisor.go`** — `create → run → recover` loop. Defaults `RecoveryRetries=4`, `RecoveryBackoff=2s`, `FlushTimeout=5m`, `LackOfAckTimeout=60s`. Server EOF is treated as a retryable disconnect. Per-episode retry budget resets after a run that opened and ran (this bar tightens in PR 2/3).

- **`errors.go`** — `errClosed` sentinel + `pauseSignal`.

Also adds a `FakeStreamRPC` test seam on `transport.Stream` so the stream package's tests can inject in-process fakes without a live gRPC server.

## Stack

This is **1 of 3**. Follow-ups will land review-driven hardening on top:

- 2/3 (#572) — **Durability + callbacks** — ack validation, callback dispatcher goroutine, per-offset OnAck, non-destructive GetUnacked, retry classification, mid-stream auth invalidation, `Config{}` sanitization.
- 3/3 (#573) — **Pause + lifecycle robustness** — pause-drain (park sender + keep receiving acks), sent-state ack watermark, healthy-run definition covering idle-but-stable connections, prevalidated raw input size, `Close` bounded during pause, CI-flake fix in transport tests.

Splits [#531](https://github.com/databricks/zerobus-sdk/pull/531).

## Test plan

- `go build ./...`, `go vet ./...`, `gofmt -l .` clean.
- `go test ./... -race` passes; ~41 tests in `internal/stream`.
- Recovery test (`TestCoreStreamRecoveryRequeuesUnacked`) verifies items in flight when a stream dies are re-sent on the recovery stream and acknowledged correctly.
- Regression tests: `Ingest` on a closed stream errors; a failed `Ingest` doesn't advance the `Flush` target (no permanent gap); `Close` unblocks an `Ingest` parked at capacity; a cancelled-ctx `Ingest` reliably errors even when a buffer slot is free; server pause signal reconnects without consuming the retry budget.

## What's not here (arrives in later PRs)

- Server ack offset validation (an ack past the highest sent offset is silently trusted here — fixed in 2/3).
- Callback goroutine (`AckCallback` fires synchronously on the receiver here — fixed in 2/3).
- Non-destructive `GetUnacked` returning `(records, error)` (destructive `[][]byte` here — fixed in 3/3).
- Pause drain (aborts on pause here instead of parking the sender and keeping the receiver reading — fixed in 3/3).
- Public `zerobus` package + Arrow Flight wire path (separate future work).

### DIFF
--- a/purego/internal/stream/ackmodel.go
+++ b/purego/internal/stream/ackmodel.go
@@ -1,0 +1,70 @@
+package stream
+
+import (
+	"github.com/databricks/zerobus-sdk/purego/internal/zerobuspb"
+)
+
+// ackModel translates a raw server response into a logical "highest fully-acked
+// offset" that the core's watermark can compare against, and classifies
+// non-ack responses so the core stays blind to the concrete wire type. It is
+// edge #2 of the design and is generic over the response type Resp: proto/JSON
+// supply ackModel[ephemeralResp]; Arrow will supply its own over a Flight
+// response, mapping a cumulative record count back to an offset.
+//
+// This interface is intentionally the minimal offset-only subset. The
+// architecture note (§3a) sketches a wider Track/Resolve/Unacked shape for the
+// Arrow record-count model, where an ack can land mid-batch and recovery must
+// slice a straddling batch. That width is deferred until the Arrow wire path
+// lands; adding it now would be unused machinery. The core stays offset-only
+// and blind to the record-vs-batch distinction regardless.
+type ackModel[Resp any] interface {
+	// classify inspects a server response and reports what it means to the core:
+	//   - kind == ackResponse: off is the highest fully-acked offset.
+	//   - kind == pauseResponse: the server asked the client to pause; pause
+	//     carries the requested duration. The core waits then reconnects.
+	//   - kind == otherResponse: anything else (ignored by the receiver).
+	// Keeping close-signal detection here means the receiver never names a
+	// concrete proto message.
+	classify(resp Resp) (kind respKind, off int64, pause pauseSignal)
+}
+
+// respKind is the category the ackModel assigns to a server response.
+type respKind int
+
+const (
+	otherResponse respKind = iota // no ack, no pause — ignored
+	ackResponse                   // carries a durability ack offset
+	pauseResponse                 // server-requested pause (close-stream signal)
+)
+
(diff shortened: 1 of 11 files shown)


zlata-stefanovic-db:
Superseded by a four-way split — each PR now fits under the 1,500-line target:

- 1/4 primitives — #575
- 2/4 core + supervisor + wirestream — #576
- 3/4 durability + callbacks + retry classification — #577
- 4/4 pause & lifecycle robustness — #573 (repurposed)

Closing.

PR #572, opened 2026-07-22
[PureGo] Ingestion core: durability, callbacks, retry classification (2/3)

zlata-stefanovic-db:
## Summary

Second of a three-PR split of #531. Stacks on #571. Addresses the durability, callback, and retry-classification review findings on the ingestion core skeleton. Public shape is unchanged; internals harden.

## Durability

- **Watermark bounded to discarded flight.** `wm.advance` now runs only to the highest offset of a record actually removed from flight by `discardThrough`, so a bogus high server offset cannot fake durability for records never durably stored.
- **Ack-model protocol errors.** New `unknownResponse` and `malformedResponse` kinds. An unknown response type fails the stream; an ack whose `DurabilityAckUpToOffset` field is absent (proto optional → would silently default to 0) fails the stream rather than fabricating a zero-offset ack.
- **Monotonic `lastEnqueued`.** `lastEnqueued.Store` is inside `offsetMu` so two concurrent `Ingest`s cannot land their `Store`s out of order and make `Flush` return success while a higher offset is still unacked.
- **Wire-size payload cap.** `MaxPayloadBytes` enforcement now checks `proto.Size` of the encoded request, not just the raw input byte count, so proto framing overhead can't push a "just under limit" batch over the server's message size limit.
- **Caller-byte cloning.** Proto encoder snapshots each record so a caller reusing a marshalling scratch buffer between `Ingest` calls can't mutate queued or replayed payloads.

## Callbacks

- **Dedicated `callbackDispatcher` goroutine.** User `OnAck` / `OnError` runs serially on its own goroutine so a slow user callback never stalls receiver or supervisor. Callbacks calling `Close` from inside the dispatcher no longer self-deadlock (reentrancy detected via an atomic flag).
- **Per-offset `OnAck`.** A cumulative server ack for offset 5 following offset 2 now yields three `OnAck(3)`, `OnAck(4)`, `OnAck(5)` events, not one `OnAck(5)`. A duplicate ack fires zero.
- **Terminal drain payload retention.** On terminal failure the supervisor stores encoded items in `retainedFailed` AND dispatches per-offset `OnError`, so `GetUnacked` and the callback consumer see the same records — the two are no longer mutually exclusive.
- **Clean-Close watermark termination.** `wm.closeForClean` fires on Close so a `WaitForOffset` waiter on an unreachable offset unblocks with an error instead of hanging on a background ctx.

## Retry classification

- **`openBudgetExceeded` self-classifies retryable.** An Open that hits our internal `RecoveryTimeout` budget is wrapped in an error whose `IsRetryable()` returns true, so a stalled dial exhausts `RecoveryRetries` instead of dying after one attempt.
- **`isRetryable` reorder.** Self-classifying wrappers are checked before generic `context.DeadlineExceeded` / `Canceled` so an internal-budget error inside a `DeadlineExceeded` chain stays retryable.
- **Mid-stream auth invalidation.** When a mid-stream error is a gRPC auth rejection, the supervisor calls `HeadersProvider.Invalidate` so the next Open re-mints instead of reconnecting with a stale credential. Adds `transport.IsAuthRejection` for the stream layer to key on.

## Config safety

- **`sanitizeConfig`.** Non-positive `MaxInflight` / timeouts / `MaxPayloadBytes` are normalized to their defaults so a caller-passed `Config{}` does not deadlock (unbuffered semaphore), spin (zero lack-of-ack timer), or hard-abort teardown immediately.
- **`Config{}` recovers by default.** `RecoveryRetries=0` is normalized to the default; explicit single-attempt streams use `RecoverySetting=RecoveryDisabled`.

## Cleanup

- Buffer `discardThrough` / `next` / `requeue` zero out departed slots so acknowledged payloads become GC-collectible instead of being pinned by the underlying slice array.
- `buffer.enqueue` and `buffer.len` are production-dead (used only by tests after the `reserve`/`append` split); moved to `export_test.go`.
- Descriptor comments on `NewEphemeralOpener` and encoder-`decode` methods scrubbed of internal "edge #1/#2/#3" numbering.

## Stack

This is **2 of 3**. Splits [#531](https://github.com/databricks/zerobus-sdk/pull/531).
- 1/3 (parent): the ingestion core skeleton — #571.
- 3/3 (follow-up): pause & lifecycle robustness + sent-state ack + CI flake — #573.

## Test plan

- `go build ./...`, `go vet ./...`, `gofmt -l .` clean.
- `go test ./... -race -count=10` passes.
- Regression coverage: concurrent `Ingest` + `Flush` (proves `lastEnqueued` monotonic), bogus-high server ack fails the stream, missing-offset ack fails the stream, per-offset `OnAck` sequence + duplicate-ack-no-callback, callback-invoked `Close` does not deadlock, `WaitForOffset` unblocks on `Close`, GetUnacked with callback preserves payloads, `Config{}` does not deadlock, blocked-ingest observes its own ctx cancel, first-send-after-idle gets full lack-of-ack budget, invalid-params open not retried, self-classified non-retryable OAuth error not retried, mid-stream auth invalidation, callback panic containment, payload-too-large single/batch.

## What's not here (arrives in 3/3)

- Pause drain: on `CloseStreamSignal` the sender is parked and the receiver keeps draining acks; today this branch still aborts and reconnects on pause (though it now enforces `StreamPausedMaxWait`).
- Sent-state ack validation: an ack for a queued-or-mid-Send record is still accepted here (offset-in-range).
- `GetUnacked` still destructively closes an active stream.
- `Config`'s `RecoveryTimeout` / `CallbackTeardownTimeout` / `StreamPausedMaxWaitMode` / raw-size precheck / `stampOffset` encoder split.
- Healthy-run definition covering idle-but-stable connections.
- CI flake fix in `transport_test.go`.

### DIFF
--- a/purego/internal/stream/ackmodel.go
+++ b/purego/internal/stream/ackmodel.go
@@ -4,25 +4,33 @@ import (
 	"github.com/databricks/zerobus-sdk/purego/internal/zerobuspb"
 )
 
-// ackModel translates a raw server response into a logical "highest fully-acked
-// offset" that the core's watermark can compare against, and classifies
-// non-ack responses so the core stays blind to the concrete wire type. It is
-// edge #2 of the design and is generic over the response type Resp: proto/JSON
-// supply ackModel[ephemeralResp]; Arrow will supply its own over a Flight
-// response, mapping a cumulative record count back to an offset.
+// ackModel translates a raw server response into a logical "highest
+// fully-acked offset" that the core's watermark can compare against, and
+// classifies non-ack responses so the core stays blind to the concrete wire
+// type. It is the receive-side per-encoding seam and is generic over the
+// response type Resp: proto/JSON supply ackModel[ephemeralResp]; Arrow will
+// supply its own over a Flight response, mapping a cumulative record count
+// back to an offset.
 //
-// This interface is intentionally the minimal offset-only subset. The
-// architecture note (§3a) sketches a wider Track/Resolve/Unacked shape for the
-// Arrow record-count model, where an ack can land mid-batch and recovery must
-// slice a straddling batch. That width is deferred until the Arrow wire path
-// lands; adding it now would be unused machinery. The core stays offset-only
-// and blind to the record-vs-batch distinction regardless.
+// This interface is intentionally the minimal offset-only subset. A wider
+// Track/Resolve/Unacked shape will be needed for the Arrow record-count
+// model, where an ack can land mid-batch and recovery must slice a
+// straddling batch; that width is deferred until the Arrow wire path lands.
+// The core stays offset-only and blind to the record-vs-batch distinction
+// regardless.
 type ackModel[Resp any] interface {
 	// classify inspects a server response and reports what it means to the core:
-	//   - kind == ackResponse: off is the highest fully-acked offset.
+	//   - kind == ackResponse: off is the highest fully-acked offset the server
+	//     claims (still requires the receiver to bound-check against sent work).
 	//   - kind == pauseResponse: the server asked the client to pause; pause
 	//     carries the requested duration. The core waits then reconnects.
-	//   - kind == otherResponse: anything else (ignored by the receiver).
+	//   - kind == unknownResponse: the response type is not one the core
+	//     recognises. The receiver fails the stream: silently ignoring
+	//     unexpected messages hides server contract violations.
+	//   - kind == malformedResponse: the response is an ack but its offset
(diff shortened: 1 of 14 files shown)


zlata-stefanovic-db:
Superseded by a four-way split — each PR now fits under the 1,500-line target:

- 1/4 primitives — #575
- 2/4 core + supervisor + wirestream — #576
- 3/4 durability + callbacks + retry classification — #577
- 4/4 pause & lifecycle robustness — #573 (repurposed)

Closing.

PR #573, opened 2026-07-22
[PureGo] Harden stream lifecycle (4/4)

zlata-stefanovic-db:
## Summary

**4 of 4** — completes the generic proto/JSON ingestion core with lifecycle, recovery, callback, transport, and bounded-memory hardening on top of #577.

## Changes

- Preserves pause deadlines while ensuring protocol, authentication, and transport failures remain authoritative.
- Prioritizes receiver terminal causes over sender errors induced by teardown, preventing retries or credential reuse after definitive server rejection.
- Centralizes credential invalidation in the supervisor, including Open/live-stream rejections that race `Close`, and normalizes table-name cache keys.
- Adds FIFO count-and-byte backpressure across queued and in-flight work.
- Charges conservative encoded retained size—including framing, per-record containers, and request overhead—so tiny or empty batches cannot bypass the memory budget.
- Makes buffer reservation rollback idempotent across `close`/`drain` races and removes obsolete timer-drain handling for Go 1.25.
- Guards logical and physical offset exhaustion.
- Makes encoding offset-independent with conservative wire-size validation and connection-local physical-to-logical ACK translation.
- Reuses per-connection send synchronization and a persistent receive pump with deterministic worker teardown.
- Strengthens callback ordering, bounded shutdown, recovery, backpressure, authentication-race, and concurrency coverage.
- Adds ACK-path and recovery-requeue benchmarks and deterministically reaps fake gRPC servers in transport tests.
- Cleans up internal test hooks, dead aliases, and stale lifecycle/backpressure documentation.

## Verification

- `go test -race ./...`
- `go vet ./...`
- `git diff --check`
- Focused lifecycle, buffer-accounting, pause-arbitration, and authentication-race tests repeated under the race detector.

## Size

16 files changed, **1,640 additions**, 297 deletions.

## Stack

- 1/4 — primitives — #575
- 2/4 — basic core stream — #576
- 3/4 — durability and callbacks — #577
- 4/4 — lifecycle and performance hardening — this PR

### DIFF
--- a/purego/internal/stream/benchmark_test.go
+++ b/purego/internal/stream/benchmark_test.go
@@ -1,0 +1,85 @@
+package stream
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+)
+
+func BenchmarkCoreStreamAckPath(b *testing.B) {
+	for _, tc := range []struct {
+		name  string
+		batch bool
+	}{
+		{name: "single"},
+		{name: "batch", batch: true},
+	} {
+		b.Run(tc.name, func(b *testing.B) {
+			rpc := newFakeRPC()
+			cfg := testConfig()
+			cfg.MaxInflight = 1024
+			cs := newCoreForTest(testParams(), cfg, newFakeOpener(rpc), nil)
+			defer cs.Close()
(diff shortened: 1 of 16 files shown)


teodordelibasic-db (approved):
LGTM first try 😆

PR #575, opened 2026-07-22
[PureGo] Ingestion primitives: buffer, encoder, ackmodel (1/4)

zlata-stefanovic-db:
## Summary

**1 of 4** — splits the ingestion core work in [#531](https://github.com/databricks/zerobus-sdk/pull/531).

Adds `internal/stream` primitives: three independent building blocks the ingestion core is composed of. Each is standalone-correct with its own tests; a subsequent PR wires them into a working stream.

## Contents

- **`buffer.go`** — bounded, offset-assigning queue with semaphore backpressure. `enqueue`/`next`/`requeue`/`discardThrough`/`drain`/`close`/`inFlight`/`len`. Ctx-cancellable at both `enqueue` and `next` so goroutines stop cleanly per stream. `doneCh` unblocks callers parked on the semaphore when the buffer closes.
- **`encoder.go`** — proto and JSON encoders build `EphemeralStream` payloads eagerly at `Ingest` time so the buffer holds `[]byte`, never live user objects. Generic `encoder[Req]` interface so the future Arrow path can slot in.
- **`ackmodel.go`** — `offsetAckModel` for proto/JSON. Distinguishes `ackResponse`, `pauseResponse`, and `otherResponse` (ignored). Generic `ackModel[Resp]` interface.
- **`errors.go`** — `errClosed` sentinel + `errUnsupportedRecordType` helper + `pauseSignal` type.

## Stack

- 1/4 (this) — primitives
- 2/4 — core + supervisor + wirestream
- 3/4 — durability + callbacks + retry classification
- 4/4 — pause & lifecycle robustness

Replaces the earlier three-way split ([#571](https://github.com/databricks/zerobus-sdk/pull/571) / [#572](https://github.com/databricks/zerobus-sdk/pull/572) / [#573](https://github.com/databricks/zerobus-sdk/pull/573)) which each exceeded 1500 lines.

## Test plan

`go test ./internal/stream/ -race` passes. Tests cover: buffer FIFO, backpressure, ctx-cancel unblock, requeue ordering, drain, closed-buffer errors, concurrent enqueue/discard; encoder single + batch for proto and JSON, empty-record rejection, batch-record extraction; ack-model classification (ack / pause / other).

### DIFF
--- a/purego/internal/stream/ackmodel.go
+++ b/purego/internal/stream/ackmodel.go
@@ -49,7 +49,11 @@ func (offsetAckModel) classify(resp ephemeralResp) (respKind, int64, pauseSignal
 		if ack.DurabilityAckUpToOffset == nil {
 			return malformedResponse, 0, pauseSignal{}
 		}
-		return ackResponse, *ack.DurabilityAckUpToOffset, pauseSignal{}
+		off := *ack.DurabilityAckUpToOffset
+		if off < 0 {
+			return malformedResponse, 0, pauseSignal{}
+		}
+		return ackResponse, off, pauseSignal{}
 	}
 	return unknownResponse, 0, pauseSignal{}
 }
--- a/purego/internal/stream/ackmodel.go
+++ b/purego/internal/stream/ackmodel.go
@@ -12,18 +12,9 @@ import (
 // TODO(arrow): the Arrow wire path will supply its own ackModel over a Flight
 // response, mapping a cumulative record count back to an offset.
 type ackModel[Resp any] interface {
-	// classify inspects a server response and reports what it means to the core:
-	//   - kind == ackResponse: off is the highest fully-acked offset.
-	//   - kind == pauseResponse: the server asked the client to pause; pause
-	//     carries the requested duration. The core waits then reconnects.
-	//   - kind == unknownResponse: an unrecognized response type. The receiver
-	//     must fail the stream rather than ignore it, since silently dropping
-	//     unexpected messages hides a wire-contract mismatch.
-	//   - kind == malformedResponse: an ack whose offset field is absent.
-	//     Treating the zero default as a real ack would fake durability for
-	//     offset 0, so the receiver must fail the stream.
-	// Keeping close-signal detection here means the receiver never names a
-	// concrete proto message.
+	// classify reports what a server response means to the core: an ack offset,
+	// a pause request, or an unknown/malformed response the receiver must fail
+	// on. Detecting these here keeps the receiver blind to concrete proto types.
 	classify(resp Resp) (kind respKind, off int64, pause pauseSignal)
 }
 
@@ -54,9 +45,7 @@ func (offsetAckModel) classify(resp ephemeralResp) (respKind, int64, pauseSignal
 		return pauseResponse, 0, pauseSignal{duration: sig.GetDuration().AsDuration()}
 	}
 	if ack := resp.GetIngestRecordResponse(); ack != nil {
-		// DurabilityAckUpToOffset is an optional field; GetDurabilityAckUpToOffset
-		// would turn an absent value into 0 and fake an ack for offset 0. Preserve
-		// field presence and classify a missing offset as malformed.
+		// Absent offset must be malformed, not a fabricated ack for offset 0.
 		if ack.DurabilityAckUpToOffset == nil {
 			return malformedResponse, 0, pauseSignal{}
 		}
--- a/purego/internal/stream/ackmodel.go
+++ b/purego/internal/stream/ackmodel.go
@@ -1,0 +1,70 @@
+package stream
+
+import (
+	"github.com/databricks/zerobus-sdk/purego/internal/zerobuspb"
+)
+
+// ackModel translates a raw server response into a logical "highest fully-acked
+// offset" that the core's watermark can compare against, and classifies
+// non-ack responses so the core stays blind to the concrete wire type. It is
+// edge #2 of the design and is generic over the response type Resp: proto/JSON
+// supply ackModel[ephemeralResp]; Arrow will supply its own over a Flight
+// response, mapping a cumulative record count back to an offset.
+//
+// This interface is intentionally the minimal offset-only subset. The
+// architecture note (§3a) sketches a wider Track/Resolve/Unacked shape for the
+// Arrow record-count model, where an ack can land mid-batch and recovery must
+// slice a straddling batch. That width is deferred until the Arrow wire path
+// lands; adding it now would be unused machinery. The core stays offset-only
+// and blind to the record-vs-batch distinction regardless.
+type ackModel[Resp any] interface {
+	// classify inspects a server response and reports what it means to the core:
+	//   - kind == ackResponse: off is the highest fully-acked offset.
+	//   - kind == pauseResponse: the server asked the client to pause; pause
+	//     carries the requested duration. The core waits then reconnects.
+	//   - kind == otherResponse: anything else (ignored by the receiver).
+	// Keeping close-signal detection here means the receiver never names a
+	// concrete proto message.
+	classify(resp Resp) (kind respKind, off int64, pause pauseSignal)
+}
+
+// respKind is the category the ackModel assigns to a server response.
+type respKind int
+
+const (
+	otherResponse respKind = iota // no ack, no pause — ignored
+	ackResponse                   // carries a durability ack offset
+	pauseResponse                 // server-requested pause (close-stream signal)
+)
+
+// ephemeralResp is the proto/JSON server response type. Aliased so the core's
+// Resp type parameter is named once here rather than spelled out at every use.
+type ephemeralResp = *zerobuspb.EphemeralStreamResponse
+
+// offsetAckModel is the proto/JSON ack model: the server sends
+// DurabilityAckUpToOffset directly as the logical offset, and a
+// CloseStreamSignal requests a pause.
+type offsetAckModel struct{}
+
+func (offsetAckModel) classify(resp ephemeralResp) (respKind, int64, pauseSignal) {
+	if resp == nil {
+		return otherResponse, 0, pauseSignal{}
+	}
+	if sig := resp.GetCloseStreamSignal(); sig != nil {
+		return pauseResponse, 0, pauseSignal{duration: sig.GetDuration().AsDuration()}
+	}
+	if ack := resp.GetIngestRecordResponse(); ack != nil {
+		return ackResponse, ack.GetDurabilityAckUpToOffset(), pauseSignal{}
+	}
+	return otherResponse, 0, pauseSignal{}
+}
+
+// newAckModel returns the proto/JSON ack model for the given record type.
+func newAckModel(rt zerobuspb.RecordType) (ackModel[ephemeralResp], error) {
+	switch rt {
+	case zerobuspb.RecordType_PROTO, zerobuspb.RecordType_JSON:
+		return offsetAckModel{}, nil
+	default:
+		return nil, errUnsupportedRecordType(rt)
+	}
+}
--- a/purego/internal/stream/ackmodel.go
+++ b/purego/internal/stream/ackmodel.go
@@ -7,22 +7,21 @@ import (
 // ackModel translates a raw server response into a logical "highest fully-acked
 // offset" that the core's watermark can compare against, and classifies
 // non-ack responses so the core stays blind to the concrete wire type. It is
-// edge #2 of the design and is generic over the response type Resp: proto/JSON
-// supply ackModel[ephemeralResp]; Arrow will supply its own over a Flight
-// response, mapping a cumulative record count back to an offset.
+// generic over the response type Resp: proto/JSON supply ackModel[ephemeralResp].
 //
-// This interface is intentionally the minimal offset-only subset. The
-// architecture note (§3a) sketches a wider Track/Resolve/Unacked shape for the
-// Arrow record-count model, where an ack can land mid-batch and recovery must
-// slice a straddling batch. That width is deferred until the Arrow wire path
-// lands; adding it now would be unused machinery. The core stays offset-only
-// and blind to the record-vs-batch distinction regardless.
+// TODO(arrow): the Arrow wire path will supply its own ackModel over a Flight
+// response, mapping a cumulative record count back to an offset.
 type ackModel[Resp any] interface {
 	// classify inspects a server response and reports what it means to the core:
 	//   - kind == ackResponse: off is the highest fully-acked offset.
 	//   - kind == pauseResponse: the server asked the client to pause; pause
 	//     carries the requested duration. The core waits then reconnects.
-	//   - kind == otherResponse: anything else (ignored by the receiver).
+	//   - kind == unknownResponse: an unrecognized response type. The receiver
+	//     must fail the stream rather than ignore it, since silently dropping
+	//     unexpected messages hides a wire-contract mismatch.
+	//   - kind == malformedResponse: an ack whose offset field is absent.
+	//     Treating the zero default as a real ack would fake durability for
+	//     offset 0, so the receiver must fail the stream.
 	// Keeping close-signal detection here means the receiver never names a
 	// concrete proto message.
 	classify(resp Resp) (kind respKind, off int64, pause pauseSignal)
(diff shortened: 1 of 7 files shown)


purego/internal/stream/ackmodel.go
```diff
@@ -0,0 +1,70 @@
+package stream
+
+import (
+	"github.com/databricks/zerobus-sdk/purego/internal/zerobuspb"
+)
+
+// ackModel translates a raw server response into a logical "highest fully-acked
+// offset" that the core's watermark can compare against, and classifies
+// non-ack responses so the core stays blind to the concrete wire type. It is
+// edge #2 of the design and is generic over the response type Resp: proto/JSON
```
teodordelibasic-db: It doesn't matter that much at this point, but I don't think there is a need to mention design choices etc. in code comments. It's okay if we remember to keep everything up to date as more PRs land, but it probably adds extra work and stuff to think about.

Just comments explaining current code state plus TODO comments for following PRs and stuff such as Arrow support would be okay, but whatever is easiest for you.

purego/internal/stream/buffer.go
```diff
@@ -0,0 +1,209 @@
...
+// list, returning the item. The sender must later call discard (on ack) or
+// requeue (on reconnect) for every item returned by next.
+//
+// Returns errClosed when the buffer has been closed and drained.
+// Returns ctx.Err() if ctx is cancelled while waiting.
+func (b *buffer[Req]) next(ctx context.Context) (item[Req], error) {
+	b.mu.Lock()
+	for len(b.queue) == 0 && !b.closed {
+		// Wake up on ctx cancellation too.
+		stop := context.AfterFunc(ctx, func() { b.cond.Broadcast() })
```
teodordelibasic-db: `Broadcast` is not ordered with `Wait`, so cancellation can land after the empty-queue check but before the waiter is registered. The callback then fires too early, and `next` can sleep forever on an already-cancelled context. The stacked core cancels this context and waits for the sender to exit, so this can hang recovery and `Close`.

Could we take `b.mu` in the callback so the broadcast cannot run until `Wait` has registered the waiter and released the lock?

```go
stop := context.AfterFunc(ctx, func() {
        b.mu.Lock()
        b.cond.Broadcast()
        b.mu.Unlock()
})
```

purego/internal/stream/ackmodel.go
```diff
@@ -0,0 +1,70 @@
...
+
+func (offsetAckModel) classify(resp ephemeralResp) (respKind, int64, pauseSignal) {
+	if resp == nil {
+		return otherResponse, 0, pauseSignal{}
+	}
+	if sig := resp.GetCloseStreamSignal(); sig != nil {
+		return pauseResponse, 0, pauseSignal{duration: sig.GetDuration().AsDuration()}
+	}
+	if ack := resp.GetIngestRecordResponse(); ack != nil {
+		return ackResponse, ack.GetDurabilityAckUpToOffset(), pauseSignal{}
```
teodordelibasic-db: `GetDurabilityAckUpToOffset` turns an absent optional field into `0`. Since stream offsets start at `0`, an empty `IngestRecordResponse` is therefore accepted as a real ack for the first record. The stacked receiver advances its watermark and discards through that offset, so `Flush` can report durability that the server never acknowledged.

Could we preserve field presence and classify the response as a protocol error instead?

```go
if ack := resp.GetIngestRecordResponse(); ack != nil {
        if ack.DurabilityAckUpToOffset == nil {
                return malformedResponse, 0, pauseSignal{}
        }
        return ackResponse, *ack.DurabilityAckUpToOffset, pauseSignal{}
}
```

purego/internal/stream/encoder.go
```diff
@@ -0,0 +1,163 @@
...
+
+func (protoEncoder) encode(offset int64, record []byte) (encodedMsg, error) {
+	if len(record) == 0 {
+		return nil, fmt.Errorf("stream: proto record must not be empty")
+	}
+	return &zerobuspb.EphemeralStreamRequest{
+		Payload: &zerobuspb.EphemeralStreamRequest_IngestRecord{
+			IngestRecord: &zerobuspb.IngestRecordRequest{
+				OffsetId: proto.Int64(offset),
+				Record:   &zerobuspb.IngestRecordRequest_ProtoEncodedRecord{ProtoEncodedRecord: record},
```
teodordelibasic-db: The queued protobuf payload still aliases caller-owned bytes. `Ingest` returns before the sender serializes the request, so reusing a scratch buffer can mutate the queued record, race with marshaling, and corrupt a recovery replay. The batch path has the same problem at [`encoder.go:72`](https://github.com/databricks/zerobus-sdk/blob/190a62d51e2eed14acaca425327c358542a6976e/purego/internal/stream/encoder.go#L72), including aliases to every inner slice.

Could we snapshot both forms when encoding?

```go
recordCopy := bytes.Clone(record)

recordCopies := make([][]byte, len(records))
for i := range records {
        recordCopies[i] = bytes.Clone(records[i])
}
```

purego/internal/stream/buffer.go
```diff
@@ -0,0 +1,209 @@
...
+	return nil
+}
+
+// next blocks until a pending item is available and moves it to the in-flight
+// list, returning the item. The sender must later call discard (on ack) or
+// requeue (on reconnect) for every item returned by next.
+//
+// Returns errClosed when the buffer has been closed and drained.
+// Returns ctx.Err() if ctx is cancelled while waiting.
+func (b *buffer[Req]) next(ctx context.Context) (item[Req], error) {
```
teodordelibasic-db, line 116: `next` only checks cancellation after `Wait`. If the queue is already non-empty, a cancelled sender skips that loop and dequeues another record. On `Close`, the buffer is marked closed but pending items remain, so the sender can continue draining them instead of stopping. During recovery or a server pause, it can similarly put more work on a stream that is being torn down.

Could we check the context before locking and again immediately before removing the item?

```go
if err := ctx.Err(); err != nil {
        return item[Req]{}, err
}
b.mu.Lock()

// After the wait loop and closed check.
if err := ctx.Err(); err != nil {
        b.mu.Unlock()
        return item[Req]{}, err
}
```

purego/internal/stream/encoder.go
```diff
@@ -0,0 +1,163 @@
...
+
+// encoder turns user records into wire messages for a given offset and recovers
+// them again for GetUnacked. It is the send-side per-encoding
+// seam. The core is generic over the wire message type Req and
+// never names a concrete proto type — proto/JSON supply encoder[encodedMsg];
+// Arrow will supply encoder[flightFrame].
+//
+// The offset is stamped into the message here so re-sends after recovery reuse
+// the same logical offset the server already saw.
+type encoder[Req any] interface {
```
teodordelibasic-db, line 27: `MaxInflight` bounds entry count, not bytes, and this encoder seam has no way to report the serialized request size. A single huge batch can consume unbounded client memory, while a request over the server's roughly 10 MiB limit fails at send time and can consume recovery attempts even though retrying cannot make it valid.

Could we expose the exact wire size here and enforce a configured limit before enqueueing?

```go
type encoder[Req any] interface {
        encode(offset int64, record []byte) (Req, error)
        encodeBatch(offset int64, records [][]byte) (Req, error)
        decode(msg Req) [][]byte
        wireSize(msg Req) int
}

func (protoEncoder) wireSize(msg encodedMsg) int {
        return proto.Size(msg)
}
```

It would also be good to reject obviously oversized raw input before cloning it, then use `wireSize` after encoding to account for protobuf framing.

purego/internal/stream/ackmodel.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	return otherResponse, 0, pauseSignal{}
```
teodordelibasic-db: Unknown and nil responses are classified as `otherResponse`, which the interface explicitly defines as ignored. A wire-contract mismatch then looks like a lack-of-ack timeout, or can remain hidden indefinitely while the stream is idle. The Rust receiver fails immediately on unexpected response payloads.

Could we give unknown and malformed responses distinct kinds and require the receiver to fail them?

```go
const (
        ackResponse respKind = iota
        pauseResponse
        unknownResponse
        malformedResponse
)
```

purego/internal/stream/buffer.go
```diff
@@ -0,0 +1,209 @@
...
+	doneCh   chan struct{} // closed when the buffer is closed/drained; unblocks sem waiters
+}
+
+func newBuffer[Req any](maxInflight int) *buffer[Req] {
+	// queue/flight grow on demand; don't preallocate to maxInflight, which with
+	// the default 1M cap would reserve tens of MB per stream before a single
+	// record is ingested. Only the semaphore is sized to the cap, since it is the
+	// backpressure gate and its capacity defines the bound.
+	b := &buffer[Req]{
+		sem:    make(chan struct{}, maxInflight),
```
teodordelibasic-db, line 67: `newBuffer(0)` creates an unbuffered semaphore with no receiver during reservation, so every `enqueue` blocks until cancellation or close. A negative capacity panics in `make`. Since this PR describes the buffer as a standalone primitive, the constructor should not accept values that break liveness.

Could we reject the invalid input at this boundary, or guarantee normalization before construction?

```go
func newBuffer[Req any](maxInflight int) (*buffer[Req], error) {
        if maxInflight <= 0 {
                return nil, fmt.Errorf("stream: max inflight must be positive")
        }
        // Construct the buffer.
}
```

A zero and negative capacity test would make the chosen contract explicit.

purego/internal/stream/encoder.go
```diff
@@ -0,0 +1,163 @@
...
+	// yields all of its records so no unacked record is silently dropped.
+	decode(msg Req) [][]byte
+}
+
+// protoEncoder builds EphemeralStream payloads for proto-encoded records (raw
+// serialized protobuf bytes), single and batched.
+type protoEncoder struct{}
+
+func (protoEncoder) encode(offset int64, record []byte) (encodedMsg, error) {
+	if len(record) == 0 {
```
teodordelibasic-db: An empty byte slice is a valid protobuf encoding for a message whose fields all have default values. Rejecting it makes such records impossible to ingest and differs from the Rust SDK, which accepts an empty `ProtoBytes` value. The batch loop repeats the same rejection at [`encoder.go:63`](https://github.com/databricks/zerobus-sdk/blob/190a62d51e2eed14acaca425327c358542a6976e/purego/internal/stream/encoder.go#L63).

Could we remove the proto record length checks while retaining the empty JSON and zero-record batch checks? A test using a default protobuf message should assert that both single and non-empty batch encoding accept its zero-length serialization.

purego/internal/stream/buffer.go
```diff
@@ -0,0 +1,209 @@
...
+		if ctx.Err() != nil {
+			b.mu.Unlock()
+			return item[Req]{}, ctx.Err()
+		}
+	}
+	if len(b.queue) == 0 {
+		b.mu.Unlock()
+		return item[Req]{}, errClosed
+	}
+	it := b.queue[0]
```
teodordelibasic-db, line 147: Reslicing `queue` and `flight` leaves the removed `payload` references in their backing arrays. Those arrays can stay live while later entries remain, retaining already-acknowledged encoded records under sustained ingestion. With a large inflight cap and large batches, the retained memory can be substantial.

Could we clear each departed slot before reslicing?

```go
it := b.queue[0]
var zero item[Req]
b.queue[0] = zero
b.queue = b.queue[1:]

// In discardThrough.
b.flight[0] = zero
b.flight = b.flight[1:]
```

purego/internal/stream/buffer.go
zlata-stefanovic-db: Yes, fixed as suggested

purego/internal/stream/ackmodel.go
zlata-stefanovic-db: Fixed. In purego/internal/stream/ackmodel.go, I now check optional-field presence directly. If DurabilityAckUpToOffset is absent, it is classified as malformedResponse instead of treating it as a real ack at 0.

purego/internal/stream/encoder.go
zlata-stefanovic-db: Fixed. In purego/internal/stream/encoder.go, proto single and batch inputs are cloned with bytes.Clone(...) before queueing, so caller buffer reuse cannot mutate queued/replayed payloads.

purego/internal/stream/buffer.go
zlata-stefanovic-db: Fixed. next() now checks ctx.Err() both before entering the wait/lock flow and again right before dequeueing. If cancellation happened, it returns immediately and does not pull another item.

purego/internal/stream/encoder.go
zlata-stefanovic-db: Fixed at seam level here. I added wireSize(msg) to the encoder interface and implemented it with proto.Size(...) for proto/json messages.
Then follow-up PRs enforce limits using this (#577) and add pre-encode raw-size checks (#573).

purego/internal/stream/ackmodel.go
zlata-stefanovic-db: Fixed in classification. ackmodel now returns explicit unknownResponse and malformedResponse kinds instead of “other/ignore”.
Follow-up core wiring (#577) treats these as protocol errors and fails the stream.

purego/internal/stream/buffer.go
zlata-stefanovic-db: Fixed. newBuffer() now normalizes non-positive values to a safe default (defaultMaxInflight) so this primitive stays live and does not deadlock/panic on bad input.

purego/internal/stream/encoder.go
zlata-stefanovic-db: Fixed. Proto encoder no longer rejects empty byte slices (valid default-message encoding). JSON empty checks remain. I also added tests for empty proto single and batch cases.

purego/internal/stream/buffer.go
zlata-stefanovic-db: Fixed. In purego/internal/stream/buffer.go, removed slots are explicitly zeroed before reslicing in next() and discardThrough(), so acked payloads become GC-eligible promptly.

purego/internal/stream/ackmodel.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	if sig := resp.GetCloseStreamSignal(); sig != nil {
+		return pauseResponse, 0, pauseSignal{duration: sig.GetDuration().AsDuration()}
+	}
+	if ack := resp.GetIngestRecordResponse(); ack != nil {
+		// Absent offset must be malformed, not a fabricated ack for offset 0.
+		if ack.DurabilityAckUpToOffset == nil {
+			return malformedResponse, 0, pauseSignal{}
+		}
+		return ackResponse, *ack.DurabilityAckUpToOffset, pauseSignal{}
```
teodordelibasic-db: nit: Can be follow up or this PR, however you prefer.

Negative durability offsets are accepted as valid acks even though stream offsets start at `0`. In the final stacked receiver, a negative ack discards nothing but still refreshes `LackOfAckTimeout` while records remain in flight. Repeated malformed acks can therefore prevent recovery and leave `WaitForOffset` blocked indefinitely.

Could we classify offsets below `0` as `malformedResponse`?

```go
off := *ack.DurabilityAckUpToOffset
if off < 0 {
        return malformedResponse, 0, pauseSignal{}
}
return ackResponse, off, pauseSignal{}
```

`ackmodel_test.go` can exercise this directly with `proto.Int64(-1)` and assert `malformedResponse`.

teodordelibasic-db (dismissed):

purego/internal/stream/ackmodel.go
zlata-stefanovic-db: Fixed here, thanks!

I updated ackmodel to treat negative durability offsets as malformed:

- In purego/internal/stream/ackmodel.go, after checking field presence, I now validate the value and return malformedResponse when DurabilityAckUpToOffset < 0 (instead of accepting it as an ack).

- Added TestClassifyMalformedAckNegativeOffset in purego/internal/stream/ackmodel_test.go using proto.Int64(-1) to assert the malformed classification.

Also ran go test ./internal/stream and it passes.

teodordelibasic-db (approved):
Nice

PR #576, opened 2026-07-22
[PureGo] Ingestion core stream and supervisor (2/4)

zlata-stefanovic-db:
## Summary

**2 of 4** — builds the basic PureGo ingestion stream from the primitives in #575.

This PR is limited to the core API, transport seam, worker lifecycle, and essential correctness. Broader durability and lifecycle hardening remains isolated in #577.

## Contents

- **Core API** — asynchronous `Ingest` / `IngestBatch`, monotonic offsets, `Flush`, `WaitForOffset`, `GetUnacked`, `Close`, and `IsClosed`.
- **Workers** — one sender, one receiver, and a supervisor that opens and recovers streams.
- **ACK ordering** — ACKs received while `Send` is unresolved are held until its outcome is known. Successful sends release the ACK; failed sends cannot establish durability.
- **Recovery budget** — immediate EOF loops consume `RecoveryRetries`. Ack progress or a stable connection interval resets the retry episode.
- **Buffer integration** — queued records move to flight, cumulative ACKs release capacity, and recovery requeues unacknowledged items.
- **Graceful close** — clean shutdown half-closes sends and drains late responses to EOF.
- **Transport seam** — generic `wireStream` / `opener` interfaces and test-only fake streams.
- **Tests** — ACK/Send races, immediate EOF exhaustion, stable idle recovery, ingestion, offsets, flush, close, batching, unacked retrieval, and replay.

## Verification

`go test -race ./...` and `go vet ./...` pass from `purego/`.

## Size

8 files changed, **1,496 additions**, 14 deletions.

## Stack

- 1/4 — primitives — #575
- 2/4 — basic core stream — this PR
- 3/4 — durability and lifecycle hardening — #577
- 4/4 — remaining integration — #573


purego/internal/stream/core.go
```diff
@@ -0,0 +1,547 @@
...
+			// Reset the lack-of-ack timer on every ack received.
+			if !lackTimer.Stop() {
+				select {
+				case <-lackTimer.C:
+				default:
+				}
+			}
+			lackTimer.Reset(cs.cfg.LackOfAckTimeout)
+
+			cs.wm.advance(offset)
```
teodordelibasic-db: An ack can make a record durable before its `Send` completes. `buffer.next` moves the item into `flight` before `stream.Send` returns, so a record blocked inside `Send` is already eligible for `wm.advance`. `Flush` or `WaitForOffset` can therefore return success for bytes that have not been transmitted.

Could we track the highest successfully sent offset separately for each connection, update it only after `Send` returns, and reject any ack above it before advancing the watermark?

```go
// runOnce: one watermark per connection.
lastSent := &atomic.Int64{}
lastSent.Store(-1)
go cs.sender(senderCtx, stream, errCh, lastSent)
go cs.receiver(stream, errCh, lastSent)

// sender: publish only after Send completes.
if err := stream.Send(it.payload); err != nil {
        errCh <- fmt.Errorf("stream: send offset %d: %w", it.offset, err)
        return
}
lastSent.Store(it.offset)

// receiver: validate before advancing durability.
if offset > lastSent.Load() {
        errCh <- fmt.Errorf("stream: ack %d exceeds last sent offset %d", offset, lastSent.Load())
        return
}
```

Could we add a wire fake that signals entry into `Send` but blocks its return, inject an ack for that offset, and assert that durability does not advance? Before the fix, `WaitForOffset` returns `nil`; afterward, the stream should fail with a protocol error.

purego/internal/stream/supervisor.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		runErr, healthy := cs.runOnce(ctx)
+
+		// ctx cancelled = clean exit via Close. Don't treat server-side EOF as clean.
+		if ctx.Err() != nil {
+			return
+		}
+
+		// Reset the retry budget after a successful Open.
```
teodordelibasic-db: An endpoint that repeatedly opens and immediately returns EOF never exhausts `RecoveryRetries`. Every successful `Open` returns `healthy=true`, the supervisor resets `failedAttempts` to zero, and the EOF path increments it back to one on every cycle.

Could we reset the recovery episode only after durable progress or after the connection remains stable for a defined interval?

```go
openedAt := time.Now()
startAck := cs.wm.current()

healthy := cs.wm.current() > startAck ||
        time.Since(openedAt) >= cs.cfg.RecoveryTimeout
return cause, healthy
```

Could we cover both sides of the contract: already-closed RPCs should exhaust the configured retry budget, while a stable idle connection that later closes should receive a fresh budget? The first case exceeds the retry budget before the fix, and the second prevents a progress-only definition from terminating healthy idle streams too aggressively.

purego/internal/stream/core.go
zlata-stefanovic-db: Send not returning does not prove the bytes were not transmitted. The server may already have received and processed them while the sender goroutine has not resumed (if an ack has arrived is the most likely case). The solution proposed will miss a case when an ack is very fast -> gets received before the Send() returns and signal an error.

It might be better to track the lastSent + currently sending records and hold the ACK temporarily for the currently sending ones, when Send() finishes we accept the held ACK.

purego/internal/stream/supervisor.go
zlata-stefanovic-db: Fixed

purego/internal/stream/core.go
```diff
@@ -0,0 +1,584 @@
...
+		// Invalid parameters cannot be fixed by reconnecting.
+		if errors.Is(err, transport.ErrInvalidParams) {
+			return wrapValidation(fmt.Errorf("stream: open: %w", err)), false
+		}
+		return fmt.Errorf("stream: open: %w", err), false
+	}
+	openedAt := time.Now()
+	startAck := cs.wm.current()
+
+	// Resend unacknowledged items on the new connection.
```
teodordelibasic-db, line 336: Recovery cannot work once the logical watermark has advanced. `EphemeralStream` offsets are connection-local physical positions: the Rust client creates a new physical offset generator for every sender task at [`rust/sdk/src/stream/grpc/sender.rs:29`](https://github.com/databricks/zerobus-sdk/blob/f6ee2394c648a81b22b4d295881538edd2d74764/rust/sdk/src/stream/grpc/sender.rs#L29) and translates cumulative physical acks back to logical offsets at [`rust/sdk/src/stream/grpc/receiver.rs:113`](https://github.com/databricks/zerobus-sdk/blob/f6ee2394c648a81b22b4d295881538edd2d74764/rust/sdk/src/stream/grpc/receiver.rs#L113). This core instead resends the buffered request with its original logical offset and passes the server's physical ack directly to `wm.advance`.

For example, if logical offset `0` was acked before a disconnect and logical offset `1` must be replayed, the new stream must send physical offset `0` and map ack `0` back to logical offset `1`. Here it sends offset `1`, and ack `0` cannot complete `WaitForOffset(1)`.

Could we keep logical offsets in the buffer, reset a separate physical counter on every successful `Open`, stamp the physical offset immediately before each `Send`, and translate cumulative physical acks through the ordered sent entries before advancing the logical watermark? The mapping could have this shape:

```go
physical := int64(len(sent))
req := cs.enc.withOffset(it.payload, physical)
sent = append(sent, it.offset)

// A cumulative physical ack identifies the corresponding logical item.
logical := sent[ackOffset]
cs.wm.advance(logical)
cs.buf.discardThrough(logical)
```

Could we also extend `TestCoreStreamRecoveryRequeuesUnacked` to ack logical offset `0`, disconnect with logical offset `1` unacked, assert that the first replay carries physical offset `0`, then ack physical offset `0` and verify that `WaitForOffset(1)` succeeds?

purego/internal/stream/core.go
```diff
@@ -0,0 +1,584 @@
...
+
+type sendEvent struct {
+	offset    int64
+	completed bool
+	err       error
+	consumed  chan struct{}
+}
+
+// runOnce operates one transport stream until a worker exits.
+// healthy reports whether Open succeeded.
```
teodordelibasic-db: The `healthy` comment is stale. It now means the connection made durable progress or remained open for `RecoveryResetAfter`, not merely that `Open` succeeded. Could we update the comment to match the retry-reset behavior?

teodordelibasic-db (commented):
Only one major comment.

purego/internal/stream/core.go
zlata-stefanovic-db: Fixed — renamed healthy to resetRecoveryBudget and updated the comment to say it is true only after durable progress or after the connection remains stable for RecoveryResetAfter.

purego/internal/stream/core.go
zlata-stefanovic-db: Fixed by separating lifetime logical offsets from connection-local physical offsets. The sender now resets a physical counter to 0 after every successful Open and stamps that physical offset immediately before Send. The receiver keeps an ordered physical-to-logical mapping, so cumulative physical ACKs advance the correct logical watermark and discard the correct buffered records.

The existing fast-ACK handling is preserved: an ACK for the currently sending physical offset waits until Send succeeds. I added recovery coverage where logical 0 is acknowledged, logical 1 and 2 are replayed as physical 0 and 1, and physical ACK 1 completes WaitForOffset(2). 

There is also coverage for a fast physical ACK 0 mapping to logical 1 after reconnect.

teodordelibasic-db (approved):
LGTM

PR #577, opened 2026-07-22
[PureGo] Ingestion core: durability, callbacks, retry classification (3/4)

### LINKED ISSUE #637: [PureGo] Internal setup budget expiry is misclassified as a terminal stream failure
An SDK-internal timeout expiring inside `Conn.Open` is classified as a terminal
stream failure, so a transient stall during setup permanently kills the stream
instead of being recovered. The supervisor's context can never carry a deadline,
so the `context.DeadlineExceeded` clause in `isRetryable` only ever fires for the
SDK's own budgets — never for a caller deadline.

## Where

`purego/internal/stream/supervisor.go`, in `isRetryable`:

```go
func isRetryable(err error) bool {
	if err == nil {
		return false
	}
	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
		return false
	}
	...
}
```

`isRetryable` is the supervisor's recovery gate: a `false` breaks the `supervise`
loop and takes the terminal-failure path, failing the watermark and reporting
every buffered record to `OnError`.

## Why a caller deadline can never reach it

The supervisor runs on its own cancel-only context, built from `context.Background()`
rather than derived from any caller context (`purego/internal/stream/core.go:204`):

```go
ctx, cancel := context.WithCancel(context.Background())
...
go cs.supervise(ctx)
```

That context has no deadline, and it is the context handed to `opener.Open`
(`core.go:366`). `Close` cancels it, but `supervise` already short-circuits on
that before classifying anything — `ctx.Err() != nil` is checked both at the top
of the loop (`supervisor.go:17-19`) and immediately after `runOnce`
(`supervisor.go:36-39`).

So by the time a `context.DeadlineExceeded` reaches `isRetryable`, it cannot be a
caller deadline or a `Close`. It is one of the SDK's own internal budgets applied
inside `transport.Conn.Open` (`purego/internal/transport/ephemeral_stream.go:76-97`):

- `defaultHeadersTimeout` — 15s, bounds `GetHeaders` (`raw_stream.go:26`)
- `defaultHandshakeTimeout` — 15s, bounds the create-stream handshake (`raw_stream.go:32`)

## Impact

A single slow token mint or a server that half-opens the stream for 15s is a
transient condition, but it is currently terminal:

- The stream fails permanently on the first occurrence. `RecoveryRetries` and
  `RecoveryBackoff` are never consulted, so a configured recovery budget provides
  no protection against the one failure mode it most obviously should cover.
- Because the failure is terminal, `HeadersProvider.Invalidate` never gets a
  second attempt to re-mint credentials, and buffered records are abandoned to
  `OnError` rather than resent on a fresh connection.
- It also makes the handshake race in #635 fatal rather than self-healing. When a
  server rejection is masked by the handshake deadline, the resulting
  `DeadlineExceeded` kills the stream outright. If internal budget expiry were
  retryable, the next attempt would surface the real `Unauthenticated`, trigger
  `Invalidate`, and re-mint — no fidelity fix required.

## Divergence from the Rust core

Rust deliberately treats a setup deadline as retryable. `DeadlineExceeded` is
absent from `UNRETRIABLE_STATUS_CODES` (`rust/sdk/src/errors.rs:136-144`), so
`is_retryable()` returns true for `CreateStreamError(DeadlineExceeded)` and the
supervisor reconnects. Rust's handshake budget (`recovery_timeout_ms`) is also
15s, so the two SDKs hit the same condition and then classify it in opposite
directions. Purego is the stricter and, I think, the wrong one.

## Proposed fix

Stop treating `context.DeadlineExceeded` as terminal in `isRetryable`:

```go
	// Close cancels the supervisor context, which supervise checks directly; a
	// deadline here can only be one of the SDK's internal setup budgets, which is
	// transient and worth reconnecting for.
	if errors.Is(err, context.Canceled) {
		return false
	}
```

Retrying stays bounded by `RecoveryRetries` and spaced by `RecoveryBackoff`, so
this cannot spin. Keeping the `context.Canceled` clause is still worthwhile as
defence against a `Close` that races the classification.

If a blanket change feels too broad, the narrower alternative is to tag the
internal budgets with a sentinel cause and key on it — `defaultHeadersTimeout`
already does this via `context.WithTimeoutCause` with
`authctx.ErrHeadersBudgetExceeded` (`ephemeral_stream.go:83`), and
`defaultHandshakeTimeout` could take a matching cause. `isRetryable` would then
treat a tagged budget expiry as retryable and leave every other deadline
terminal. That is more code for the same outcome given the context analysis
above, but it is explicit about intent and survives a future refactor that gives
the supervisor a deadline-bearing context.

## Testing notes

- A test asserting that a stream whose first `Open` exceeds the handshake budget
  reconnects (rather than going terminal) would pin this. `transport`'s
  `hangHandshake` fake server plus `SetDefaultHandshakeTimeout` already provide
  the ingredients.
- Existing coverage to check for regressions: `isRetryable`'s terminal path is
  what surfaces errors via `GetUnacked` / `OnError`, so any test asserting a
  terminal failure on a timeout will need updating — and is worth reviewing to
  confirm it was asserting intent rather than current behaviour.

zlata-stefanovic-db:
## Summary

**3 of 4** — hardens the basic ingestion stream from #576 with durability validation, asynchronous callbacks, recovery classification, bounded shutdown, and input limits.

The final lifecycle, byte-backpressure, and hot-path worker hardening is intentionally isolated in #573.

## Changes

- **Durability validation** — malformed, unknown, and future acknowledgements cannot advance the logical watermark.
- **Physical ACK mapping** — connection-local wire offsets map back to lifetime logical offsets across reconnects.
- **Send-state preservation** — acknowledgements racing `Send` are deferred until the send outcome is known.
- **Callbacks** — lossless range-coalesced dispatch runs outside stream workers and supports reentrant `Close`.
- **Unacked retention** — terminal and clean-close drains remain available through `GetUnacked`.
- **Pause handling** — sending parks while late acknowledgements drain before recovery.
- **Shutdown safety** — blocked sends and graceful drains are bounded, and unreachable offset waiters are released.
- **ACK timing** — lack-of-ack timing starts only after records enter flight.
- **Input bounds** — encoded payload and batch-count limits reject oversized requests before transport retries.
- **Backpressure concurrency** — capacity is reserved before offset assignment and expensive encoding runs outside the offset lock.
- **Recovery and authentication** — Open attempts are bounded, retry classification is preserved, recovery budgets reset after progress or a stable connection, and authentication rejection refreshes cached headers.

## Verification

- `go test -race ./...`
- `go vet ./...`

Run from `purego/`.

## Size

12 files changed, **2,893 additions**, 287 deletions.

## Stack

- 1/4 — primitives — #575
- 2/4 — basic core stream — #576
- 3/4 — durability, callbacks, and recovery — this PR
- 4/4 — lifecycle and performance hardening — #573


purego/internal/stream/core.go
```diff
@@ -340,52 +533,138 @@ func (cs *CoreStream[Req, Resp]) runOnce(ctx context.Context) (cause error, rese
...
+	for {
+		select {
+		case senderErr := <-senderExitCh:
+			senderExited = true
+			drainPause()
+			if !senderParked {
+				cause = senderErr
+				break waitLoop
+			}
+		case receiverErr := <-receiverExitCh:
```
teodordelibasic-db, line 701: A real receiver failure after a pause is replaced with the earlier `pauseSignal`. This turns an `Unauthenticated`, malformed-response, or transport failure into a retry that does not consume the recovery budget. It also bypasses `HeadersProvider.Invalidate`, so a rejected cached token can be reused indefinitely.

Could we preserve the pause only for a clean receiver exit and otherwise return `receiverErr`?

```go
if receiverErr != nil || observedPause == nil {
        cause = receiverErr
} else {
        cause = *observedPause
}
```

A focused regression can queue a positive-duration `CloseStreamSignal` followed by a malformed response through `fakeRPC.recvs`. Before the fix the terminal error is the pause; afterward it contains `"unusable server response"`.

purego/internal/stream/supervisor.go
```diff
@@ -107,24 +129,28 @@ func isRetryable(err error) bool {
...
-	}
+	// Honor self-classifying wrapped errors.
 	var re retryableError
 	if errors.As(err, &re) {
 		return re.IsRetryable()
 	}
+	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+		return false
+	}
+	// Treat remaining failures as transient.
```
teodordelibasic-db, line 173: `isRetryable` treats every wrapped gRPC status that lacks the private `IsRetryable` interface as transient. That includes `InvalidArgument`, `Unauthenticated`, `PermissionDenied`, `OutOfRange`, `Unimplemented`, and `NotFound`, so deterministic rejections consume the full backoff budget and can resend pending data. Initial authentication failures also repeat for the full budget after `Open` has already invalidated the credentials.

Could we classify the permanent status codes explicitly, while keeping initial authentication refresh as a separate one-shot exception?

```go
switch status.Code(err) {
case codes.InvalidArgument, codes.Unauthenticated, codes.PermissionDenied,
        codes.OutOfRange, codes.Unimplemented, codes.NotFound:
        return false
}
return true
```

The existing `TestIsRetryable` table is a direct seam for permanent and transient status cases. Each permanent case returns `true` before this change and `false` afterward.

purego/internal/stream/core.go
```diff
@@ -253,17 +403,27 @@ func (cs *CoreStream[Req, Resp]) Flush(ctx context.Context) error {
...
-// Batch items expand into their original records.
+// GetUnacked returns records that were ingested but never acknowledged, one
+// entry per record. A batched buffer item expands to all of its records (not
+// just the first), so no unacked record is silently dropped. It closes the
+// stream first (idempotent) to ensure the buffer is fully drained and no new
+// items are added.
 func (cs *CoreStream[Req, Resp]) GetUnacked() [][]byte {
 	cs.Close()
-	items := cs.buf.drain()
+	items := cs.takeRetainedUnacked()
```
teodordelibasic-db: `GetUnacked` consumes the recovery data on read. `takeRetainedUnacked` clears the retained slice and `buf.drain` clears the remaining buffer, so a second or concurrent caller gets an empty result. A diagnostic read can therefore remove the records before the actual retry or persistence path retrieves them. The Rust SDK keeps an immutable closed-stream snapshot and returns a clone on every call.

Could we consolidate the buffer into retained storage once after `Close`, then return a cloned snapshot without clearing it?

```go
cs.Close()
cs.termMu.Lock()
if items := cs.buf.drain(); len(items) > 0 {
        cs.retainedUnacked = append(cs.retainedUnacked, items...)
}
items := slices.Clone(cs.retainedUnacked)
cs.termMu.Unlock()
```

The decoded record bytes should also be cloned because the proto decoder currently returns slices backed by the retained payload. Extending `TestCoreStreamGetUnackedReturnsItems` to call `GetUnacked` twice and mutate the first result would cover both properties: the current code returns an empty second result, while the fixed code should return the original bytes again.

purego/internal/stream/core.go
```diff
@@ -16,46 +19,69 @@ const (
...
 	DrainTimeout time.Duration
+	// MaxPayloadBytes caps one encoded ingest request. Non-positive values use
+	// DefaultMaxPayloadBytes.
+	MaxPayloadBytes int
+	// MaxBatchRecords caps the number of records in one batch independently of
+	// byte size. Non-positive values use DefaultMaxBatchRecords.
+	MaxBatchRecords int
+	// CallbackTeardownTimeout bounds asynchronous callback worker teardown.
+	CallbackTeardownTimeout time.Duration
 	// StreamPausedMaxWait caps a server-requested pause.
```
teodordelibasic-db: `Config` cannot represent immediate recovery from a server pause. Both `0` and negative durations mean no client cap, while the Rust, existing Go, Python, and C++ SDKs distinguish an unset value from an explicit zero. If the public Pure Go options inherit this field, correcting the shape later will require an API change.

Could we preserve presence separately, for example:

```go
StreamPausedMaxWait *time.Duration

if cap := cs.cfg.StreamPausedMaxWait; cap != nil &&
        (wait <= 0 || *cap < wait) {
        wait = *cap
}
```

A focused test can keep one record in flight, send a positive pause duration, set an explicit zero cap, and assert that the next `Open` starts without waiting for the server duration. The existing pause fakes and opener counter provide that seam; the current representation cannot make the assertion pass.

purego/internal/stream/core.go
```diff
@@ -340,52 +533,138 @@ func (cs *CoreStream[Req, Resp]) runOnce(ctx context.Context) (cause error, rese
...
-func (cs *CoreStream[Req, Resp]) gracefulTeardown(stream wireStream[Req, Resp], errCh <-chan error) {
+// gracefulTeardown stops sending and drains responses to EOF.
+// DrainTimeout prevents shutdown from hanging.
+func (cs *CoreStream[Req, Resp]) gracefulTeardown(
+	stream wireStream[Req, Resp],
+	senderExited, receiverExited bool,
+	senderExitCh, receiverExitCh <-chan error,
+) {
+	if !senderExited {
+		timer := time.NewTimer(cs.cfg.DrainTimeout)
```
teodordelibasic-db: `DrainTimeout` is applied to two sequential waits, so `Close` can take almost `2 * DrainTimeout` if the sender exits just before the first timer and the receiver then consumes the second timer. The configuration and `Close` docs describe one bound for graceful shutdown.

Could we create one timer at the start of `gracefulTeardown` and share its channel across both phases?

```go
timer := time.NewTimer(cs.cfg.DrainTimeout)
defer timer.Stop()

// Use timer.C for both the sender-exit and receiver-drain selects.
```

purego/internal/stream/core_test.go
teodordelibasic-db: `Close` explicitly allows callback delivery to finish asynchronously, so reading `ackCount` immediately after `Close` returns races the dispatcher. With `go test -race ./internal/stream -run '^TestCoreStreamCloseDrainsGracefully$' -count=200`, this assertion failed 6 of 200 iterations in the reviewed checkout.

Could we wait for the callback condition instead?

```go
waitCondition(t, func() bool { return cb.ackCount() == 1 }, time.Second)
```

This still proves that the late acknowledgment was drained without assuming callback completion is part of `Close`.

purego/internal/stream/core.go
zlata-stefanovic-db: #573 (next stacked PR) addresses this already - it reworks runOnce for this: the wait loop always sets cause = receiverErr now (no more swapping in the pause), and a pause is only surfaced on a clean io.EOF - a real error always wins via handleTerminal. So your pause-then-malformed case ends with the real error, burns the recovery budget, and still calls Invalidate. Kept it in #573 since it rides on the sender/receiver channel refactor there.

purego/internal/stream/core_test.go
zlata-stefanovic-db: Agreed, that assertion races the dispatcher. Alreadyt fixed in #573 in the same way as your suggestion: waitCondition(t, func() bool { return cb.ackCount() == 1 }, time.Second).  I deferred it there because #573 contains the lifecycle/drain/callback teardown hardening.

purego/internal/stream/supervisor.go
zlata-stefanovic-db: Fixed, added transport.IsTerminalStatus(err) in transport/headers.go that classifies those permanent status codes (plus FailedPrecondition) as non-retryable.

purego/internal/stream/core.go
zlata-stefanovic-db: Fixed. Replaced takeRetainedUnacked with consolidateUnacked (in core.go): under termMu, it folds any still-buffered items into retainedUnacked once, then returns slices.Clone(...) without clearing. Repeated/concurrent calls all see the same snapshot - matching Rust's immutable closed-stream snapshot.

purego/internal/stream/core.go
zlata-stefanovic-db: Changed the field to *time.Duration: nil = no cap (honor the full server duration); a non-nil value caps the wait; an explicit durationPtr(0) = reconnect immediately.

TestCoreStreamPauseZeroCapReconnectsImmediately - keeps a record in flight, has the server request a 1-hour pause, sets an explicit zero cap, and asserts the stream reopens (openCount == 2) without waiting the server duration.

purego/internal/stream/core.go
zlata-stefanovic-db: Fix (in core.go): one shared timer created at the top of gracefulTeardown, and its channel is reused across both phases, so the whole graceful shutdown is bounded by a single DrainTimeout.

purego/internal/stream/core.go
```diff
@@ -238,86 +346,162 @@ func (cs *CoreStream[Req, Resp]) setServerID(id string) {
...
 
-// Flush waits for all currently queued records.
-// It applies DefaultFlushTimeout when ctx has no deadline.
+// Flush blocks until every record ingested so far is acknowledged by the
+// server. Returns nil once all are durable, or an error if the stream fails
+// or the wait budget expires.
+//
+// Flush always enforces Config.FlushTimeout as an upper bound, even when the
+// caller context has no deadline.
 func (cs *CoreStream[Req, Resp]) Flush(ctx context.Context) error {
```
teodordelibasic-db, line 465: `Close` calls `Flush`, which snapshots `lastEnqueued`, before publishing any closing state. While it waits for that target's acknowledgment, another goroutine can append the next offset and return success. Once the earlier target is acknowledged, `Close` cancels the supervisor and abandons the newly accepted record.

There is no valid sequential ordering for that interleaving: if `Close` wins, the ingest must fail; if the ingest wins, `Close` must include its offset in the durability boundary. Could we publish a closing state atomically with the flush-target snapshot, then recheck that state under the admission lock before appending? That would ensure every ingest either completes before the close target is captured or returns an error.

purego/internal/stream/supervisor.go
```diff
@@ -29,33 +50,48 @@ func (cs *CoreStream[Req, Resp]) supervise(ctx context.Context) {
...
 		}
 
 		if runErr == nil {
-			// EOF is a retryable disconnect.
+			// Server closed the stream cleanly (EOF). Treat as a retryable
+			// disconnect and recover.
 			runErr = fmt.Errorf("stream: server closed the stream")
 		}
 
+		var openErr *openFailure
```
teodordelibasic-db, line 94: When the first `Open` returns `Unauthenticated` or `PermissionDenied`, the transport invalidates the cached credentials. The supervisor then applies `IsTerminalStatus` and closes the stream immediately, so it never calls `GetHeaders` again to use refreshed credentials. This fixes repeated auth retries, but it also prevents the one refresh attempt that invalidation is meant to enable.

Could we add a one-shot exception for an initial auth rejection when recovery is enabled and the retry budget is nonzero? Authentication failures after that first retry, including live-stream and reconnect failures, should remain terminal.

A regression can make the first `Open` reject stale credentials and the second succeed, then separately verify that a second initial auth rejection is terminal. It should also assert that invalidation occurs exactly once per rejection.

purego/internal/stream/core.go
```diff
@@ -238,86 +346,162 @@ func (cs *CoreStream[Req, Resp]) setServerID(id string) {
...
+//
+// If flush cannot complete within FlushTimeout, Close proceeds with teardown and
+// any remaining records are abandoned (retrievable via GetUnacked, or reported
+// through the AckCallback's OnError). On a clean shutdown the live stream is
+// torn down gracefully (half-close, then drain any straggling acks to EOF,
+// bounded by DrainTimeout) so the server observes an orderly END_STREAM rather
+// than an abrupt reset; see gracefulTeardown.
 func (cs *CoreStream[Req, Resp]) Close() {
 	cs.closeOnce.Do(func() {
+		_ = cs.Flush(context.Background())
```
teodordelibasic-db: `Close` now promises to flush pending records, but `_ = cs.Flush(context.Background())` discards timeout and terminal errors. A caller cannot distinguish fully durable shutdown from teardown with abandoned records unless it performs a separate diagnostic read. The background context also gives the caller no way to abort a stalled close; with the default configuration the implicit flush can occupy the calling goroutine for five minutes.

Could we return and memoize the flush error while still completing teardown, so repeated `Close` calls observe the same result? If per-call cancellation is part of the intended API, `Close` also needs a caller context or a separate abort path.

Existing Go `Close() error` and Rust `close()` returning `Result` both surface flush failure after releasing resources, so returning the durability result would preserve the established SDK contract.

purego/internal/stream/core.go
```diff
@@ -201,6 +305,10 @@ func NewCoreStream[Req, Resp any](
 	ackMdl ackModel[Resp],
 	callback AckCallback,
 ) *CoreStream[Req, Resp] {
+	cfg = sanitizeConfig(cfg)
```
teodordelibasic-db, line 314: The constructor copies `Config` by value but retains the caller's `StreamPausedMaxWait` pointer. If the caller reuses or mutates that duration after construction, the write can race `effectivePauseWait` and change live stream behavior. This differs from `DescriptorProto`, which the same constructor snapshots before starting the supervisor.

Could we copy the pointed-to value before storing the configuration?

```go
if cfg.StreamPausedMaxWait != nil {
        cap := *cfg.StreamPausedMaxWait
        cfg.StreamPausedMaxWait = &cap
}
```

A deterministic regression can construct a stream with an explicit zero cap, mutate the caller's original duration before sending a long pause, and verify that the stream still reconnects immediately.

purego/internal/stream/core.go
zlata-stefanovic-db: Done, now Close() uses the same lock for marking the stream as closing and loading lastEnqueued.

purego/internal/stream/supervisor.go
zlata-stefanovic-db: Fixed: Recovery-enabled streams may retry exactly once when the first Open fails authentication.

purego/internal/stream/core.go
zlata-stefanovic-db: Fixed: Close now returns the flush error after completing cleanup.

purego/internal/stream/core.go
zlata-stefanovic-db: Fixed: the constructor copies the duration value into stream-owned storage. Later caller mutations cannot affect the stream.

purego/internal/stream/core.go
```diff
@@ -382,52 +652,144 @@ func (cs *CoreStream[Req, Resp]) runOnce(ctx context.Context) (cause error, rese
...
+		select {
+		case ps := <-pauseCh:
+			recordPause(ps)
+		default:
+		}
+	}
+waitLoop:
+	for {
+		select {
+		case senderErr := <-senderExitCh:
```
teodordelibasic-db, line 694: An `io.EOF` from `Send` can replace the RPC's authoritative status from `Recv`. Go gRPC uses `io.EOF` for a server-side send failure and exposes the actual authentication, schema, or protocol status through `Recv`. If the sender exit wins this select, cleanup waits for the receiver but discards its result, so the stream retries the generic EOF and can skip credential invalidation or resend records after a deterministic rejection.

Could we retain the receiver result during cleanup and prefer its non-cancellation status when the sender cause is `io.EOF`?

```go
receiverErr := <-receiverExitCh
receiverExited = true
if errors.Is(cause, io.EOF) &&
        receiverErr != nil &&
        !errors.Is(receiverErr, context.Canceled) {
        cause = receiverErr
}
```

A coordinated wire fake can make `Send` return `io.EOF`, then have `Abort` release `Recv` with `Unauthenticated`. The current path reopens without invalidating; the fixed path should invalidate once and terminate without another `Open`.

purego/internal/stream/core.go
```diff
@@ -363,13 +624,22 @@ type sendEvent struct {
...
+	openTimedOut := errors.Is(openCtx.Err(), context.DeadlineExceeded)
+	cancelOpen()
 	if err != nil {
+		openErr := &openFailure{cause: fmt.Errorf("stream: open: %w", err)}
 		// Invalid parameters cannot be fixed by reconnecting.
 		if errors.Is(err, transport.ErrInvalidParams) {
-			return wrapValidation(fmt.Errorf("stream: open: %w", err)), false
+			return wrapValidation(openErr), false
+		}
+		// A deadline that happens to expire alongside a permanent rejection must
```
teodordelibasic-db, line 637: The Open deadline can change an explicitly non-retryable provider error into a retryable timeout. The guard preserves permanent gRPC statuses, but a non-retryable OAuth `TokenError` has status `Unknown`, so it is wrapped in `openBudgetExceeded`. That outer wrapper wins the `IsRetryable` lookup and turns a permanent OAuth failure into repeated token requests.

Could we preserve any error that already implements the retryability contract and use `openBudgetExceeded` only for unclassified timeout errors?

```go
var classified retryableError
if openTimedOut && ctx.Err() == nil &&
        !transport.IsTerminalStatus(err) &&
        !errors.As(err, &classified) {
        return &openBudgetExceeded{cause: openErr}, false
}
```

A variant of `terminalTimeoutOpener` can return `classifiedError{retryable: false}` after its context expires. It currently consumes every retry; after the fix it should stop after one `Open`.

teodordelibasic-db (commented):
This one is a potential preexisting issue, we can address it in a separate PR, just leaving it here:

[`purego/internal/transport/raw_stream.go:201`](https://github.com/databricks/zerobus-sdk/blob/68e8865ebe98a359799bf82b4dc3ab8b07bd5374/purego/internal/transport/raw_stream.go#L201)

An authentication or validation rejection that arrives at the handshake deadline is discarded. The timeout branch cancels the RPC and waits for `Recv`, but ignores the result and always returns `DeadlineExceeded`. `Conn.Open` therefore cannot invalidate rejected credentials, and the stream can retry a permanent setup failure for the full recovery budget.

Could we inspect the reaped result and preserve a non-cancellation error before falling back to the handshake deadline?

```go
case <-hctx.Done():
        teardown()
        r := <-done
        if r.err != nil &&
                !errors.Is(r.err, context.Canceled) &&
                !errors.Is(r.err, context.DeadlineExceeded) &&
                !errors.Is(r.err, io.EOF) {
                return fmt.Errorf("await ready response: %w", r.err)
        }
        return fmt.Errorf("await ready response: %w", hctx.Err())
```

A transport regression can use an already-expired deadline and the existing teardown-gated `blockingRPC`, returning `Unauthenticated` when teardown releases `Recv`. The current code returns `DeadlineExceeded`; the fixed path should preserve the authentication status.

purego/internal/stream/core.go
zlata-stefanovic-db: Fixed with one tweak to your snippet: I wait for the receiver before stream.Close() instead of after, bounded by DrainTimeout. Otherwise the Close itself can turn the server's status into Canceled, so the fix would work in a fake but be flaky for real. Safe to wait first since a failed send doesn't end the receiver — it's still sitting in Recv.

purego/internal/stream/core.go
zlata-stefanovic-db: Fixed with one refinement though: skipping the wrapper for any self-classified error also catches errors that are only non-retryable because our own deadline cut them short — a TokenError from an interrupted mint looks exactly like that, and openCtx always has RecoveryTimeout on it. That's the timeout the wrapper is for, so a blanket guard would kill the stream on one slow token endpoint. So deniesRetry excludes a DeadlineExceeded cause.

teodordelibasic-db (approved):

PR #582, opened 2026-07-23
[Rust] Refresh one stale credential during initial stream setup

teodordelibasic-db:
## Summary

Initial stream setup can now use one existing recovery retry after an `Unauthenticated` or `PermissionDenied` response. The SDK invalidates the rejected headers provider before retrying, allowing providers that support invalidation—including built-in OAuth—to refresh a stale credential.

This applies to standard gRPC and Arrow Flight streams. The retry shares the configured recovery budget and setup deadline; a repeated authentication rejection remains terminal. Global auth-error classification, reconnect behavior, and mid-stream handling are unchanged.

Go and .NET integration tests assert the one-shot setup retry without claiming their callback providers refresh credentials.

## Testing

- Rust: `make check` and `make test`
- Go: `make check` and `make test` against the branch FFI
- Mock-server coverage for success after one retry, repeated rejection, disabled/exhausted recovery, permanent errors, and stalled invalidation
- .NET wrapper coverage in CI


rust/sdk/src/arrow_stream.rs
```diff
@@ -490,26 +489,51 @@ impl ZerobusArrowStream {
...
-            headers_provider,
-            sdk_identifier,
-        )
-        .await?;
-
-        let result = Self::start_stream_connection(client, table_properties, options).await;
+        // Share one deadline across connection setup and auth-rejection invalidation.
+        // This preserves the original auth error if a custom provider stalls instead of
+        // reclassifying the attempt as a retryable setup timeout.
+        let attempt_deadline =
```
nikolaobradovic-db: Large timeout values can panic.
let attempt_deadline = tokio::time::Instant::now() + Duration::from_millis(recovery_timeout_ms);
The same issue exists at connection.rs. The public API accepts any u64; unchecked Instant addition can panic on supported platforms instead of returning an error. Use overflow-safe timeout(duration, ...) and calculate remaining invalidation time with saturating_sub.

rust/tests/src/rust_tests.rs
```diff
@@ -186,6 +191,403 @@ mod stream_initialization_and_basic_lifecycle_tests {
...
+    async fn test_initial_auth_invalidation_timeout_preserves_one_retry_limit(
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        setup_tracing();
+
+        let (mock_server, server_url) = start_mock_server().await?;
+        mock_server
+            .inject_responses(
+                TABLE_NAME,
+                vec![
+                    MockResponse::Error {
```
nikolaobradovic-db, line 597: MockResponse::Error {
    status: tonic::Status::unauthenticated("first rejected credential"),
    delay_ms: 0,
},
Both Rust transport tests return auth errors immediately, so using a fresh full timeout for invalidation would still pass. Delay the setup rejection and assert the entire attempt remains bounded by one timeout.

nikolaobradovic-db (dismissed):

nikolaobradovic-db (approved):

rust/sdk/src/arrow_stream.rs
teodordelibasic-db: Fixed both Arrow and gRPC paths using duration-based timeouts and saturating_sub for the remaining invalidation budget. Added regression tests for u64::MAX; the old code panics and the fixed code passes.

rust/tests/src/rust_tests.rs
teodordelibasic-db: Updated both transport tests to delay the setup rejection and assert the full attempt remains within the shared timeout budget. Verified they fail when invalidation receives a fresh timeout and pass with the shared budget.

PR #584, opened 2026-07-23
[Rust] Introduce arrow_schema_from_uc_columns_with_options to allow clients to have variant fields with an extension

flaviofcruz:

Introduces `arrow_schema_from_uc_columns_with_options` which has the option to annotate VARIANT fields with the canonical arrow.parquet.variant extension marker so that downstream consumers can recover which fields are variants.

The zerobus server will strip the marker so that ingestion can work with it.

## How is this tested?
Unit tests

### DIFF
--- a/rust/sdk/src/schema.rs
+++ b/rust/sdk/src/schema.rs
@@ -685,7 +685,9 @@ fn sanitize_message_name(name: &str) -> String {
 ///
 /// Notable Arrow choices, all dictated by the Databricks Arrow Flight server:
 /// `STRING` / `DECIMAL` → `LargeUtf8`, `VARIANT` →
-/// `Struct<metadata: LargeBinary not null, value: LargeBinary not null>`,
+/// `Struct<metadata: LargeBinary not null, value: LargeBinary not null>`
+/// annotated with the canonical `arrow.parquet.variant` extension marker so
+/// downstream consumers can recover which fields are variants,
 /// `BINARY` → `LargeBinary`,
 /// `DATE` → `Date32`, `TIMESTAMP` → `Timestamp(Microsecond, Some("UTC"))`,
 /// `TIMESTAMP_NTZ` → `Timestamp(Microsecond, None)`, `ARRAY<T>` → `List` with
@@ -733,11 +735,7 @@ fn uc_column_to_arrow_field(column: &UcColumn) -> Result<arrow_schema::Field, Sc
         complex_type_to_arrow_field(&column.name, &complex, column.nullable)
     } else {
         let p = parse_uc_top_level_type(&column.type_name)?;
-        Ok(arrow_schema::Field::new(
-            &column.name,
-            map_primitive_to_arrow(p),
-            column.nullable,
-        ))
+        Ok(primitive_arrow_field(&column.name, p, column.nullable))
     }
 }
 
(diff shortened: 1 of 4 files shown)


rust/sdk/src/schema.rs
```diff
@@ -733,11 +735,7 @@ fn uc_column_to_arrow_field(column: &UcColumn) -> Result<arrow_schema::Field, Sc
         complex_type_to_arrow_field(&column.name, &complex, column.nullable)
     } else {
         let p = parse_uc_top_level_type(&column.type_name)?;
-        Ok(arrow_schema::Field::new(
-            &column.name,
-            map_primitive_to_arrow(p),
-            column.nullable,
-        ))
+        Ok(primitive_arrow_field(&column.name, p, column.nullable))
```
teodordelibasic-db: This can break callers that use `arrow_schema_from_uc_columns` for the stream schema but construct their `RecordBatch` schema separately, which was valid in `v2.4`. For example, an existing application can look like this:

```rust
let stream_schema = Arc::new(arrow_schema_from_uc_columns(&uc_columns)?);
let stream = sdk
    .stream_builder()
    .table(TABLE_NAME)
    .oauth(CLIENT_ID, CLIENT_SECRET)
    .arrow(Arc::clone(&stream_schema))
    .build_arrow()
    .await?;

// The application builds the same documented physical VARIANT shape itself.
let variant_fields = Fields::from(vec![
    Field::new("metadata", DataType::LargeBinary, false),
    Field::new("value", DataType::LargeBinary, false),
]);
let batch_schema = Arc::new(Schema::new(vec![Field::new(
    "attrs",
    DataType::Struct(variant_fields.clone()),
    false,
)]));
let attrs = StructArray::new(
    variant_fields,
    vec![
        Arc::new(LargeBinaryArray::from(vec![b"m".as_slice()])),
        Arc::new(LargeBinaryArray::from(vec![b"v".as_slice()])),
    ],
    None,
);
let batch = RecordBatch::try_new(batch_schema, vec![Arc::new(attrs)])?;

stream.ingest_batch(batch).await?;
```

In `v2.4`, both schemas have the same physical `Struct<metadata, value>` field and no marker, so the final call succeeds. With this change, `stream_schema` has the `arrow.parquet.variant` metadata but `batch_schema` does not. Arrow includes field metadata in schema equality, so `ingest_batch` rejects the batch locally with:

```text
RecordBatch schema does not match stream schema
```

The physical data is still identical, and the request never reaches the server, so the server-side marker stripping cannot help.

Could we keep the released helpers' unmarked output in `v2.5` and add an opt-in marked conversion API? That preserves existing callers while allowing Vector and new callers to request the marker explicitly. For example:

```rust
pub fn arrow_schema_from_uc_columns(
    columns: &[UcColumn],
) -> Result<Schema, SchemaError> {
    arrow_schema_from_uc_columns_impl(columns, false)
}

pub fn arrow_schema_from_uc_columns_with_variant_extensions(
    columns: &[UcColumn],
) -> Result<Schema, SchemaError> {
    arrow_schema_from_uc_columns_impl(columns, true)
}
```

A test can assert the released helper stays unmarked and the opt-in helper marks top-level and nested `VARIANT` fields.

I would say this is a niche case, but it is breaking across a minor version bump. We can discuss further...

flaviofcruz: It's reasonable to keep the same behavior and you are correct that `ingest_batch` would fail under that specific scenario (though server will be OK). I added arrow_schema_from_uc_columns_with_options which accepts an options struct. Arrow has other variants and we might want to introduce other schema transformation options.

Also added test to exercise the specific edge case you raised.

rust/sdk/src/schema.rs
```diff
@@ -676,33 +676,73 @@ fn sanitize_message_name(name: &str) -> String {
...
+///
+/// ```
+/// # use databricks_zerobus_ingest_sdk::schema::ArrowSchemaOptions;
+/// let mut options = ArrowSchemaOptions::default();
+/// options.annotate_variant_extension = true;
+/// ```
+#[cfg(feature = "arrow-flight")]
+#[derive(Clone, Debug, Default)]
+#[non_exhaustive]
+pub struct ArrowSchemaOptions {
```
teodordelibasic-db, line 692: nit while resolving conflicts: Could we add a short example of using this opt-in API to `rust/README.md` and the Arrow example?

teodordelibasic-db (dismissed):
LGTM

teodordelibasic-db (approved):
LGTM

PR #585, opened 2026-07-24
[Go] Add application user-agent support

teodordelibasic-db:

This adds backward-compatible application identification to the Go SDK:

- keeps the existing two-argument `NewZerobusSdk` constructor and adds
  `NewZerobusSdkWithOptions` plus `WithApplicationName`;
- sends `zerobus-sdk-go/1.3.0` by default and appends the configured application
  name for server-side attribution;
- constructs the Rust core through the additive C builder API, including the Go
  SDK identifier and application name;
- refreshes the bundled FFI archives for Linux AMD64/ARM64, macOS AMD64/ARM64,
  and Windows AMD64 from the security-scanned release build;
- enables thin LTO for release FFI builds and rejects archives at GitHub's
  100 MiB file limit or without the required builder exports; and
- updates godoc, the README/API reference, changelog, and examples.

The Rust core previously supplied its own SDK identifier to Go requests, so the
service could not distinguish Go clients or attribute them to an application.
The C builder support already exists from the C++ SDK work; this PR wires the Go
wrapper to it without changing existing callers. Thin LTO is needed because the
current Rust dependency graph otherwise produces Linux static archives larger
than GitHub permits.

## How is this tested?

- Added default/custom user-agent tests for both protobuf streaming and Arrow
  Flight.
- Manually ran a local Go client using
  `WithApplicationName("shinkansen-go-ua-harness/1.0")` against a local
  Shinkansen server. Server logs showed:

  ```text
  user_agent=zerobus-sdk-go/1.3.0 shinkansen-go-ua-harness/1.0 tonic/0.14.5
  sdk_name=zerobus-sdk-go sdk_version=1.3.0
  ```

### DIFF
--- a/go/README.md
+++ b/go/README.md
@@ -188,26 +188,22 @@ func main() {
     }
     defer stream.Close()
 
-    // 4. Send record to server and get offset.
-    // IngestRecordOffset returns as soon as the record is queued; the offset is a
-    // logical sequence number assigned to this record. The server round-trip
-    // happens in the background.
-    offset, err := stream.IngestRecordOffset(`{"id": 1, "message": "Hello"}`)
-    if err != nil {
-        log.Fatal(err)
+    // 4. Queue records without waiting for each acknowledgment.
+    records := []string{
+        `{"id": 1, "message": "Hello"}`,
+        `{"id": 2, "message": "World"}`,
     }
-    log.Printf("Record queued for ingestion with offset %d", offset)
+    for _, record := range records {
+        if _, err := stream.IngestRecordOffset(record); err != nil {
+            log.Fatal(err)
+        }
+    }
 
-    // 5. Wait for the server to acknowledge the record is durably written.
-    //
-    // WaitForOffset confirms this one record, which is all this example needs.
-    // For real workloads, the idiomatic flow is to ingest many records in a loop
-    // and call stream.Flush() once at the end. See the "Idiomatic high-throughput
-    // flow" under Usage Guide → Ingest Data.
-    if err := stream.WaitForOffset(offset); err != nil {
+    // 5. Wait once for every queued record to be acknowledged.
+    if err := stream.Flush(); err != nil {
         log.Fatal(err)
     }
-    log.Println("Record confirmed by server")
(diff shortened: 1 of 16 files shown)


go/README.md
```diff
@@ -1019,12 +1034,39 @@ Main entry point for the SDK.
...
 ```
 
 Creates a new SDK instance.
 
 - `zerobusEndpoint` - Zerobus gRPC service endpoint
 - `unityCatalogURL` - Unity Catalog URL for OAuth token acquisition
+- `SdkOption` - An optional SDK-construction setting passed to
+  `NewZerobusSdkWithOptions`
+- `WithApplicationName` - Appends a trimmed application identifier to the
+  default wire user-agent, `zerobus-sdk-go/1.3.0`; blank names are ignored
```
andrijast-db: I don't think there a need to specify what's already in the user-agent - it can become stale. This concrete version can also be confusing.

go/tests/user_agent_test.go
```diff
@@ -0,0 +1,181 @@
...
+	return mockServer, serverURL
+}
+
+func TestDefaultUserAgentIdentifiesGoSdk(t *testing.T) {
+	mockServer, serverURL := startUserAgentTestServer(t)
+
+	openStreamForUserAgent(t, serverURL)
+
+	userAgent := mockServer.GetLastUserAgent()
+	if !strings.Contains(userAgent, "zerobus-sdk-go/1.3.0") {
```
andrijast-db: Will this need to be updated when bumping the version? If yes, can we reference version.go or remove the version specifier?

andrijast-db (dismissed):

go/tests/user_agent_test.go
teodordelibasic-db: Yup, nice suggestion, done.

go/README.md
teodordelibasic-db: Good point about staleness, changed to `zerobus-sdk-go/<version>`. I think it would be okay for clients to know their application name isn't the only component in the header.

andrijast-db (approved):

PR #587, opened 2026-07-24
[Rust] Enforce per-batch Arrow Flight acknowledgement deadlines

teodordelibasic-db:
### High-level behavior

For Arrow Flight streams, this PR changes `server_lack_of_ack_timeout_ms` from a timeout around each wait for a server response into a deadline derived from the oldest pending batch.

The old response timeout could be mostly consumed while the stream was idle because accepting the first batch did not restart the wait. Response traffic could also start a new timeout window even when it made no durable progress. As a result, recovery timing depended on when responses arrived instead of how long pending work had been waiting.

With this change, no acknowledgement timer runs while the pending set is empty. Once a batch becomes pending, the acknowledgement processor uses that batch's timestamp to calculate the deadline. Receiving a response does not replace that timestamp with a new timeout window.

### Implementation details

- Store an `enqueued_at` timestamp on every pending batch and derive the active acknowledgement deadline from the oldest pending batch.
- Notify the acknowledgement processor when the pending set changes from empty to non-empty, so a newly accepted batch gets a full timeout window after an idle period.
- Separate the idle wait from the pending-work wait. The idle path waits only for a response or newly pending work; the pending path races responses against the oldest batch's deadline.
- Let one ready response win an exact deadline tie for the same pending head and acknowledged-record state, then re-check that the current oldest batch is still expired before triggering recovery.
- Assign replayed batches a new timestamp while rebuilding them for the recovered connection.
- Document the Arrow Flight timeout behavior and its relationship with `max_inflight_batches` in the configuration docs and README.
- Clarify the acknowledgement-watermark changelog entry: delayed, duplicate, and backward watermarks are absorbed, while a forward watermark beyond the submitted-record count is rejected.

TODO: Graceful-close handling keeps its existing pause-deadline behavior in this change. Follow-up work will keep the oldest pending acknowledgement deadline active during graceful close.

## How is this tested?

- Unit test: a partial acknowledgement followed by a duplicate response does not replace the oldest batch's original deadline.
- Unit test: a valid acknowledgement already ready at expiry is applied before timeout recovery.
- Unit test: ready malformed or otherwise non-progressing responses cannot repeatedly win an expired-deadline tie.
- Arrow Flight E2E test: idle before ingestion, then verify the newly pending batch receives its full acknowledgement window.
- Arrow Flight E2E test: trigger timeout recovery, replay the pending batch, and verify the replay receives a fresh deadline and completes.

### DIFF
--- a/rust/sdk/src/arrow_stream.rs
+++ b/rust/sdk/src/arrow_stream.rs
@@ -19,10 +19,8 @@ use arrow_flight::{FlightClient, PutResult};
 use arrow_ipc::writer::IpcWriteOptions;
 use bytes::Bytes;
 use futures::{Stream, StreamExt};
-#[cfg(feature = "test-hooks")]
-use tokio::sync::Notify;
-use tokio::sync::{mpsc, watch, Mutex, OwnedSemaphorePermit, Semaphore};
-use tokio::time::{sleep, Duration};
+use tokio::sync::{mpsc, watch, Mutex, Notify, OwnedSemaphorePermit, Semaphore};
+use tokio::time::{sleep, Duration, Instant};
 use tokio_retry::strategy::FixedInterval;
 use tokio_retry::RetryIf;
 use tonic::metadata::MetadataValue;
@@ -45,6 +43,9 @@ use crate::ZerobusResult;
 /// Type alias for the batch sender channel, wrapped for thread-safe sharing.
 type BatchSender = Arc<Mutex<Option<mpsc::Sender<Result<RecordBatch, FlightError>>>>>;
 
+/// Stream of Arrow Flight responses carrying acknowledgments from the server.
+type FlightResponseStream = Pin<Box<dyn Stream<Item = Result<PutResult, FlightError>> + Send>>;
+
 /// Test-only barrier used to pause `reconnect` at a precise point — the new connection
 /// is established but pending ranges are not yet rebuilt — so a test can schedule a
 /// concurrent ingest or `close()`.
@@ -65,6 +66,11 @@ struct ReconnectRebuildBarrier {
 /// partial ack has landed before it proceeds.
 #[cfg(feature = "test-hooks")]
 type AckAppliedGate = Arc<Mutex<Option<Arc<Notify>>>>;
+
+/// Test-only gate: when armed, `process_acks` fires after observing an empty pending
+/// set and immediately before parking without an ack deadline.
+#[cfg(feature = "test-hooks")]
+type AckIdleGate = Arc<Mutex<Option<Arc<Notify>>>>;
 
 /// Test-only barrier that parks `close()` after the supervisor and sender are gone but
 /// before pending batches are finalized, allowing cancellation-safe teardown tests.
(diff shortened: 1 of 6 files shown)


rust/sdk/src/arrow_stream.rs
```diff
@@ -1297,54 +1335,117 @@ impl ZerobusArrowStream {
...
-                } else if all_acked {
-                    info!("All in-flight batches acknowledged during graceful close. Triggering recovery.");
-                    return Err(ZerobusError::StreamClosedError(tonic::Status::unavailable(
-                        "All in-flight batches acked during graceful close",
-                    )));
+            let force_ack_deadline = matches!(
+                oldest_pending,
+                Some((head_identity, ack_deadline))
+                    if expiry_tie_winner == Some(head_identity)
+                        && Instant::now() >= ack_deadline
```
nikolaobradovic-db: When both deadlines are expired and expiry_tie_winner is still unset, the graceful-close precheck can return before entering the response-first select!. A ready ACK is never parsed, so last_acked_records stays stale and recovery replays data the server already acknowledged.

The graceful-close precheck must defer to the ACK tie path whenever the oldest pending ACK deadline is expired but has not spent its one ready-response opportunity. Conceptually:

let ack_deadline_expired = oldest_pending
    .is_some_and(|(_, deadline)| Instant::now() >= deadline);
let grant_ack_tie = ack_deadline_expired && !force_ack_deadline;
if let Some(graceful_deadline) = pause_deadline {
    if Instant::now() >= graceful_deadline && !grant_ack_tie {
        return graceful_close_timeout();
    }
}
Then the existing response-first select! grants one ready ACK opportunity. If it removes the expired head, preserve its watermark; if it does not make progress, expiry_tie_winner forces the ACK deadline on the next iteration.

Add a unit regression with an expired pending batch, an equally expired graceful-close deadline, and a ready valid ACK; assert the ACK watermark/pending state updates before recovery is triggered.

rust/sdk/src/arrow_stream.rs
```diff
@@ -1297,54 +1335,117 @@ impl ZerobusArrowStream {
...
-            let result = if let Some(deadline) = pause_deadline {
-                tokio::select! {
-                    biased;
-                    _ = tokio::time::sleep_until(deadline) => {
-                        continue;
+            enum AckEvent {
+                Response(Option<Result<PutResult, FlightError>>),
+                AckDeadline,
+                GracefulCloseDeadline,
+            }
```
nikolaobradovic-db: Move this enum at the top of the process acks function.

rust/sdk/src/arrow_stream.rs
```diff
@@ -1297,54 +1335,117 @@ impl ZerobusArrowStream {
...
+
+            let event = match oldest_pending {
```
nikolaobradovic-db: Code is barely readable, create some helpers: 
async fn wait_while_idle(...) -> AckEvent
async fn wait_with_pending_deadline(...) -> AckEvent
Then the loop becomes:

let event = match oldest_pending {
    None => wait_while_idle(...).await,
    Some(deadline) => wait_with_pending_deadline(...).await,
};

rust/sdk/src/arrow_stream.rs
```diff
@@ -102,10 +105,26 @@ struct PendingBatch {
     /// Cumulative record count after this batch.
     /// Batch is fully acked when `acked_records >= end_record`.
     end_record: u64,
+    /// Time this batch most recently became pending on the active connection.
+    enqueued_at: Instant,
     /// Backpressure permit; dropping it frees one `max_inflight_batches` slot.
     _permit: OwnedSemaphorePermit,
 }
 
+type PendingBatchIdentity = (OffsetId, u64, u64);
```
nikolaobradovic-db: Convert to struct.

nikolaobradovic-db: Rename expiry_tie_winner to deadline_tie_granted_for.
Move AckEvent to module scope.
Extract the None and Some branches into wait_while_idle() and wait_with_pending_deadline().
Keep deadline tie handling entirely inside the pending helper.

rust/sdk/src/arrow_stream.rs
teodordelibasic-db: Moved to top of file, does that sound okay?

rust/sdk/src/arrow_stream.rs
teodordelibasic-db: Done, introduced several helpers, `process_acks` looks much better now IMO.

rust/sdk/src/arrow_stream.rs
teodordelibasic-db: Fixed, ack/graceful-close deadline ties now preserve a ready valid ack before recovery, with a regression test covering both deadlines expiring together.

rust/sdk/src/arrow_stream.rs
teodordelibasic-db: Done.

PR #588, opened 2026-07-27
[PureGo] Add stable stream IDs

### LINKED ISSUE #488: [Go][PureGo] Introduce a stable client-side stream ID separate from the rotating server ID
## Summary

The PureGo transport exposes `Stream.ID()`, which returns the server-assigned ID from the create-stream handshake. Recovery now creates a new transport stream and receives a new server ID, but one logical `CoreStream` survives across those connections.

PureGo therefore needs two explicit identifiers:

- a stable, client-generated logical stream ID for lifetime correlation;
- the current server-assigned connection ID for service-side debugging.

## Current behavior

- `confirmCreateStream` reads `CreateStreamResponse.stream_id`.
- `transport.Stream.ID()` returns that server ID.
- `CoreStream.runOnce` calls `Open` again during recovery, producing a new transport `Stream` and server ID each time.
- `CoreStream` currently exposes no ID of its own.
- Transport errors include the current server ID, but there is no stable identifier that connects errors before and after recovery.

The original issue assumed one transport object's atomic ID would be mutated during recovery. The implementation instead replaces the transport stream. This changes the mechanism, not the observability problem.

## Rust SDK behavior

Rust keeps the initial server ID in the outer `ZerobusStream`, while the supervisor uses a connection-local server ID that changes after recovery. This gives some outer operations a stable label, but it is still an initial server ID rather than an explicit client-generated logical ID, and recovery logs do not share one stable correlation key.

PureGo should make the distinction explicit rather than inheriting that ambiguity.

## Decided API

- `CoreStream.ID()` returns a client-generated ID created once in `NewCoreStream` and never changed.
- `CoreStream.ServerID()` returns the current server-assigned ID and changes after successful recovery.
- `transport.Stream.ServerID()` names the transport value accurately.
- Keep `transport.Stream.ID()` temporarily as a deprecated alias of `ServerID()` to avoid an unnecessary rename break.

## Implementation notes

- Generate the logical ID before the supervisor starts so it is immediately available and race-free.
- Store the current server ID atomically in `CoreStream`; update it only after `Open` succeeds.
- Extend the internal `wireStream` seam with `ServerID()` so reconnect handling remains protocol-agnostic.
- Keep logical and server IDs separate in tests and future logging fields.

## Acceptance criteria

- `ID()` is non-empty and unchanged across reconnects.
- `ServerID()` reflects the first handshake and changes after recovery.
- Concurrent ID reads pass the race detector.
- Failed Open attempts do not replace the last successful server ID.
- Existing transport callers can continue using deprecated `Stream.ID()` during migration.

## Part of

Pure-Go SDK effort — tracking issue #467; recovery foundation landed in #576.

zlata-stefanovic-db:

Add separate identifiers for a logical PureGo ingestion stream and its current server connection:

- `CoreStream.ID()` returns a client-generated ID that remains stable across recovery.
- `CoreStream.ServerID()` returns the most recently opened server-assigned ID.
- `transport.Stream.ServerID()` makes the connection-scoped meaning explicit.
- `transport.Stream.ID()` remains as a deprecated compatibility alias.
- The internal wire-stream seam now provides the server ID to the recovery layer.

Recovery replaces the underlying transport stream and receives a new server ID. Keeping that value as the only identifier makes one logical stream difficult to correlate across reconnects. The new logical ID is generated before the supervisor starts, while the server ID is updated only after a successful open.

The public PureGo API is still planned; this change establishes the internal semantics it can expose without conflating logical and connection identifiers.

Closes #488.

## How is this tested?

- Added tests for logical ID stability across recovery.
- Added tests for ID uniqueness and concurrent reads.
- Added coverage for server ID rotation and preservation after a failed reconnect.
- Ran `go test -race ./...` from `purego/`.
- Ran `go vet ./...` from `purego/`.

### DIFF
--- a/purego/README.md
+++ b/purego/README.md
@@ -26,8 +26,11 @@ Implemented packages:
   tokens (client-credentials flow) via `OAuthTokenProvider`, with per-table
   token caching and proactive refresh; `StaticHeadersProvider` returns a fixed
   header set for tests or externally managed credentials.
+- `internal/stream` — buffers records, tracks acknowledgements, and reconnects
+  interrupted streams. Each logical stream has a stable client ID, while
+  `ServerID` identifies its most recently opened server stream.
 
-Planned layers: ingest/ack state management, recovery, and the public API.
+Planned layer: the public API.
 
 ## Regenerating the protobuf bindings
 
--- a/purego/internal/stream/core.go
+++ b/purego/internal/stream/core.go
@@ -2,10 +2,14 @@ package stream
 
 import (
 	"context"
+	"crypto/rand"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/databricks/zerobus-sdk/purego/internal/transport"
@@ -152,6 +156,9 @@ type CoreStream[Req, Resp any] struct {
 	wm       *watermark
 	callback AckCallback
 
+	clientID string
+	serverID atomic.Pointer[string]
+
 	// ingestMu serializes offset assignment and enqueue.
 	ingestMu sync.Mutex
 	// nextOffset is protected by ingestMu.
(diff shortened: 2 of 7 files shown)


teodordelibasic-db (approved):
LGTM

PR #592, opened 2026-07-27
[Rust] Preserve acknowledgement deadlines during Arrow Flight graceful close

teodordelibasic-db:

This is the final change in the stack and depends on the per-batch acknowledgement-deadline change.

This PR preserves Arrow Flight acknowledgement deadlines while a server-initiated graceful close is in progress.

- Continue racing the oldest pending batch's acknowledgement deadline against the graceful-close deadline.
- Recover at whichever deadline expires first.
- Preserve the earliest close deadline when repeated close signals arrive instead of extending the recovery bound.
- Apply a valid acknowledgement already ready at either deadline before deciding to recover.
- Document that graceful close does not pause or extend pending acknowledgement deadlines.

Without this change, entering graceful close suppresses the acknowledgement timeout until the close wait ends. A server close signal can therefore extend the time an unacknowledged batch remains pending, and repeated signals can move that recovery bound further into the future.

## How is this tested?

- A unit test verifies an outstanding acknowledgement deadline still triggers recovery during a longer graceful-close window.
- A unit test verifies repeated close signals can shorten but never extend the original close deadline.
- A unit test verifies a valid acknowledgement already ready when the graceful-close or acknowledgement deadline expires wins the tie before recovery.

### DIFF
--- a/rust/NEXT_CHANGELOG.md
+++ b/rust/NEXT_CHANGELOG.md
@@ -21,6 +21,7 @@
 
 ### Bug Fixes
 
+- **Arrow Flight — acknowledgment deadlines remain active during graceful close** (Beta): a server close signal no longer pauses or extends an already-running batch deadline. Repeated close signals preserve the earliest recovery deadline, while a valid acknowledgement already ready at either deadline is applied before recovery is considered.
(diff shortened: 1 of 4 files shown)


teodordelibasic-db:
Tracked by #656. This draft was based on the obsolete effort/* stack; the graceful-rotation change will be reopened as a fresh rust/* PR stacked on the deadline-only #654.

PR #605, opened 2026-07-28
[Rust] Enforce per-batch Arrow Flight acknowledgement deadlines

teodordelibasic-db:
### High-level behavior

For Arrow Flight streams, this PR changes `server_lack_of_ack_timeout_ms` from a timeout around each wait for a server response into a deadline derived from the oldest pending batch.

The old response timeout could be mostly consumed while the stream was idle because accepting the first batch did not restart the wait. Response traffic could also start a new timeout window even when it made no durable progress. As a result, recovery timing depended on when responses arrived instead of how long pending work had been waiting.

With this change, no acknowledgement timer runs while the pending set is empty. Once a batch becomes pending, the acknowledgement processor uses that batch's timestamp to calculate the deadline. Receiving a response does not replace that timestamp with a new timeout window.

### Implementation details

- Store an `enqueued_at` timestamp on every pending batch and derive the active acknowledgement deadline from the oldest pending batch.
- Notify the acknowledgement processor when the pending set changes from empty to non-empty, so a newly accepted batch gets a full timeout window after an idle period.
- Separate the idle wait from the pending-work wait. The idle path waits only for a response or newly pending work; the pending path races responses against the oldest batch's deadline.
- Let one ready response win an exact deadline tie for the same pending head and acknowledged-record state, then re-check that the current oldest batch is still expired before triggering recovery.
- Assign replayed batches a new timestamp while rebuilding them for the recovered connection.
- Document the Arrow Flight timeout behavior and its relationship with `max_inflight_batches` in the configuration docs and README.
- Clarify the acknowledgement-watermark changelog entry: delayed, duplicate, and backward watermarks are absorbed, while a forward watermark beyond the submitted-record count is rejected.

TODO: Graceful-close handling keeps its existing pause-deadline behavior in this change. Follow-up work will keep the oldest pending acknowledgement deadline active during graceful close.

## How is this tested?

- Unit test: a partial acknowledgement followed by a duplicate response does not replace the oldest batch's original deadline.
- Unit test: a valid acknowledgement already ready at expiry is applied before timeout recovery.
- Unit test: ready malformed or otherwise non-progressing responses cannot repeatedly win an expired-deadline tie.
- Arrow Flight E2E test: idle before ingestion, then verify the newly pending batch receives its full acknowledgement window.
- Arrow Flight E2E test: trigger timeout recovery, replay the pending batch, and verify the replay receives a fresh deadline and completes.

### DIFF
--- a/rust/NEXT_CHANGELOG.md
+++ b/rust/NEXT_CHANGELOG.md
@@ -21,7 +21,8 @@
 
 ### Bug Fixes
 
-- **Arrow Flight — invalid acknowledgment watermarks are rejected** (Beta): ack progress is now monotonic, so delayed or duplicate responses cannot move the durable watermark backward. A response claiming more records than were actually submitted on the active connection is rejected without making buffered, unsent records appear durable.
+- **Arrow Flight — acknowledgment deadlines now start when work becomes pending** (Beta): no timer runs while the stream is idle. The timeout is an absolute deadline for the oldest pending batch: acknowledgments that leave that batch pending do not refresh it, replayed batches receive a fresh deadline on their recovered connection, and malformed or non-progressing responses cannot indefinitely postpone recovery. A valid acknowledgment already ready at expiry is applied before recovery is considered.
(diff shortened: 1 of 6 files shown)


teodordelibasic-db:
Superseded by #654, rebuilt on current main under the required rust/* branch and including ACK-deadline enforcement through graceful close and transport drain.

PR #636, opened 2026-07-29
[PureGo] Preserve server rejection on handshake timeout

### LINKED ISSUE #635: [PureGo] Handshake deadline discards the server's rejection status
A handshake that runs out of time reports `DeadlineExceeded` even when the server
already answered with a permanent rejection. The reaped `Recv` result is
discarded, so the real status never leaves the transport.

## Where

`purego/internal/transport/raw_stream.go`, in `rawStream.handshake`:

```go
select {
case <-hctx.Done():
	// Unblock recv, then wait so the goroutine can't outlive this call.
	teardown()
	<-done
	return fmt.Errorf("await ready response: %w", hctx.Err())
case r := <-done:
	if r.err != nil {
		return fmt.Errorf("await ready response: %w", r.err)
	}
	resp = r.resp
}
```

The timeout branch waits for the in-flight `Recv` so the goroutine cannot outlive
the call, which is correct, but `<-done` throws the result away and the branch
always returns `hctx.Err()`.

## Impact

When the server's `Unauthenticated`, `PermissionDenied`, or schema rejection
arrives in the same window as the handshake deadline, that status is lost, and
two things follow:

- `Conn.Open` cannot classify the failure, so `isAuthRejection(err)` is false and
  `HeadersProvider.Invalidate` is never called. Rejected credentials stay cached.
- The stream layer sees a plain timeout. `transport.IsTerminalStatus` reports
  false for it, so the supervisor treats a permanent setup failure as transient
  and retries it for the full recovery budget, resending records each time.

## Why it looks like an oversight rather than a deliberate choice

`handshake` already applies the opposite rule a few lines earlier, for the same
underlying reason — that gRPC hides the real status behind an opaque error on the
send path:

```go
// A gRPC Send that fails with io.EOF means the stream aborted server-side; the
// real status (e.g. an auth rejection) is on Recv, so fall through to it rather
// than returning the opaque EOF and losing the code isAuthRejection keys on.
if err := sendSetup(s.rpc); err != nil && !errors.Is(err, io.EOF) {
	return err
}
```

## Proposed fix

Inspect the reaped result and prefer a real status over the deadline:

```go
case <-hctx.Done():
	teardown()
	r := <-done
	if r.err != nil &&
		!errors.Is(r.err, context.Canceled) &&
		!errors.Is(r.err, context.DeadlineExceeded) &&
		!errors.Is(r.err, io.EOF) {
		return fmt.Errorf("await ready response: %w", r.err)
	}
	return fmt.Errorf("await ready response: %w", hctx.Err())
```

The cancellation and EOF exclusions matter because `teardown()` itself unblocks
`Recv`: whatever it produces as a result of our own cancellation must not be
mistaken for a server answer.

## Suggested test

A transport-level regression can use an already-expired handshake deadline
together with a teardown-gated blocking RPC, so `Recv` returns `Unauthenticated`
only once `teardown` releases it. Today the handshake reports
`DeadlineExceeded`; after the fix it must report the `Unauthenticated` status.

## Notes

Reported by @teodordelibasic-db while reviewing #577, and pre-existing on `main`
rather than introduced there — #577 does not touch `raw_stream.go`. It is the
transport-side counterpart of a bug already fixed inside #577, where an `Open`
that timed out alongside a permanent gRPC rejection was wrapped as retryable.
That fix keeps a terminal status from being masked, but it can only work if the
status survives the transport, which is what this issue is about.

### LINKED ISSUE #644: [PureGo] rawStream handshake masks server DeadlineExceeded as teardown artifact
## Summary
`rawStream.handshake` currently treats gRPC `codes.DeadlineExceeded` as a teardown artifact in `isTeardownArtifact`, which is too broad for the current `Conn.Open` path.

Because `streamCtx` is created with `context.WithoutCancel(...)` + `context.WithCancel(...)`, teardown cancels the RPC context but does not set a deadline. A teardown-provoked gRPC outcome should therefore be `codes.Canceled` on this path; `codes.DeadlineExceeded` can represent a real server status.

## Impact
When `hctx` has expired, `handshake` prefers `hctx.Err()` over any error classified as teardown artifact. With the current classifier, a genuine server `codes.DeadlineExceeded` can be dropped and replaced by `context.DeadlineExceeded`, changing how downstream logic classifies/retries the failure.

## Proposed fix
Narrow the gRPC teardown-artifact classification to `codes.Canceled` (while keeping bare-context and EOF checks for non-gRPC implementations).

## Test gap / follow-up
Add coverage that exercises a server `status.Error(codes.DeadlineExceeded, ...)` in the handshake-timeout arbitration path and asserts that the server status is preserved (not replaced by `hctx.Err()`).

## Context
Raised during review of [PR #636](https://github.com/databricks/zerobus-sdk/pull/636) as a preexisting, out-of-scope issue.

zlata-stefanovic-db:
## Summary
- preserve a terminal server status in the `raw_stream` handshake expiry path instead of always returning `hctx.Err()`
- decide the reported cause on `hctx`'s state rather than on which `select` case won, so a response and an expiry becoming ready together cannot make the error a coin flip
- call `teardown` whenever `hctx` has expired, including when the recv branch observed the result first
- exclude cancellation-shaped outcomes (`isTeardownArtifact`) so an RPC the SDK itself cancelled is not misread as a server rejection
- add regression coverage for both directions of the rule, plus a determinism check

## Why

When the server's rejection lands in the same window as the handshake deadline,
the status was discarded and `Open` returned a bare `DeadlineExceeded`. Two
things follow from losing the code:

- `isAuthRejection(err)` is false, so `HeadersProvider.Invalidate` is never
  called and rejected credentials stay cached.
- The stream layer sees a context error. `isRetryable` in
  `internal/stream/supervisor.go` treats `context.DeadlineExceeded` as
  **terminal**, so the supervisor breaks its recovery loop, fails the watermark,
  and abandons every buffered record to `OnError`. A permanent setup failure that
  a fresh token would fix instead kills the stream outright.

Note for anyone reading #635: its impact section predicts the opposite outcome —
that a `transport.IsTerminalStatus` returns false and the supervisor "retries for
the full recovery budget, resending records each time". That function does not
exist, and the stream layer does no gRPC-code classification at all. The real
behaviour is terminal-on-first-occurrence, which is worse than the issue
describes.

## Why the exclusion list is necessary rather than a heuristic

An earlier revision of this PR tried to replace the shape-based exclusions with
ordering: trust any result reaped *before* `teardown()`, on the theory that it
must predate the cancellation and therefore came from the server.
`TestOpenHonorsCallerDeadline` rejects that. `Conn.Open` installs a second
canceller — `context.AfterFunc(handshakeCtx, cancelStream)` — which fires as soon
as `handshakeCtx` expires, independently of `handshake`'s own `teardown`. A
`codes.Canceled` can already be sitting in `done` before `teardown` is called, so
arrival order says nothing about provenance and the error's shape is the only
signal available.

## Scope

This makes the reported cause faithful; it does not change the fact that an
expired `hctx` always fails the handshake, and it does not resurrect a readiness
response that arrived too late (`Open` already fails that case via its
post-handshake bridge check).

It also does not address the related defect in #637: because `isRetryable` treats
any `context.DeadlineExceeded` as terminal, an internal setup budget expiring is
itself misclassified as permanent. Fixing that would make this class of masking
self-healing — the next attempt would surface the real status and re-mint — which
is how the Rust core already behaves (`DeadlineExceeded` is absent from
`UNRETRIABLE_STATUS_CODES`). The two changes are complementary and independent.

Fixes #635

## Testing
- `gofmt -l .`, `go vet ./...`
- `go test -race -count=3 ./...` from `purego/`
- new: `TestRawStreamHandshakeTimeoutPrefersServerRejection` (rejection replaces
  `DeadlineExceeded`), `TestRawStreamHandshakeTimeoutKeepsDeadlineForTeardownArtifacts`
  (table over gRPC `Canceled`/`DeadlineExceeded`, bare context errors, and EOF),
  `TestRawStreamHandshakeTimeoutRejectionIsDeterministic` (200 iterations across
  both `select` interleavings)

### DIFF
--- a/purego/internal/transport/raw_stream.go
+++ b/purego/internal/transport/raw_stream.go
@@ -248,12 +248,14 @@ func (s *rawStream[Req, Resp]) handshake(
 // than something the server said, in which case the expiry that triggered the
 // cancellation is the truer cause. The canceller may be handshake's own teardown
 // or a caller-side bridge onto the same context (as Conn.Open installs), so this
-// keys on the shape of the error, not on who cancelled. gRPC reports a cancelled
-// RPC context as codes.Canceled; a non-gRPC bidiRPC may surface a bare context
-// error instead, and EOF carries no status at all.
+// keys on the shape of the error, not on who cancelled. On this path, gRPC
+// reports a cancelled RPC context as codes.Canceled; codes.DeadlineExceeded can
+// therefore represent a real server status and must not be masked as teardown
+// noise. A non-gRPC bidiRPC may surface a bare context error instead, and EOF
+// carries no status at all.
 func isTeardownArtifact(err error) bool {
 	switch status.Code(err) {
-	case codes.Canceled, codes.DeadlineExceeded:
+	case codes.Canceled:
 		return true
 	}
 	return errors.Is(err, context.Canceled) ||
--- a/purego/internal/transport/raw_stream.go
+++ b/purego/internal/transport/raw_stream.go
@@ -10,6 +10,8 @@ import (
 	"time"
 
 	"github.com/databricks/zerobus-sdk/purego/internal/authctx"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 // errHeadersBudgetExceeded tags the internal header-resolution budget (see
@@ -201,7 +203,15 @@ func (s *rawStream[Req, Resp]) handshake(
 	case <-hctx.Done():
 		// Unblock recv, then wait so the goroutine can't outlive this call.
 		teardown()
-		<-done
+		r := <-done
+		if r.err != nil &&
+			!errors.Is(r.err, context.Canceled) &&
+			!errors.Is(r.err, context.DeadlineExceeded) &&
+			!errors.Is(r.err, io.EOF) &&
+			status.Code(r.err) != codes.Canceled &&
+			status.Code(r.err) != codes.DeadlineExceeded {
+			return fmt.Errorf("await ready response: %w", r.err)
+		}
 		return fmt.Errorf("await ready response: %w", hctx.Err())
 	case r := <-done:
 		if r.err != nil {
--- a/purego/internal/transport/raw_stream.go
+++ b/purego/internal/transport/raw_stream.go
@@ -172,6 +172,10 @@ func (s *rawStream[Req, Resp]) gracefulClose(ctx context.Context) error {
 // outlives the call. teardown must cancel the RPC's context and is safe to call
 // more than once. Reusers (e.g. a future Arrow/Flight wireStream) must supply it.
 //
+// An hctx expiry always fails the handshake, but a terminal status the server
+// managed to send first is reported in place of hctx.Err(), so a rejection racing
+// the deadline still reaches the caller as a rejection.
+//
 // The hooks are protocol-specific: sendSetup writes the first message;
 // confirmReady validates the first response and returns the stream ID ("" if
 // none).
@@ -198,29 +202,39 @@ func (s *rawStream[Req, Resp]) handshake(
 		done <- recvResult{resp, err}
 	}()
 
-	var resp *Resp
+	var r recvResult
 	select {
 	case <-hctx.Done():
 		// Unblock recv, then wait so the goroutine can't outlive this call.
 		teardown()
-		r := <-done
-		if r.err != nil &&
-			!errors.Is(r.err, context.Canceled) &&
-			!errors.Is(r.err, context.DeadlineExceeded) &&
-			!errors.Is(r.err, io.EOF) &&
-			status.Code(r.err) != codes.Canceled &&
-			status.Code(r.err) != codes.DeadlineExceeded {
+		r = <-done
+	case r = <-done:
+	}
+
+	// An expired hctx always fails the handshake. Decide on hctx's state rather
+	// than on which case above won, because when both are ready select picks
+	// between them at random — keying off the winner would make the reported cause
+	// a coin flip.
+	if hctx.Err() != nil {
+		// The hctx branch above already tore down; this covers the recv branch, where
+		// the expiry landed a moment later. teardown tolerates repeat calls.
+		teardown()
+		// Cancelling the RPC makes recv report that cancellation instead of anything
+		// the server said, so such an outcome describes the expiry rather than a
+		// rejection. A terminal status the server did manage to send outranks
+		// hctx.Err(): Open keys credential invalidation on it, and the stream layer
+		// keys terminal-vs-transient recovery on it.
+		if r.err != nil && !isTeardownArtifact(r.err) {
 			return fmt.Errorf("await ready response: %w", r.err)
 		}
 		return fmt.Errorf("await ready response: %w", hctx.Err())
-	case r := <-done:
-		if r.err != nil {
-			return fmt.Errorf("await ready response: %w", r.err)
-		}
-		resp = r.resp
 	}
 
-	id, err := confirmReady(resp)
+	if r.err != nil {
+		return fmt.Errorf("await ready response: %w", r.err)
+	}
+
+	id, err := confirmReady(r.resp)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


purego/internal/transport/raw_stream.go
```diff
@@ -219,3 +243,20 @@ func (s *rawStream[Req, Resp]) handshake(
...
+// isTeardownArtifact reports whether err describes the RPC being cancelled rather
+// than something the server said, in which case the expiry that triggered the
+// cancellation is the truer cause. The canceller may be handshake's own teardown
+// or a caller-side bridge onto the same context (as Conn.Open installs), so this
+// keys on the shape of the error, not on who cancelled. gRPC reports a cancelled
+// RPC context as codes.Canceled; a non-gRPC bidiRPC may surface a bare context
+// error instead, and EOF carries no status at all.
+func isTeardownArtifact(err error) bool {
+	switch status.Code(err) {
+	case codes.Canceled, codes.DeadlineExceeded:
```
teodordelibasic-db: `codes.DeadlineExceeded` is not a teardown artifact on the current `Conn.Open` path. [`streamCtx`](https://github.com/databricks/zerobus-sdk/blob/40faa6346e65b2564c547044e4526bf2674ecd22/purego/internal/transport/ephemeral_stream.go#L101) is built with `context.WithoutCancel` and `context.WithCancel`, so it has no deadline. Calling `cancelStream` makes `grpc-go` report `codes.Canceled`; a `codes.DeadlineExceeded` returned by `Recv` can therefore be a server status.

Treating that status as teardown noise replaces it with `hctx.Err()`. This also changes recovery behavior today: [`isRetryable`](https://github.com/databricks/zerobus-sdk/blob/40faa6346e65b2564c547044e4526bf2674ecd22/purego/internal/stream/supervisor.go#L110) treats the context error as terminal, while a gRPC status with the same code does not satisfy `errors.Is(err, context.DeadlineExceeded)` and remains retryable. Could we narrow the gRPC case to `codes.Canceled` and keep the bare context checks below it?

```go
switch status.Code(err) {
case codes.Canceled:
        return true
}
```

The existing `grpc deadline exceeded` table case can then become a preservation case using `serverErr := status.Error(codes.DeadlineExceeded, "server deadline")` and asserting `errors.Is(got, serverErr)`. That assertion deterministically fails with the current classifier and passes after the switch is narrowed.

teodordelibasic-db (commented):
Pre-existing, unrelated to PR, at `purego/internal/transport/transport_test.go:529`:

The regression suite does not exercise the user-visible failure path where an auth rejection has already been captured, the open context expires, and `HeadersProvider.Invalidate` must still run. The new handshake tests cover arbitration with fakes, while this existing test covers only an immediate rejection without deadline pressure, so the `AfterFunc` bridge, gRPC wrapping, and invalidation are never exercised together.

Could we add a deterministic `Conn.Open` test with a client-stream interceptor that first captures the real `RecvMsg` result and then holds it until `openCtx.Done()` before returning it?

```go
func (s *heldRecvErrorStream) RecvMsg(m any) error {
        err := s.ClientStream.RecvMsg(m)
        <-s.release // openCtx.Done()
        return err
}
```

With the in-memory server returning `codes.Unauthenticated`, this guarantees the trailer is captured before the deadline but reaches the handshake only after `hctx` has expired. The test can assert that `Open` returns `codes.Unauthenticated`, not `context.DeadlineExceeded`, and that `Invalidate` is called once.

purego/internal/transport/raw_stream.go
zlata-stefanovic-db: Implemented both, thanks for the detailed suggestions

Added a deterministic Conn.Open regression using a held RecvMsg interceptor, so we now cover the “auth rejection captured first, open deadline expires before return” race. The test asserts we return Unauthenticated (not DeadlineExceeded) and still call Invalidate once.

Also narrowed teardown-artifact gRPC classification to codes.Canceled only. codes.DeadlineExceeded is now preserved as a possible server status, and the tests were updated to assert that preservation path.

teodordelibasic-db (approved):

PR #643, opened 2026-07-30
[PureGo] Add public zerobus API package

zlata-stefanovic-db:
## Summary

Adds the public `purego/zerobus` package tracked in #467 on top of the merged stream, transport, and authentication layers.

- add the public `SDK` API: `New`, `CreateStream`, `CreateStreamWithProvider`, and `Close`
- add the public `Stream` API for single and batch ingestion, context-aware admission, flushing, offset waiting, unacknowledged-record recovery, IDs, and lifecycle management
- use Protocol Buffers as the default record type, with explicit `WithProto` and `WithJSON` options
- expose recovery timeout/backoff, lack-of-ack timeout, buffering, batching, server-pause, TLS, application-name, and ack-callback controls
- share the OAuth token cache across streams and make `SDK.Close` terminate tracked streams before closing the gRPC connection
- eagerly validate stream arguments and application names, reject plaintext HTTP endpoints, and correctly normalize HTTPS, bare-host, resolver, and IPv6 endpoints
- expose structured retryability through `Error`, `Retryable(err)`, and the internal stream classifier
- add public documentation and tests built around pipelined ingestion followed by one `Flush`
- verify that the module builds and tests with `CGO_ENABLED=0`

## Notes

- Stream open is asynchronous by default: `CreateStream` validates arguments and returns a stream immediately while token resolution and the first handshake run in the background. Terminal connection failures surface through `Flush`, `WaitForOffset`, or the ack callback.
- Opt-in wait-ready creation and full first-open context semantics are added in follow-up PR #646.
- Public API examples are added in stacked PR #645, and the build-only release workflow is added in stacked PR #647.

## Test plan

- [x] `cd purego && go build ./...`
- [x] `cd purego && CGO_ENABLED=0 go test ./...`
- [x] `gofmt` clean

### DIFF
--- a/.github/workflows/ci-purego.yml
+++ b/.github/workflows/ci-purego.yml
@@ -98,3 +98,10 @@ jobs:
       # covers data races regardless of target OS; no Windows entry needed.
       - name: Run tests (race detector)
         run: go test -race ./...
+
+      # The primary SDK claim is CGO-free usage. Keep an explicit non-race test
+      # run with CGO disabled so CI catches accidental CGO regressions.
+      - name: Run tests (CGO disabled)
+        env:
+          CGO_ENABLED: "0"
+        run: go test ./...
--- a/.github/workflows/ci-purego.yml
+++ b/.github/workflows/ci-purego.yml
@@ -99,8 +99,6 @@ jobs:
       - name: Run tests (race detector)
         run: go test -race ./...
 
-      # The primary SDK claim is CGO-free usage. Keep an explicit non-race test
-      # run with CGO disabled so CI catches accidental CGO regressions.
       - name: Run tests (CGO disabled)
         env:
           CGO_ENABLED: "0"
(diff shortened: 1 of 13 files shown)


.github/workflows/ci-purego.yml
```diff
@@ -98,3 +98,10 @@ jobs:
       # covers data races regardless of target OS; no Windows entry needed.
       - name: Run tests (race detector)
         run: go test -race ./...
+
+      # The primary SDK claim is CGO-free usage. Keep an explicit non-race test
+      # run with CGO disabled so CI catches accidental CGO regressions.
```
elenagaljak-db: not sure we need this elaborate comment

elenagaljak-db (dismissed):
Looks good, the only thing we should pay extra attention to is to keep the same (where possible) behavior as the original go sdk. For example, in core.go some paths return 0 when erroring, where the original would return -1.

core.go:
```
if cs.isClosed() {
	if err := cs.terminalErr(); err != nil {
		return 0, err
	}
	return 0, errClosed
```

elenagaljak-db (approved):

PR #646, opened 2026-07-31
[PureGo] Add opt-in fail-fast create with wait-ready

### LINKED ISSUE #642: [PureGo] CreateStream does not fail fast on unrecoverable open errors (diverges from cgo/Rust)
### Summary

In the pure-Go SDK, `SDK.CreateStream` / `CreateStreamWithProvider` are pure wrappers over the async ingestion core (`internal/stream`). They validate arguments and return a ready `*Stream` immediately, while the first connection — token mint plus the create-stream handshake — proceeds in the background and recovers on failure.

As a result, an **unrecoverable open failure** (bad credentials, unknown/unauthorized table, invalid descriptor rejected by the server) does **not** surface from `CreateStream`. It surfaces later, on the first `Flush`, `WaitForOffset`, or through the ack callback's `OnError`.

This is currently documented on `CreateStream`:

```go
// The stream opens asynchronously: CreateStream validates its arguments and
// returns a ready stream immediately, while the first connection (token mint
// plus handshake) proceeds in the background and recovers on failure. A
// connection error that cannot be recovered — bad credentials, an unknown
// table — therefore surfaces on the first Flush, WaitForOffset, or ack
// callback, not from CreateStream.
```

### Divergence from the other SDKs

The cgo Go SDK (`go/`) and the Rust core both **block at stream-creation time**: Rust's `stream_builder().build().await` waits on a first-open result channel (`stream_init_result_tx` in `rust/sdk/src/stream/grpc/supervisor.rs`), and the cgo `CreateStream` inherits that blocking behavior through the FFI. So a caller migrating from those SDKs gets a create-time error there but a deferred (first-`Flush`) error in pure-Go for the same misconfiguration.

### Why it's deferred today

The pure-Go `internal/stream` core is deliberately fire-and-forget: `NewCoreStream` starts the supervisor and returns a stream that is "immediately ready for Ingest," with the first transport stream opened in the background (see the `NewCoreStream` doc comment). There is no first-open readiness signal to wait on, so the public layer has nothing to block on without a core change. Adding one was intentionally left out of the public-API PR to keep it scoped to the `zerobus` package and avoid touching the merged, reviewed core.

### Proposed follow-up

Add an opt-in create-time readiness gate so `CreateStream` can fail fast on unrecoverable open errors, matching the cgo/Rust SDKs:

1. Add a first-open readiness signal to `internal/stream` (mirroring Rust's `stream_init_result_tx`): the supervisor publishes the outcome of the first `Open` (success, or a terminal error) exactly once, e.g. via a `WaitReady(ctx) error` method on `CoreStream`.
2. Have the public `CreateStream`/`CreateStreamWithProvider` block on it, bounded by the caller's `ctx` (the parameter is already threaded through for this).
3. Keep it opt-in or behavior-compatible where it matters — e.g. a `WithWaitForReady`/`WithAsyncOpen` stream option — so the async-by-default design stays available for callers who prefer it.

Recoverable failures should still recover in the background as they do now; only a **terminal** first-open failure should short-circuit `CreateStream`.

### Notes

- Context: raised while building the public API layer (the `purego-api` work under #467).
- No behavior change is required for correctness — records are never lost; the error simply surfaces one call later. This is an ergonomics / cross-SDK-parity improvement.

zlata-stefanovic-db:
## Summary
- follow up on https://github.com/databricks/zerobus-sdk/issues/642
- add a first-open readiness signal to the pure-Go core stream and expose `WaitReady(ctx)` for first-open outcome waiting
- add `zerobus.WithWaitForReady()` so `CreateStream` and `CreateStreamWithProvider` can optionally block and fail fast on terminal first-open errors
- apply the creation context to token resolution, handshake, retry backoff, and every attempt before first-open succeeds in wait-ready mode
- preserve context values while detaching caller cancellation and deadlines in the default asynchronous mode
- detach creation-context cancellation after first-open succeeds so the live stream and later recovery remain owned by `Stream.Close` and `SDK.Close`
- document both context contracts and add coverage for value propagation, blocked token resolution and handshake cancellation, retry-backoff cancellation, and post-readiness stream lifetime

## Test plan
- [x] `cd purego && CGO_ENABLED=0 go test ./...`


purego/internal/stream/core.go
```diff
@@ -348,13 +359,40 @@ func NewCoreStream[Req, Resp any](
...
 	cs.lastEnqueued.Store(-1)
-	go cs.supervise(ctx)
+	go cs.supervise(lifecycleCtx, openingCtx)
 	return cs
 }
 
+// WaitReady waits for the first-open result. It returns nil once the stream has
+// opened at least once, or an error if first-open fails terminally or the
+// stream is closed before first-open succeeds.
+func (cs *CoreStream[Req, Resp]) WaitReady(ctx context.Context) error {
```
teodordelibasic-db, line 373: `WaitReady` can discard an already-published first-open result. If `readyCh` and `ctx.Done()` are both closed when this `select` runs, Go may choose the context branch. `createStream` then calls `terminate()` and returns a context error even though the handshake succeeded, or it masks a terminal server rejection already stored in `readyErr`.

Could we recheck `readyCh` before returning the context error so the completed first-open result takes precedence? For example:

```go
func (cs *CoreStream[Req, Resp]) WaitReady(ctx context.Context) error {
        readyResult := func() error {
                cs.readyMu.Lock()
                defer cs.readyMu.Unlock()
                return cs.readyErr
        }

        select {
        case <-cs.readyCh:
                return readyResult()
        case <-ctx.Done():
                select {
                case <-cs.readyCh:
                        return readyResult()
                default:
                        return ctx.Err()
                }
        }
}
```

zlata-stefanovic-db: Implemented, thanks for the catch
WaitReady now re-checks readyCh when ctx.Done() fires and prefers the already-published first-open result (readyErr) over returning ctx.Err(). Also added a regression test covering both published-success and published-terminal-error vs canceled context.

teodordelibasic-db (approved):
LGTM

PR #648, opened 2026-07-31
[Rust] Reject server over-acks

### LINKED ISSUE #638: [Rust] Receiver silently absorbs server over-ack and can wedge the stream
### Summary

The gRPC receiver silently absorbs a server ack that exceeds what the client has actually sent, and unconditionally advances its ack watermark to the bogus value. This can leave later legitimate records permanently unacknowledged, wedging the stream until the lack-of-ack timeout fires.

### Location

`rust/sdk/src/stream/grpc/receiver.rs:113-130`

```rust
for _offset_to_ack in (last_acked_offset + 1)..=durability_ack_up_to_offset {
    if let Ok(record) = landing_zone.remove_observed() {
        // ... fulfil oneshot / callback ...
    }
    // else: remove_observed() returned Err (nothing left observed) -> silently skipped
}
drop(map);
last_acked_offset = durability_ack_up_to_offset;   // advanced unconditionally
```

### Behavior

If the server sends `durability_ack_up_to_offset` larger than the highest physical offset actually observed/sent, `remove_observed()` returns `Err` for the excess iterations and the `if let Ok(...)` swallows it with no error. Worse, `last_acked_offset` is then set to the inflated server value regardless. Because subsequent acks are only processed in the range `(last_acked_offset + 1)..=`, any later real records whose physical offsets fall at or below the inflated watermark are never acked — their oneshots/callbacks never fire, and the stream eventually trips the server-lack-of-ack timeout.

Absent/negative offsets are only partially guarded: the fully-absent `durability_ack_up_to_offset` case is rejected (`receiver.rs:99-110`), but an out-of-range positive value is not.

### Why it matters

A buggy or misbehaving server (or a protocol desync after recovery) causes silent data-durability confusion rather than a clean, observable failure. Records reported to the user as pending are never resolved.

### Suggested fix

Validate each ack against the highest in-flight offset and treat an over-ack as a fatal protocol violation that tears the stream down (letting the supervisor decide whether to reconnect), rather than advancing the watermark past what was sent.

### Reference: how the pure-Go core handles this

The pure-Go reimplementation validates every ack against the in-flight set and fails fast on an over-ack:

`purego/internal/stream/core.go:1044-1050`
```go
highest, ok := cs.buf.highestInFlight()
if !ok || offset > highest {
    return fmt.Errorf("stream: server ack offset %d exceeds highest in-flight offset %d", offset, highest)
}
```

It also treats malformed/negative-offset acks as fatal (`purego/internal/stream/ackmodel.go:44-50`).

Pybsama:

Fixes #638.

This change makes the Rust gRPC receiver reject malformed cumulative durability acknowledgements before they mutate client state. It:

- tracks the highest physical offset handed to each active gRPC connection;
- synchronizes outbound channel handoff and sent-watermark publication, so an observed request blocked on channel capacity is not considered sent;
- rejects negative and beyond-sent ACK offsets before removing records or completing oneshots/callbacks;
- treats duplicate and regressive cumulative ACKs as no-ops, preserving a monotonic watermark; and
- retains an explicit invariant failure if an otherwise valid ACK cannot be applied to the observed queue.

A landing-zone observed count is not a safe upper bound because `observe()` happens before the bounded outbound channel handoff. A connection-local sent watermark closes that concurrency window and is reset together with the physical offset generator on recovery.

## How is this tested?

- `cargo test -p databricks-zerobus-ingest-sdk` (162 unit tests plus 24 passing doc tests; 9 doc tests ignored by existing annotations)
- `cargo test -p tests --test rust_tests` (93 integration tests)
- deterministic capacity-one sender test proving an observed request blocked in `reserve()` does not advance the sent watermark
- mock gRPC regressions for a single record and a five-record batch, both receiving ACK 1 after only physical offset 0 was sent
- receiver regressions for negative, duplicate, regressive, valid, and beyond-sent ACKs, including the `1 -> 0 -> 2` sequence
- `make check`
- `git diff --check`

### DIFF
--- a/rust/NEXT_CHANGELOG.md
+++ b/rust/NEXT_CHANGELOG.md
@@ -21,6 +21,11 @@
 
 ### Bug Fixes
 
+- **Rust gRPC — reject server over-acks**: cumulative durability
+  acknowledgements that exceed the highest request sent on the active
+  connection now fail the stream before any record is reported durable or the
+  acknowledgement watermark advances.
+
 - **Arrow Flight — invalid acknowledgment watermarks are rejected** (Beta): ack progress is now monotonic, so delayed or duplicate responses cannot move the durable watermark backward. A response claiming more records than were actually submitted on the active connection is rejected without making buffered, unsent records appear durable.
 
 ### Documentation
--- a/rust/sdk/src/stream/grpc/receiver.rs
+++ b/rust/sdk/src/stream/grpc/receiver.rs
@@ -12,13 +12,34 @@ use tokio::time::Duration;
 use tokio_util::sync::CancellationToken;
 use tracing::{error, info, instrument, span, Level};
 
-use super::types::{CallbackMessage, OneshotMap, RecordLandingZone};
+use super::types::{CallbackMessage, OneshotMap, RecordLandingZone, SentOffsetWatermark};
 use super::{ZerobusStream, STREAM_TEARDOWN_DRAIN_TIMEOUT_MS};
 use crate::databricks::zerobus::ephemeral_stream_response::Payload as ResponsePayload;
 use crate::databricks::zerobus::{
     CloseStreamSignal, EphemeralStreamResponse, IngestRecordResponse,
 };
 use crate::{OffsetId, StreamConfigurationOptions, ZerobusError, ZerobusResult};
+
+fn validate_ack_offset(
+    ack_offset: OffsetId,
+    last_acked_offset: OffsetId,
+    highest_sent_offset: OffsetId,
+) -> ZerobusResult<bool> {
+    if ack_offset < 0 {
+        return Err(ZerobusError::InvalidStateError(format!(
+            "Server ack offset {ack_offset} is negative"
+        )));
+    }
+    if ack_offset <= last_acked_offset {
+        return Ok(false);
+    }
+    if ack_offset > highest_sent_offset {
+        return Err(ZerobusError::InvalidStateError(format!(
+            "Server ack offset {ack_offset} exceeds highest sent offset {highest_sent_offset}"
+        )));
+    }
+    Ok(true)
+}
 
 impl ZerobusStream {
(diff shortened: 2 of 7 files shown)


Pybsama:
Revalidated this change against current `main` (`122572a3598d0172dadfe4d90680ef4602a7c141`, 56 commits after the public base). The public branch remains unchanged at the original Verified + DCO head `2ef51f9bd864ae8b35c72a556b7daa8e85ac04bd` and is currently conflicting.

Current-main evidence:
- The four production files have the same stable patch-id as the public commit: `0ad60c65b58976e976dcb6674fbd73714b32d547`.
- The two over-ack integration tests fail on unmodified current main (0 passed / 2 failed) and pass after the port (2/2).
- `make check` passes (workspace fmt + Clippy with `-D warnings`).
- SDK tests pass: 181 unit tests; 24 doc tests pass and 9 remain ignored.
- The main Rust integration suite passes: 93/93.
- `cargo test --workspace -- --test-threads=1` passes across the workspace.

One current-main OAuth timing test, `hung_refresh_falls_back_to_cached_token`, fails when the 20-test OAuth binary runs in parallel on both an unmodified `main` worktree and the patched worktree (19/20), while the same test passes alone on both sides. The serialized workspace run passes, and this change does not touch the OAuth files.

I also tested composition with adjacent PR #792 (`a5b6b332499086fa25c2b8734dfaf50fce6d9ae2`). The combined tree passes `make check`, both over-ack tests, the duplicate-ACK timeout test from #792, and the full combined Rust integration suite (94/94). Only a successfully applied forward ACK resets the #792 progress deadline; duplicate/regressive ACKs remain no-ops and invalid ACKs terminate before state mutation.

An independent concurrency review found 0 Critical and 0 Important issues and gave GO. The conflict-resolved tree is staged locally only; I have not pushed a new commit or made a new DCO/signature attestation, so this PR remains Draft.

PR #649, opened 2026-07-31
[Rust] Fix Arrow Flight rotation H2 resets

teodordelibasic-db:

Before this PR:

1. Zerobus sent a close signal that started a bounded grace period for the active Arrow Flight connection.
2. The SDK paused new sends, but decided whether it had finished waiting by checking whether the entire pending-batch collection was empty.
3. That collection mixed records already submitted on the closing connection with records accepted after the pause and buffered for replay.
4. The closing connection could never acknowledge those buffered records, so they could unnecessarily consume the grace period.
5. Recovery then dropped the bidirectional Flight call without first completing its request body and draining its response.
6. Tonic consequently cancelled the live HTTP/2 stream, producing `RESET(CANCEL)` on the server even when reconnect and replay later succeeded.

After this PR:

1. The SDK receives a server close signal and atomically pauses sends while snapshotting the record watermark submitted on the active connection.
2. It waits only for acknowledgments up to that connection-local target, bounded by the advertised grace and the configured ACK-wait cap.
3. Records accepted after the pause remain buffered for replay and do not extend the old connection's ACK target.
4. The SDK stops the request body at a `FlightData` boundary, emits request EOF, and drains late acknowledgments, peer status, or response EOF under the same bounded rotation deadline.
5. It reconnects and replays only the unacknowledged suffix when recovery is enabled. With recovery disabled, it still cleans up the active transport and leaves unacknowledged batches retrievable.
6. Permanent peer or protocol errors observed during rotation retain precedence over later retryable transport status.

To achieve this, the PR:

- Adds `RequestBodyControl` to half-close the active request without dropping the HTTP/2 stream.
- Uses one narrow rotation state machine: open, waiting for the submitted ACK snapshot, then draining.
- Reserves up to 500 ms of a normal server grace period for request EOF and response drain. A zero configured ACK wait skips only ACK waiting; bounded transport cleanup still runs.
- Applies acknowledgments received during the drain before reconnecting, which avoids replaying data that became durable after request EOF.
- Preserves only the permanent-error precedence required on the active rotated connection.
- Updates the Rust configuration documentation, README, changelogs, and C FFI field documentation. The FFI signature and ABI are unchanged.

This is intentionally scoped to normal server-initiated rotation. Explicit close during recovery and teardown of incomplete replacement requests retain their existing best-effort behavior and are follow-up work.

## How is this tested?

Added focused tests:

- `test_server_rotation_half_closes_before_reconnect`: the active request reaches EOF without a pre-EOF reset before replay starts.
- `test_server_rotation_without_recovery_still_half_closes`: cleanup still runs without reconnecting, and the unacknowledged batch remains retrievable.
- `test_zero_ack_wait_still_half_closes`: a zero ACK wait skips only ACK waiting and still performs bounded transport cleanup.
- `test_post_eof_ack_is_applied_before_replay`: an acknowledgment received after request EOF prevents unnecessary replay.
- `test_post_eof_permanent_error_is_preserved`: a permanent peer status received after request EOF is returned unchanged.
- `test_rotation_drain_timeout_falls_back_to_reconnect`: an open response cannot prevent bounded recovery.
- `test_invalid_close_ack_still_half_closes_request`: an invalid acknowledgment remains terminal even if a later retryable status arrives.
- `test_zero_server_grace_uses_bounded_local_drain`: a zero advertised grace still gets a bounded best-effort cleanup attempt.

Manual before/after comparison of the published Rust SDK 2.5.0 and these changes, verified from server-side logs:

- Before: the server observed one rotation-window `RESET(CANCEL)` before the post-rotation flush.
- After: the server observed a clean request `END_STREAM` followed by response `END_STREAM`, no rotation-window reset, successful recovery, and a completed post-rotation flush.

### DIFF
--- a/rust/sdk/src/arrow_stream.rs
+++ b/rust/sdk/src/arrow_stream.rs
@@ -1723,41 +1723,27 @@ impl ZerobusArrowStream {
                         }
                     };
 
-                    if ack.is_close_signal() {
+                    if ack.is_close_signal() && matches!(&rotation, RotationState::Open) {
                         let server_duration_ms = ack.close_stream_duration_ms.unwrap_or(0);
-                        let (new_ack_deadline, new_drain_deadline) = Self::rotation_deadlines(
+                        let (ack_deadline, drain_deadline) = Self::rotation_deadlines(
                             server_duration_ms,
                             options.stream_paused_max_wait_time_ms,
                         );
-
-                        match &mut rotation {
-                            RotationState::Open => {
-                                let target_records = Self::pause_and_snapshot_submitted(
-                                    &ingest_mutex,
-                                    &is_paused,
-                                    &submitted_records,
-                                )
-                                .await;
-                                rotation = RotationState::WaitingForAcks {
-                                    target_records,
-                                    ack_deadline: new_ack_deadline,
-                                    drain_deadline: new_drain_deadline,
-                                };
-                                info!(
-                                    server_duration_ms,
-                                    target_records, "Server requested graceful stream rotation"
-                                );
-                            }
-                            RotationState::WaitingForAcks {
-                                ack_deadline,
-                                drain_deadline,
-                                ..
-                            } => {
-                                *ack_deadline = (*ack_deadline).min(new_ack_deadline);
-                                *drain_deadline = (*drain_deadline).min(new_drain_deadline);
-                            }
-                            RotationState::Draining(_) => unreachable!(),
-                        }
+                        let target_records = Self::pause_and_snapshot_submitted(
+                            &ingest_mutex,
+                            &is_paused,
+                            &submitted_records,
+                        )
+                        .await;
+                        rotation = RotationState::WaitingForAcks {
+                            target_records,
+                            ack_deadline,
+                            drain_deadline,
+                        };
+                        info!(
+                            server_duration_ms,
+                            target_records, "Server requested graceful stream rotation"
+                        );
                     }
 
                     if ack.ack_up_to_records > 0 {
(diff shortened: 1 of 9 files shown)


rust/sdk/src/arrow_stream.rs
```diff
@@ -1364,198 +1452,337 @@ impl ZerobusArrowStream {
...
+                                    target_records,
+                                    ack_deadline: new_ack_deadline,
+                                    drain_deadline: new_drain_deadline,
+                                };
+                                info!(
+                                    server_duration_ms,
+                                    target_records, "Server requested graceful stream rotation"
+                                );
+                            }
+                            RotationState::WaitingForAcks {
```
nikolaobradovic-db: Could we calculate these deadlines only when transitioning from Open? The server emits a single close signal (close_stream_tx.take()), so supporting duplicate signals by tightening deadlines adds unnecessary complexity. We can ignore subsequent close signals while still processing any ACK watermark they carry below:


if ack.is_close_signal() && matches!(&rotation, RotationState::Open) {
    let (ack_deadline, drain_deadline) = Self::rotation_deadlines(
        ack.close_stream_duration_ms.unwrap_or(0),
        options.stream_paused_max_wait_time_ms,
    );
    let target_records = Self::pause_and_snapshot_submitted(
        &ingest_mutex,
        &is_paused,
        &submitted_records,
    )
    .await;
    rotation = RotationState::WaitingForAcks {
        target_records,
        ack_deadline,
        drain_deadline,
    };
}

nikolaobradovic-db (dismissed):

rust/sdk/src/arrow_stream.rs
teodordelibasic-db: Makes sense, done.

nikolaobradovic-db (approved):

PR #651, opened 2026-07-31
[ArrowFlight] Refactor Arrow SDK for better readability/maintability.

nikolaobradovic-db:

- Split the monolithic Arrow Flight stream into focused API, connection, ACK, supervisor, and batch modules.
- Added cohesive Supervisor and AckProcessor runtime owners, typed rotation deadlines, and tighter internal visibility.
- Encapsulated connection and pending-batch state behind ownership-oriented APIs.
- Fixed sender lifetime during initial supervisor handoff and recovery: replacement senders are published only after replay succeeds, while failed or cancelled replay promptly closes incomplete request channels.
- Updated module documentation and internal references. No public API changes.

## How is this tested?

Existing tests, pure refactor expect for small sender lifetime fix.


rust/sdk/src/stream/arrow/mod.rs
```diff
@@ -0,0 +1,1040 @@
...
+
+const LOG_TARGET: &str = module_path!();
+
+type BatchSender = Arc<Mutex<Option<mpsc::Sender<Result<RecordBatch, FlightError>>>>>;
+
+/// Test-only barrier used to pause `reconnect` at a precise point — the new connection
+/// is established but pending ranges are not yet rebuilt — so a test can schedule a
+/// concurrent ingest or `close()`.
+#[cfg(feature = "test-hooks")]
+type ReconnectRebuildGate = Arc<Mutex<Option<ReconnectRebuildBarrier>>>;
```
teodordelibasic-db, line 58: Eh minor nit and not sure, but WDYT about groupingdefinitions of test hook types in a separate file?

teodordelibasic-db (commented):
Looks great. IMO it would be good to unify with folder structure of gRPC and create `stream/arrow` instead of this separate `arrow_stream` folder, not sure if there is a reason not to do that. Also we have a couple of remaining files in `src/` named `arrow_*.rs`, we can look into finding an appropriate name for those as well.

rust/sdk/src/stream/arrow/mod.rs
nikolaobradovic-db: I think it should stay as is.

nikolaobradovic-db:
> Looks great. IMO it would be good to unify with folder structure of gRPC and create `stream/arrow` instead of this separate `arrow_stream` folder, not sure if there is a reason not to do that. Also we have a couple of remaining files in `src/` named `arrow_*.rs`, we can look into finding an appropriate name for those as well.

Done.

teodordelibasic-db (approved):

PR #654, opened 2026-08-03
[Rust][Arrow] Make ACK deadlines pending-relative

### LINKED ISSUE #655: [Rust][Arrow] Make ACK deadlines pending-relative
## Summary

For Arrow Flight streams, `server_lack_of_ack_timeout_ms` is currently tied to response waits rather than to submitted work. Idle time can consume most of a newly submitted batch's timeout window, while responses without sufficient durable progress can defer recovery.

Make the timeout an absolute deadline derived from the oldest batch pending on the active connection.

## Scope

- Add an enqueue timestamp to each pending batch.
- Run no ACK timer while no submitted batch is pending.
- Wake the ACK processor when work becomes pending after an idle period.
- Derive the active deadline from the oldest submitted batch's timestamp.
- Do not refresh that deadline on responses or partial ACK progress.
- Re-check the same pending head under the pending lock before failing, closing the ACK/deadline race.
- Let one already-ready response win an exact expiry tie for a pending head, without allowing repeated responses to defer that same expired head.
- Refresh timestamps when batches are rebuilt for replay on a replacement connection.
- Update configuration documentation, the Rust README, and `NEXT_CHANGELOG.md`.

## Acceptance criteria

- A deterministic paused-time regression idles for most of the configured timeout, submits a batch, delays its ACK by less than one full batch-relative timeout, and succeeds.
- The existing true-timeout behavior remains covered.
- Duplicate, zero-progress, and partial responses cannot extend an expired pending head indefinitely.
- A replayed batch receives a full timeout on the replacement connection.
- Deadline expiry is checked against the same oldest pending batch under lock.

## Non-goals

- ACK-deadline behavior during server-requested graceful rotation; tracked by #656.
- Coordination of an explicit client `close()` with recovery; tracked by #657.
- ACK watermark validation and monotonic progress, which were addressed by #591.

teodordelibasic-db:

For Arrow Flight streams, make `server_lack_of_ack_timeout_ms` an absolute deadline derived from the oldest batch pending during normal stream operation.

The previous response-relative timeout could consume most of a future batch window while the stream was idle, and responses without enough durable progress could start a new timeout window. Recovery timing therefore depended on response timing rather than how long submitted work had remained unacknowledged.

This PR:

- timestamps each pending batch and wakes the ACK processor when newly submitted work becomes pending;
- waits without an ACK timer while no submitted batch is pending;
- derives the active deadline from the oldest submitted batch and re-checks the same head under the pending lock before failing;
- allows one already-ready response to win an exact expiry tie without allowing repeated responses to defer the same expired head;
- preserves pending timestamps while rebuilding replay state, then refreshes all replayed batches together after replay completes so ACK processing receives a full timeout window on the replacement connection;
- validates configured ACK and recovery deadlines against the platform monotonic clock and caps server-advertised graceful-rotation periods at one year;
- wakes recovery when a closed request sender leaves accepted work buffered but not submitted, while allowing an already-ready ACK or server status to take precedence over the local send failure; and
- documents the pending-relative timeout behavior and its relationship with `max_inflight_batches`.

## How is this tested?

- `test_ack_timeout_starts_when_batch_becomes_pending`: verifies idle time does not consume a future batch ACK window.
- `ready_ack_wins_expired_deadline_tie`: verifies a ready ACK wins an exact deadline tie.
- `expired_head_gets_only_one_ready_response`: verifies repeated ready responses cannot defer the same expired head.
- `deadline_recheck_requires_same_pending_head`: verifies a fired deadline cannot fail a replacement head after an ACK race and returns the pending count from the validated snapshot.
- `replay_rebuild_preserves_pending_timestamp` and `replay_deadlines_share_completion_timestamp`: verify replay rebuild does not start deadlines early and all rebuilt batches share the replay-completion timestamp.
- `test_replay_ack_deadline_starts_after_replay_completes`: verifies slow multi-batch replay does not consume the recovered connection ACK window.
- `test_closed_request_sender_starts_recovery`: verifies an accepted but unsubmitted batch triggers recovery and is replayed when the request sender closes.
- `request_send_error_is_retryable_unavailable`: verifies the request-send failure retains its retryable transport classification.
- `ready_ack_is_applied_before_reported_request_send_failure`: verifies buffered durable progress is applied before recovery starts.
- `ready_server_error_wins_reported_request_send_failure`: verifies a real permanent server status is not masked by synthetic `Unavailable`.
- `test_unrepresentable_recovery_timeout_is_rejected`, `test_unrepresentable_ack_timeout_is_rejected`, and `test_unrepresentable_runtime_recovery_deadline_is_rejected`: verify unrepresentable configured and runtime deadlines return explicit errors.
- `unrepresentable_ack_deadline_is_rejected` and `server_rotation_grace_is_capped`: verify internal ACK deadline validation and the graceful-rotation cap.

Fixes #655.

### DIFF
--- a/rust/README.md
+++ b/rust/README.md
@@ -805,6 +805,15 @@ Also accepts `0` or `no`.
 | `ack_callback` | `Option<Arc<dyn AckCallback>>` | `None` | Optional callback for acknowledgment notifications |
 | `callback_max_wait_time_ms` | `Option<u64>` | `None` | Maximum time to wait for callback processing to complete after closing the stream (`None` = wait indefinitely, `Some(x)` = wait up to `x` ms) |
 
+For Arrow Flight streams *(Beta)*, `server_lack_of_ack_timeout_ms` is an
+absolute limit on how long the oldest batch may remain pending during normal
+stream operation, not an inactivity timeout. No timer runs while the stream is
+idle. A batch's deadline starts when it becomes pending; responses and partial
+acknowledgments do not pause or extend it. Replayed batches receive a fresh
+deadline on the recovered connection. Configure this timeout together with
+`max_inflight_batches` so the server can acknowledge a full allowed backlog
+within the timeout.
+
 
 **Example:**
 
--- a/rust/README.md
+++ b/rust/README.md
@@ -814,6 +814,10 @@ deadline on the recovered connection. Configure this timeout together with
 `max_inflight_batches` so the server can acknowledge a full allowed backlog
 within the timeout.
 
+Arrow stream construction rejects `recovery_timeout_ms` and
+`server_lack_of_ack_timeout_ms` values whose deadlines cannot be represented by
+the platform monotonic clock. Server-advertised graceful-rotation periods are
+capped at one year.
 
 **Example:**
 
--- a/rust/README.md
+++ b/rust/README.md
@@ -810,7 +810,8 @@ absolute limit on how long the oldest batch may remain pending during normal
 stream operation, not an inactivity timeout. No timer runs while the stream is
 idle. A batch's deadline starts when it becomes pending; responses and partial
 acknowledgments do not pause or extend it. Replayed batches receive a fresh
-deadline on the recovered connection. Configure this timeout together with
+deadline after the full replay completes and acknowledgment processing can
+resume on the recovered connection. Configure this timeout together with
 `max_inflight_batches` so the server can acknowledge a full allowed backlog
 within the timeout.
 
--- a/rust/NEXT_CHANGELOG.md
+++ b/rust/NEXT_CHANGELOG.md
@@ -8,6 +8,11 @@
 
 ### Bug Fixes
 
+- **Arrow Flight — acknowledgment deadlines are pending-relative** (Beta): no
+  timer runs while a stream is idle. During normal stream operation, each batch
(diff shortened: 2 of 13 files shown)


rust/README.md
```diff
@@ -805,6 +805,20 @@ Also accepts `0` or `no`.
...
+acknowledgments do not pause or extend it. Replayed batches receive a fresh
+deadline after the full replay completes and acknowledgment processing can
+resume on the recovered connection. Configure this timeout together with
+`max_inflight_batches` so the server can acknowledge a full allowed backlog
+within the timeout.
+
+Arrow stream construction rejects `recovery_timeout_ms` and
+`server_lack_of_ack_timeout_ms` values whose deadlines cannot be represented by
+the platform monotonic clock. Server-advertised graceful-rotation periods are
+capped at one year.
```
nikolaobradovic-db, line 821: Why one year?

nikolaobradovic-db (approved):

rust/README.md
teodordelibasic-db: No real reason, should probably be brought down much lower if we want to guard the client against a bad server.

nikolaobradovic-db: It should be capped at no more than 1 minute.

PR #666, opened 2026-08-03
[PureGo] Add dynamic proto stream with UC schema fetch

zlata-stefanovic-db:
## Summary
- Add end-to-end dynamic proto ingestion path in PureGo with `CreateDynamicProtoStream`, including UC schema fetch, UC-schema-to-descriptor conversion, and JSON-to-protobuf runtime conversion.
- Introduce new internal modules (`internal/ucschema`, `internal/schema`, `internal/dynamicproto`) with deterministic mapping logic, typed errors, and focused unit tests.
- Add SDK options for dynamic schema fetch behavior/caching, dynamic stream JSON ingest helpers, README quickstart, example app, and changelog entries.

## Test plan
- [x] Validate converter success/failure coverage in `internal/dynamicproto/converter_test.go`
- [x] Validate UC fetch error handling in `internal/ucschema/fetcher_test.go`
- [x] Validate UC schema conversion parity cases in `internal/schema/schema_test.go`
- [x] Validate dynamic descriptor caching/stream wiring checks in `zerobus/dynamic_proto_test.go`

## Notes
- This is a stacked PR on top of `purego-api-wait-ready-open` to keep scope focused.
- Existing unrelated untracked docs in the working tree were intentionally excluded.


purego/examples/dynamic/single/main.go
```diff
@@ -0,0 +1,54 @@
...
+
+	stream, err := openStream(sdk, cfg)
+	if err != nil {
+		log.Fatalf("create dynamic stream: %v", err)
+	}
+	defer stream.Close()
+
+	now := config.NowMicros()
+	records := [][]byte{
+		[]byte(config.MakeOrderJSON(1, "Alice Smith", "Wireless Mouse", 2, 25.99, "pending", now)),
```
andrijast-db: nit: `NowMicros` and `MakeOrderJSON` are more utils than config

purego/examples/dynamic/single/main.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"context"
+	"log"
+
+	"github.com/databricks/zerobus-sdk/purego/examples/config"
+	"github.com/databricks/zerobus-sdk/purego/zerobus"
+)
+
+func openStream(sdk *zerobus.SDK, cfg config.Settings) (*zerobus.DynamicProtoStream, error) {
+	return sdk.CreateDynamicProtoStream(context.Background(), cfg.TableName,
```
andrijast-db: It's pretty handy that this fetches schema direct from UC. But do we want to give more control to the user here?

For example, how about including an option to build the descriptor manually in code? I could see some use cases where someone would like to construct proto dynamically, but without making a call to UC.

purego/examples/dynamic/single/main.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		log.Fatalf("create dynamic stream: %v", err)
+	}
+	defer stream.Close()
+
+	now := config.NowMicros()
+	records := [][]byte{
+		[]byte(config.MakeOrderJSON(1, "Alice Smith", "Wireless Mouse", 2, 25.99, "pending", now)),
+		[]byte(config.MakeOrderJSON(2, "Bob Johnson", "Mechanical Keyboard", 1, 89.99, "shipped", now)),
+		[]byte(config.MakeOrderJSON(3, "Carol Williams", "USB-C Hub", 3, 45.00, "delivered", now)),
```
andrijast-db: Any plans to support anything other than JSON? For example, pass on plain go structs. It would probably be better perf-wise.

purego/internal/dynamicproto/converter.go
```diff
@@ -0,0 +1,67 @@
+// Package dynamicproto converts JSON payloads into protobuf bytes at runtime.
```
teodordelibasic-db, line 1: Main concern here is performance, as it was in Rust, due to converting JSON into Protobuf at runtime. We should see if it makes sense to expose a way for clients to construct Protobuf messages when using these streams. We should also see about perf tradeoff. cc: @andrijast-db @danilonajkov-db

zlata-stefanovic-db: Will do some benchmarking for this

purego/examples/dynamic/single/main.go
zlata-stefanovic-db: This is already possible by constructing a DescriptorProto, passing its serialized bytes to WithProto, and creating records with dynamicpb.NewMessage. No .proto generation is required. I’ll add an example demonstrating this runtime-only path.

zlata-stefanovic-db: It'd look something like this:
```
descriptor := &descriptorpb.DescriptorProto{
    Name: proto.String("User"),
    Field: []*descriptorpb.FieldDescriptorProto{
        {
            Name:   proto.String("id"),
            Number: proto.Int32(1),
            Type:   descriptorpb.FieldDescriptorProto_TYPE_INT64.Enum(),
        },
        {
            Name:   proto.String("username"),
            Number: proto.Int32(2),
            Type:   descriptorpb.FieldDescriptorProto_TYPE_STRING.Enum(),
        },
    },
}

descriptorBytes, err := proto.Marshal(descriptor)
```
Do you think this fits the idea? @andrijast-db

zlata-stefanovic-db: I added examples/proto/runtime

purego/examples/dynamic/single/main.go
zlata-stefanovic-db: moved them to a util file

purego/examples/dynamic/single/main.go
andrijast-db: Cool, didn't know about WIthProto.

Why are we then introducing new stream type (DynamicProtoStream)? Why not just expose a function that can fetch proto from UC and pass that to WithProto? And just add `IngestJSONOffset` to regular stream. That will unify all dimensions. I'd say that's much cleaner, especially that call to UC will be explicit now. What do you think?

zlata-stefanovic-db: Yeah can do that as well - this choice was just to make the average user experience cleaner  - a normal user could skip the fetch from UC function and that directly happens when they open the stream (no need to know what happens in the background), but for a more advanced user it would make sense to separate these.

andrijast-db: For a user, it will roughly look something like:

```
desc := fetchProtoFromUC(uc_params)
stream := createStream().withProto(desc)

stream.ingest(struct {11, "jedan"})
stream.ingestJSON(makeJson(12, "dva"))
```

purego/examples/dynamic/single/main.go
zlata-stefanovic-db: Will do some benchmarking for perf before adding these. I'd delegate it to a follow-up PR as arbitrary Go structs would require more reflection and a larger mapping contract

purego/examples/dynamic/single/main.go
zlata-stefanovic-db: yes, exactly. i just reduced it to only creating the stream for simplicity

zlata-stefanovic-db: Divided it in the suggested way now

purego/internal/dynamicproto/converter.go
zlata-stefanovic-db: I added DynamicProtoStream.MessageDescriptor() and an example showing how clients can construct a dynamicpb.Message, marshal it, and use IngestRecordOffset, bypassing JSON conversion. (from the discussion with Andrija)

An encoding-only benchmark using the 8-field Order example showed:

JSON → dynamic protobuf: 6.15 µs/op, 45 allocs
Direct dynamic protobuf: 2.25 µs/op, 21 allocs (~2.7x faster)
Generated protobuf: 0.43 µs/op, 10 allocs (~14x faster)
This lets us retain the convenient JSON API while providing a faster protobuf path for throughput-sensitive clients.

purego/examples/dynamic/single/main.go
```diff
@@ -0,0 +1,65 @@
...
+	"log"
+
+	"github.com/databricks/zerobus-sdk/purego/examples/config"
+	"github.com/databricks/zerobus-sdk/purego/examples/internal/exampleutil"
+	"github.com/databricks/zerobus-sdk/purego/zerobus"
+)
+
+func openStream(sdk *zerobus.SDK, cfg config.Settings) (*zerobus.Stream, error) {
+	ctx := context.Background()
+	descriptor, err := sdk.FetchProtoDescriptor(
```
andrijast-db: nit: `FetchProtoDescriptorFromUC`

purego/zerobus/options.go
```diff
@@ -79,10 +98,9 @@ func WithJSON() StreamOption {
 	}
 }
 
-// WithProto selects Protocol Buffer record encoding. descriptorProto is the
-// serialized message descriptor (a FileDescriptorProto / DescriptorProto) the
-// service uses to interpret the raw protobuf record bytes; it is required for
-// proto streams.
+// WithProto selects Protocol Buffer record encoding. descriptorProto may be a
+// serialized DescriptorProto or FileDescriptorProto. Runtime JSON conversion
```
teodordelibasic-db: `WithProto` cannot safely accept serialized `FileDescriptorProto` bytes as currently implemented. The create request forwards those bytes unchanged in `descriptor_proto`. We should narrow `WithProto` to serialized `DescriptorProto` bytes.

purego/internal/dynamicproto/converter.go
```diff
@@ -0,0 +1,75 @@
...
+// EncodeJSONBytes converts one JSON payload to protobuf bytes.
+func (c *Converter) EncodeJSONBytes(record []byte) ([]byte, error) {
+	if c == nil || c.message == nil {
+		return nil, fmt.Errorf("dynamicproto: converter is not initialized")
+	}
+	if len(bytes.TrimSpace(record)) == 0 {
+		return nil, fmt.Errorf("dynamicproto: record is empty")
+	}
+	msg := dynamicpb.NewMessage(c.message)
+	unmarshal := protojson.UnmarshalOptions{DiscardUnknown: true}
```
teodordelibasic-db: Unknown keys can be accepted and acknowledged after their values have been discarded. A typo in a nullable field becomes missing data, and a field added while the descriptor cache is stale is silently removed from the record.

Could we make strict parsing the default and offer lenient parsing only as an explicit opt-in?

```go
unmarshal := protojson.UnmarshalOptions{}
```

`TestConverter_UnknownFieldIgnored` is a focused regression seam. It currently requires an unknown key to succeed; the default-path assertion should reject it, with a separate success case only if an ignore-unknown option is added.

purego/zerobus/sdk.go
```diff
@@ -294,6 +368,134 @@ func validateStreamArgs(tableName string, sc streamConfig) error {
...
+	}
+	if strings.TrimSpace(clientID) == "" {
+		return nil, fmt.Errorf("clientID is required")
+	}
+	if strings.TrimSpace(clientSecret) == "" {
+		return nil, fmt.Errorf("clientSecret is required")
+	}
+
+	cacheKey := dynamicSchemaCacheKey(s.ucEndpoint, tableName, clientID, clientSecret)
+	if desc, ok := s.getDynamicDescriptorFromCache(cacheKey); ok {
```
teodordelibasic-db, line 398: `FetchProtoDescriptor` can return the old schema for five minutes after a table change, with no refresh, invalidation, or cache-bypass API. The cache key contains endpoint, table, and credentials but no schema identity. A caller that deliberately fetches again before creating a new stream still receives the stale descriptor, and the converter can silently discard newly added nullable fields because it accepts unknown JSON keys.

Could we make an explicit fetch fresh by default, or provide a clear force-refresh path and document the cache TTL and 128-entry cap in the godoc and README? For example:

```go
func (s *SDK) InvalidateProtoDescriptor(tableName string)
```

It would also be useful to evict the cached entry when stream creation reports a schema mismatch.

All of this can be a follow up.

purego/internal/schema/schema.go
```diff
@@ -0,0 +1,689 @@
...
+	}
+	var c complexTypeJSON
+	if err := json.Unmarshal(data, &c); err != nil {
+		return err
+	}
+	r.complex = &c
+	return nil
+}
+
+type complexTypeJSON struct {
```
teodordelibasic-db, line 347: Nullable arrays and maps lose data semantics in the new JSON path. Every collection becomes a protobuf `repeated` or map field, and `containsNull` plus `valueContainsNull` are not parsed. An omitted collection, JSON `null`, and an empty collection therefore produce identical protobuf bytes, while null elements and null map values are rejected even when the UC schema permits them.

Could we capture these flags and reject collection shapes or JSON values that cannot be represented until there is a server-compatible presence mapping?

```go
type complexTypeJSON struct {
    // existing fields
    ContainsNull      *bool `json:"containsNull"`
    ValueContainsNull *bool `json:"valueContainsNull"`
}
```

A focused converter test can assert that explicit collection `null` is rejected rather than encoding identically to omitted and empty, and that null elements and values follow the UC flags.

purego/internal/ucschema/fetcher.go
```diff
@@ -0,0 +1,357 @@
...
+func isRetryableTransportError(ctx context.Context, err error) bool {
+	if err == nil {
+		return false
+	}
+	if ctx.Err() != nil {
+		return errors.Is(context.Cause(ctx), errRequestTimeout)
+	}
+	var nerr net.Error
+	if errors.As(err, &nerr) {
+		return true
```
teodordelibasic-db, line 292: Permanent certificate and hostname failures are reported as retryable because every `net.Error` and `url.Error` returns `true`. Callers honoring `zerobus.Retryable` can keep retrying a configuration error that cannot recover.

Could we match the established classifier in `internal/auth`, keeping timeouts and connection-level `net.OpError` failures retryable while leaving certificate verification errors terminal?

```go
var nerr net.Error
if errors.As(err, &nerr) && nerr.Timeout() {
    return true
}
var opErr *net.OpError
return errors.As(err, &opErr)
```

An untrusted `httptest.NewTLSServer` is a deterministic regression seam: `FetchTableSchema` currently returns a retryable `FetchError`, and it should be non-retryable after the fix while the existing client-timeout test remains retryable.

purego/README.md
```diff
@@ -82,6 +80,88 @@ first successful open, so it does not terminate the live stream.
...
+To skip JSON conversion while retaining UC schema discovery, build dynamic
+protobuf messages from the stream descriptor:
+
+```go
+descriptor := stream.MessageDescriptor()
+idField := descriptor.Fields().ByName("id")
+
+for _, id := range ids {
+    message := dynamicpb.NewMessage(descriptor)
+    message.Set(idField, protoreflect.ValueOfInt64(id))
```
teodordelibasic-db: The copy-paste example assigns an `int64` reflection value to the documented `id INT` field, whose generated descriptor is `TYPE_INT32`. `dynamicpb.Message.Set` panics on that type mismatch. Could we use the same type as the complete dynamic example?

```go
message.Set(idField, protoreflect.ValueOfInt32(int32(id)))
```

purego/zerobus/sdk.go
```diff
@@ -294,6 +368,134 @@ func validateStreamArgs(tableName string, sc streamConfig) error {
...
+		WorkspaceEndpoint: s.ucEndpoint,
+		ClientID:          clientID,
+		ClientSecret:      clientSecret,
+		HTTPClient:        s.httpClient,
+		RequestTimeout:    s.dynamicSchemaFetchTimeout,
+	})
+	if err != nil {
+		return nil, err
+	}
+	tableSchema, err := fetcher.FetchTableSchema(ctx, tableName)
```
teodordelibasic-db, line 412: Concurrent misses for the same table each mint an OAuth token, fetch the same UC schema, build the same descriptor, and then overwrite the same cache entry. This multiplies startup load and makes throttling more likely.

Could we coalesce work per cache key with a keyed in-flight result or `singleflight`, while preserving independent cancellation for waiting callers?

`TestSDKFetchProtoDescriptor_CacheHit` already has token and schema counters. Starting several callers together and holding the handlers until they are all waiting would assert one token request and one schema request. It observes one pair per caller before coalescing and one pair total afterward.

purego/internal/schema/schema.go
```diff
@@ -0,0 +1,689 @@
...
+
+func newMessageCollector() *messageCollector {
+	return &messageCollector{
+		used:           make(map[string]struct{}),
+		mapEntryOwners: make(map[string]string),
+	}
+}
+
+func (c *messageCollector) reserveMapEntry(fieldName, path string) error {
+	entryName := sanitizeMessageName(fieldName) + "Entry"
```
teodordelibasic-db, line 515: Distinct valid fields such as `foo_bar` and `foo__bar` both require `FooBarEntry`, so the entire table schema is rejected. Struct helper names already use deterministic numeric suffixes, and Rust applies the same approach to map entries.

Could we reserve each map entry through `uniqueName` and remember the assigned name per field?

```go
entryName := c.uniqueName(sanitizeMessageName(fieldName) + "Entry")
c.mapEntryNames[fieldName] = entryName
```

`TestDescriptorFromUCColumns_RejectsMapEntryNameCollision` is the exact seam to flip: the two valid fields should produce distinct entry names and a descriptor accepted by `protodesc.NewFile`.

teodordelibasic-db (commented):
Some of these can probably be follow ups to not explode the PR further 🙂

purego/examples/dynamic/single/main.go
zlata-stefanovic-db: fixed

purego/zerobus/options.go
zlata-stefanovic-db: Changed the docs regarding this for clarity - descriptor proto might only be serialized DescriptorProto bytes now.

purego/zerobus/sdk.go
zlata-stefanovic-db: Linking this to a follow-up PR

purego/zerobus/sdk.go
zlata-stefanovic-db: Linking this to a follow-up PR

purego/internal/dynamicproto/converter.go
zlata-stefanovic-db: Fixed, Unknown fields now produce an error before the record is queued. Removed DiscardUnknown: true, changed TestConverter_UnknownFieldIgnored into TestConverter_UnknownFieldRejected, and updated an integration test that had intentionally used an unknown field as padding. No lenient opt-in added for now, might in the follow up if we think it's needed

zlata-stefanovic-db: The lenient parsing could be useful in some very specific cases like producers sending extra metadata that is intentionally ignored or similar, but it would be a new StreamOption, wiring, docs, etc. so we might want to consider it later

purego/README.md
zlata-stefanovic-db: Yes, thank you. Changed to ```protoreflect.ValueOfInt32(int32(id))```

purego/internal/ucschema/fetcher.go
zlata-stefanovic-db: Fixed in isRetryableTransportError in purego/internal/ucschema/fetcher.go. It now treats only network timeouts, *net.OpError connection failures, and the SDK’s own request timeout as retryable. TLS certificate and hostname verification failures now remain non-retryable.

I also added TestFetchTableSchema_TLSCertificateErrorNotRetryable and retained TestFetchTableSchema_HTTPClientTimeoutRetryable to cover both sides of the classification.

purego/internal/schema/schema.go
zlata-stefanovic-db: Fixed by capturing containsNull and valueContainsNull in complexTypeJSON and validating them in validateCollectionRepresentation.

UC schemas with nullable collection fields, null array elements, or null map values are now rejected because protobuf cannot preserve those distinctions. rejectNullCollections also recursively rejects explicit JSON null collection fields before ingestion.

Added focused schema and converter tests, and documented the limitation in the README. Full nullable-collection representation can remain a follow-up.

purego/internal/schema/schema.go
zlata-stefanovic-db: Fixed in messageCollector.reserveMapEntry and mapComplexTypeToProtobuf.

One nuance: I can't apply the Rust suffixing approach - assigning a numeric suffix such as FooBarEntry2 is rejected by Go’s protobuf validator because map entries must use the exact implicit name derived from the field. Compatible maps whose names normalize to the same entry name now share that canonical descriptor. Incompatible map shapes return a clear error because protobuf cannot represent them safely.

TestDescriptorFromUCColumns_SharesCompatibleMapEntryNameCollision verifies that the resulting descriptor is accepted by protodesc.NewFile.

purego/zerobus/sdk.go
zlata-stefanovic-db: Tracking here: https://github.com/databricks/zerobus-sdk/issues/676

purego/zerobus/sdk.go
zlata-stefanovic-db: Tracking here: https://github.com/databricks/zerobus-sdk/issues/676

teodordelibasic-db (dismissed):

teodordelibasic-db (approved):

PR #673, opened 2026-08-04
[Rust][FFI] Add Arrow C Data ingestion - zero copy

nikolaobradovic-db:

- Add zerobus_arrow_stream_ingest_c_data, an additive C API that transfers canonical ArrowArray/ArrowSchema ownership without an IPC encode/decode round trip while still be able to use existing Flight encoding (compression, chunking, dictionary hydration, ACK, recovery, and error behavior is preserved)
- Add the disabled-by-default internal-arrow-c-data SDK feature so Zerobus native bindings can share one importer, Arrow type identity, validation, and ownership implementation.
- Retain C-specific pointer handling, panic containment, runtime bridging, and ABI marker types in zerobus-ffi.
- Add rlib output only for workspace integration tests that call the real C ABI; shipped FFI artifacts remain static/dynamic libraries.
- Add dedicated C ABI lifetime coverage for release-after-ACK and terminal retention, plus wire-level compression, chunking, dictionary payload, and nested ownership tests.

## How is this tested?

arrow_c_data_ffi_tests added

### DIFF
--- a/rust/ffi/src/arrow.rs
+++ b/rust/ffi/src/arrow.rs
@@ -380,7 +380,11 @@ pub extern "C" fn zerobus_arrow_stream_ingest_c_data(
 ) -> i64 {
     ffi_guard(result, -1, move || {
         if array.is_null() || schema.is_null() {
-            write_error_result(result, "ArrowArray and ArrowSchema pointers are required", false);
+            write_error_result(
+                result,
+                "ArrowArray and ArrowSchema pointers are required",
+                false,
+            );
             return -1;
         }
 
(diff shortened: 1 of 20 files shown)


teodordelibasic-db (dismissed):

teodordelibasic-db (approved):

rust/ffi/src/arrow.rs
elenagaljak-db, line 304: Only relevant thing that I found: 
`zerobus_arrow_stream_free` does not wait for the background task the way close() does
close() stops the supervisor and waits for it to finish. free() (via Drop) only asks it to stop, then returns. After `ingest_c_data`, the SDK may still hold the batch in that task, so the caller’s release callback can run later on a runtime thread — after free() has already returned.

If the caller destroys state the callback needs at free() time (allocator, wrapper context, etc.), the late callback can crash. I managed to reproduce this with a test and my code panicked.

This is unlikely after a normal flush → close → free, but realistic on failure/cleanup paths where the stream is freed while batches are still retained. 
Also, this behavior existed before this PR but it matters more now because C Data retains caller-owned buffers and release callbacks, not just SDK-owned IPC copies.

nikolaobradovic-db: I think I need @teodordelibasic-db to finish https://github.com/databricks/zerobus-sdk/pull/677 to properly fix this use-after-free, and it can be done as a follow-up.

PR #675, opened 2026-08-05
[Python] Support Python 3.13 and 3.14

### LINKED ISSUE #652: Python SDK crashes on CPython 3.14 — bump PyO3 to support new interpreters
### Problem

The Python SDK is not compatible with CPython 3.14 and segfaults on `ZerobusSdk` instantiation. Two compounding issues:

1. **No install guard.** `python/pyproject.toml` declares `requires-python = ">=3.9"` with no upper bound, so `pip install` succeeds on 3.14. Because the wheel is built with PyO3's stable ABI (`abi3-py39`), pip selects the prebuilt `cp39-abi3` wheel on 3.14 rather than failing.
2. **Outdated bindings.** We're on PyO3 0.20.3 (`python/rust/Cargo.toml`), which only supports CPython through 3.12. CPython 3.14 support first landed in PyO3 0.25 and was hardened in 0.26.

### Proper fix

- Bump PyO3 from 0.20 → 0.26 and migrate off the deprecated `pyo3-asyncio` (successor: `pyo3-async-runtimes`). This is a non-trivial upgrade with breaking API changes in `python/rust/src/lib.rs`.
- Add 3.13 and 3.14 to the CI matrix (`.github/workflows/ci-python.yml`, currently 3.9–3.12) and to the `classifiers` in `pyproject.toml`.

elenagaljak-db:
Fixes #652

  **WHY:** The SDK segfaults on CPython 3.14 when you create a `ZerobusSdk`. Two things
  combine to cause it. The bindings used PyO3 0.20, which supports CPython only up to 3.12.
  And `requires-python` declared no upper bound, so pip installed the package on 3.14 anyway
  - the wheel targets the stable ABI (`abi3`), which matches any CPython >= 3.9, so pip
  selected the prebuilt `cp39-abi3` wheel instead of failing. The stable ABI only holds if
  the binding uses the limited API correctly, and 0.20 predates 3.13 and 3.14 entirely. CI
  compounded this by stopping at 3.12, so nothing signalled the problem before users hit it.

  **WHAT:**

  **Upgraded PyO3 0.20 -> 0.29.0** and replaced the deprecated `pyo3-asyncio` crate with
  `pyo3-async-runtimes` 0.29. CPython 3.14 support first landed in PyO3 0.25 and was hardened
  in 0.26; the linked issue proposed 0.26, but 0.29.0 is current and the migration cost is
  effectively identical, so it buys more runway before the next forced upgrade.

  The jump required migrating to the `Bound<'py, T>` API and absorbing several renames:
  `Python::with_gil` -> `attach`, `Python::allow_threads` -> `detach`, `downcast` -> `cast`,
  `PyObject` -> `Py<PyAny>`. The `gil-refs` escape hatch was removed after 0.22, so this
  could not be staged across smaller steps.

  Two decisions in here are not obvious from the diff:

  - **`StreamConfigurationOptions` implements `Clone` by hand.** `Py<T>` is no longer
    unconditionally `Clone` in 0.29. The quick fix is PyO3's `py-clone` feature, but it
    *panics* when the interpreter is detached - precisely the state of the async
    stream-builder paths that clone `ack_callback`. That would have traded a compile error
    for a runtime crash, so the hand-written impl uses `clone_ref` under `Python::attach`
    instead.
  - **`FromPyObject` is becoming opt-in**, which forced a per-class choice rather than a
    blanket flag. Value and config types genuinely extracted from arguments get
    `from_py_object`. The two subclassable bases (`AckCallback`, `HeadersProvider`) get
    `skip_from_py_object`, because clone-based extraction would silently strip a Python
    subclass - a subtle correctness bug rather than a warning.

  **Capped `requires-python` at `>=3.9,<3.15`.** The missing upper bound is the mechanism
  that made the crash reachable at all, so leaving it open would preserve the same trap for
  3.15. Note this is `<3.15` and not `<=3.14`: pip compares full versions, so `<=3.14` would
  reject 3.14.1 and later - including 3.14.4, which this PR verifies works. The bound moves
  up once CI covers the next release.

  **Split the `pyarrow` requirement by Python version.** pyarrow dropped CPython 3.9 *before*
  it added 3.14, so no single version spans the range this SDK supports:

  | pyarrow | 3.9 | 3.13 | 3.14 |
  |---------|-----|------|------|
  | <= 21.0.0 | yes | yes | no |
  | >= 22.0.0 | no  | yes | yes |

  The `arrow` extra now uses environment markers (`< 22.0` below 3.14, `>= 22.0` from 3.14
  up), and `requirements-dev.lock` is regenerated as a universal lock with the same split.
  The previous `< 20.0` ceiling excluded every version that works on 3.14, which made the

  | pyarrow | 3.9 | 3.13 | 3.14 |
  |---------|-----|------|------|
  | <= 21.0.0 | yes | yes | no |
  | >= 22.0.0 | no  | yes | yes |

  The `arrow` extra now uses environment markers (`< 22.0` below 3.14, `>= 22.0` from 3.14
  up), and `requirements-dev.lock` is regenerated as a universal lock with the same split.
  The previous `< 20.0` ceiling excluded every version that works on 3.14, which made the
  extra uninstallable there even though Arrow ingestion itself is fine on 3.14. This is a
  constraint of pyarrow's own support window, not of our code.

  **Added 3.13 and 3.14 to CI** on Linux and Windows (Windows previously ran only 3.12).

  **Also:** `tests/test_arrow.py` now uses `pytest.importorskip("pyarrow")`. It imported
  pyarrow at module level, so on a core-only install - no `arrow` extra, a supported and
  probably common configuration - the
**Local tests not covered by the unit suite:**

- **Live ingestion against a real workspace on 3.9, 3.13, and 3.14** using the examples in
  `python/examples/`: OAuth, stream creation, ingest, flush, close. This is the path the
  unit tests deliberately skip (they make no network connections), so it was worth
  exercising directly.
- **Both ends of the abi3 range from the same wheel** -3.9 floor and 3.14 ceiling - since
  a stable-ABI mistake would surface at the boundaries.
- **`requires-python` bound**: installs succeed on 3.9.25 and 3.14.4. The 3.14.4 case is
  the one that a `<=3.14` bound would have silently broken.
- **`pip install "...[arrow]"`** resolves pyarrow 21.0.0 on 3.9 and 25.0.0 on 3.14, with
  Arrow tests included in both runs.
- **Core-only install (no pyarrow)**: 106 passed, 1 skipped - previously 0 tests ran at all.
- **Public API unchanged**, verified by diffing the introspected class and method surface of
  a build of `main` against this branch: byte-identical. No type-stub changes needed, and
  not a semver break.
- rustfmt, clippy, pycodestyle, black, and isort are clean.

**Not tested:** free-threaded builds (`3.14t`) are out of scope -`abi3` wheels cannot load
into a free-threaded interpreter, and supporting it needs version-specific wheels plus a
thread-safety audit. Regular 3.14 is what this fixes.

Co-authored by Isaac.


python/rust/Cargo.toml
```diff
@@ -2,14 +2,15 @@
...
+rust-version = "1.83"
 
 [lib]
 name = "_zerobus_core"
 crate-type = ["cdylib"]
 
 [dependencies]
-pyo3 = { version = "0.20", features = ["extension-module", "abi3-py39"] }
-pyo3-asyncio = { version = "0.20", features = ["tokio-runtime"] }
+pyo3 = { version = "0.29.0", features = ["extension-module", "abi3-py39"] }
```
teodordelibasic-db: `0.29.0` is still a caret requirement in Cargo, so both dependencies may resolve to `0.29.1`. The current `Cargo.lock` keeps builds on `0.29.0`, but regenerating it can select the JFrog-blocked `0.29.1` and reintroduce the build failure. Maybe we should make the workaround an exact constraint while it is needed?

```toml
pyo3 = { version = "=0.29.0", features = ["extension-module", "abi3-py39"] }
pyo3-async-runtimes = { version = "=0.29.0", features = ["tokio-runtime"] }
```

python/README.md
```diff
@@ -75,13 +75,18 @@ pip install databricks-zerobus-ingest-sdk
...
 - **macOS**: x86_64, arm64
 - **Windows**: x86_64
 
 ### Python Version
 
-Requires **Python 3.9 or higher**.
+Requires **Python 3.9-3.14**. The SDK is tested on CPython 3.9 through 3.14.
+
+The wheel uses the CPython stable ABI (`abi3`), so one wheel works on every
+supported version. The free-threaded builds (like`3.14t`) are not
```
teodordelibasic-db: Missing space between `like` and `3.14t`.

python/pyproject.toml
```diff
@@ -7,7 +7,7 @@ name = "databricks-zerobus-ingest-sdk"
 dynamic = ["version"]
 description = "Databricks Zerobus Ingest SDK for Python"
 readme = "README.md"
-requires-python = ">=3.9"
+requires-python = ">=3.9,<3.15"
```
teodordelibasic-db, line 10: We should probably update `python/NEXT_CHANGELOG.md` and `python/CONTRIBUTING.md` to reflect this.

python/requirements-dev.lock
teodordelibasic-db, line 1: This file contains both `pyarrow` branches because it was generated with `--universal --python-version 3.9`. `make lock-dev` from `python/Makefile` omits those flags, so its output depends on the local Python version and can drop one branch. Maybe we should make the target use the same command as the lockfile?

```make
uv pip compile --universal --python-version 3.9 \
        --generate-hashes --no-emit-index-url \
        --exclude-newer '2026-03-19T00:00:00Z' \
        --output-file requirements-dev.lock requirements-dev.in
```

.github/workflows/ci-python.yml
teodordelibasic-db, line 81: CI uses `maturin develop`, which rebuilds the extension separately for every Python version. The original crash happened when a wheel built on Python `3.9` was installed on Python `3.14`. Could we build that wheel once on Python `3.9` and install the same file in a Python `3.14` smoke test?

```yaml
# After downloading the Python 3.9 wheel in a Python 3.14 job:
- run: python -m pip install dist/*.whl
- run: >-
    python -c 'from zerobus.sdk.sync import ZerobusSdk;
    ZerobusSdk("https://test-workspace.zerobus.region.cloud.databricks.com",
    "https://test-workspace.cloud.databricks.com")'
```

teodordelibasic-db (commented):
Minor comments, cool change.

.github/workflows/ci-python.yml
```diff
@@ -79,6 +91,66 @@ jobs:
...
+        run: cp ../LICENSE LICENSE
+
+      - name: Build abi3 wheel on Python 3.9
+        run: |
+          python -c 'import sys; assert sys.version_info[:2] == (3, 9), sys.version'
+          python -m pip install --upgrade pip
+          python -m pip install maturin==1.8.3
+          python -m maturin build --release --manifest-path rust/Cargo.toml --out dist
+
+      - name: Install and import the 3.9 wheel on every supported version
```
teodordelibasic-db, line 137: The wheel builds successfully, but this smoke test imports the checkout instead of the installed wheel because the job runs from `python/`. The log shows it loading `python/zerobus/__init__.py` and then failing to find `zerobus._zerobus_core`. Maybe we can run the interpreter in isolated mode so the checkout is not added to `sys.path`?

```bash
"/tmp/abi3-$v/bin/python" -I - <<'PY'
```

teodordelibasic-db (approved):

python/rust/Cargo.toml
elenagaljak-db: Fixed

python/README.md
elenagaljak-db: Fixed

python/pyproject.toml
elenagaljak-db: Added

python/requirements-dev.lock
elenagaljak-db: Fixed

.github/workflows/ci-python.yml
elenagaljak-db: Yep, fixed

PR #678, opened 2026-08-05
[PureGo] Improve descriptor cache lifecycle

### LINKED ISSUE #676: [PureGo] Improve UC descriptor cache lifecycle
## Problem

`FetchProtoDescriptorFromUC` caches descriptors for five minutes, but callers cannot bypass or invalidate that cache after a Unity Catalog schema change. An explicit fetch can therefore return stale schema data until the TTL expires.

Concurrent cache misses for the same endpoint, table, and credentials also run independently. Each caller can mint an OAuth token, fetch the same UC schema, build the same descriptor, and overwrite the same cache entry, increasing startup load and throttling risk.

Strict unknown-field parsing now prevents newly added fields from being silently discarded, but stale descriptors can still cause avoidable ingestion failures.

## Proposed scope

- Define explicit refresh semantics for `FetchProtoDescriptorFromUC`.
  - Either fetch fresh by default, or provide a force-refresh/cache-bypass option.
  - Consider an API such as `InvalidateProtoDescriptor(tableName string)`.
- Document the current five-minute TTL and 128-entry limit.
- Consider evicting the cached descriptor when stream creation reports a schema mismatch.
- Coalesce in-flight descriptor fetches per cache key.
- Preserve independent cancellation for callers waiting on shared work.

## Acceptance criteria

- A caller can deliberately obtain a fresh descriptor after a UC schema change.
- Concurrent misses for one cache key produce one token request and one schema request.
- Cancelling one waiting caller does not cancel shared work needed by other callers.
- Cache TTL, capacity, refresh, and invalidation behavior are documented and tested.

## Review context

- Cache refresh/invalidation: https://github.com/databricks/zerobus-sdk/pull/666#discussion_r3714553823
- Concurrent miss coalescing: https://github.com/databricks/zerobus-sdk/pull/666#discussion_r3714604513

zlata-stefanovic-db:
## Summary

- Add `RefreshProtoDescriptorFromUC` so callers can bypass a stale cached descriptor and replace it after a Unity Catalog table schema change.
- Rename/clarify the fetch API as `FetchProtoDescriptorFromUC` (with `FetchProtoDescriptor` kept as a deprecated alias).
- Coalesce compatible concurrent descriptor requests for the same table and credentials:
  - ordinary cache misses share one UC fetch
  - concurrent refreshes share one refresh request and never join an older ordinary fetch
  - ordinary misses may join an in-flight refresh when no cached descriptor is available
- Preserve independent caller cancellation: canceling one waiter does not cancel shared work still needed by others; abandoned fetches stop when the last waiter leaves.
- Cancel in-flight descriptor fetches on `SDK.Close`, wait for them to finish, and keep cache writes generation-aware so superseded/stale results cannot overwrite fresher entries.
- Document the 5-minute / 128-entry cache limits and coalescing/refresh behavior in the README and godoc.
- Add regression coverage for refresh bypass, refresh-vs-fetch precedence, coalescing, cancellation isolation, last-waiter abandonment, Close cancellation, and failed shared-fetch retry.

## Test plan

- [x] `cd purego/examples && go test ./...`

Closes #676

### DIFF
--- a/purego/NEXT_CHANGELOG.md
+++ b/purego/NEXT_CHANGELOG.md
@@ -11,8 +11,12 @@
 - Added opt-in wait-ready stream creation whose context bounds the complete
   first-open process; asynchronous creation now preserves context values while
   detaching caller cancellation.
-- Added `FetchProtoDescriptor` for building protobuf descriptors from Unity
-  Catalog table schemas.
+- Added `FetchProtoDescriptorFromUC` for building protobuf descriptors from
+  Unity Catalog table schemas.
+- Added `RefreshProtoDescriptorFromUC` for bypassing and replacing a cached
+  descriptor after a table schema change.
+- Coalesced concurrent descriptor fetches for the same table and credentials
+  while preserving independent caller cancellation.
 - Added `Stream.IngestJSONOffset` and `Stream.IngestJSONRecordsOffset`. JSON
   streams queue JSON directly; proto streams convert it before ingestion.
 
--- a/purego/NEXT_CHANGELOG.md
+++ b/purego/NEXT_CHANGELOG.md
@@ -15,8 +15,8 @@
   Unity Catalog table schemas.
 - Added `RefreshProtoDescriptorFromUC` for bypassing and replacing a cached
   descriptor after a table schema change.
-- Coalesced concurrent descriptor fetches for the same table and credentials
-  while preserving independent caller cancellation.
+- Coalesced compatible descriptor requests while keeping refreshes newer than
+  ordinary fetches, preserving caller cancellation, and stopping abandoned work.
 - Added `Stream.IngestJSONOffset` and `Stream.IngestJSONRecordsOffset`. JSON
   streams queue JSON directly; proto streams convert it before ingestion.
 
--- a/purego/README.md
+++ b/purego/README.md
@@ -92,6 +92,20 @@ streams and convert it to protobuf on proto streams.
 Fetch the table schema explicitly, then create a regular proto stream. The
 stream can accept either protobuf bytes or JSON converted at ingest time.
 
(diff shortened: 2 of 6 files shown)


andrijast-db:
Now that we've extracted fetching logic into a separate function, I don't think we need caching there. Since fetch is explicit, client can fetch once and reuse as much as needed. Caching just adds more complexity, and creates this weird API where we introduce `RefreshProtoDescriptorFromUC` function as well.

zlata-stefanovic-db:
> Now that we've extracted fetching logic into a separate function, I don't think we need caching there. Since fetch is explicit, client can fetch once and reuse as much as needed. Caching just adds more complexity, and creates this weird API where we introduce `RefreshProtoDescriptorFromUC` function as well.

Thank you, looking into this

zlata-stefanovic-db:
> Now that we've extracted fetching logic into a separate function, I don't think we need caching there. Since fetch is explicit, client can fetch once and reuse as much as needed. Caching just adds more complexity, and creates this weird API where we introduce `RefreshProtoDescriptorFromUC` function as well.

@teodordelibasic-db should we reconsider having the descriptor cache? Or just adding into the docs an instruction of how to properly use the fetch from UC function (only using it once) since it is separate from opening the stream now?

zlata-stefanovic-db:
Closing in favor of removing the descriptor cache entirely.

@andrijast-db is right: now that fetching the UC descriptor is an explicit, standalone call rather than a step inside stream creation, the SDK does not need to cache it. Callers fetch once and reuse the returned descriptor bytes for as many streams as they want, which makes the cache (plus its TTL, entry limit, coalescing, and the extra `RefreshProtoDescriptorFromUC` API) pure complexity with no benefit.

Replacing this with a PR that deletes the cache and documents the fetch-once-and-reuse contract in the README and godoc.

PR #682, opened 2026-08-06
[PureGo] Generalize stream durability model

zlata-stefanovic-db:
## Summary

Generalizes the internal stream core so one implementation can serve both the existing atomic protocols (Protocol Buffers, JSON) and a future record-count protocol, with no public API change.

- **Durability units** — acknowledgments are tracked as cumulative unit counts over submitted ranges rather than one ack per item, so a protocol can acknowledge part of a batch.
- **Partial replay slicing** — on reconnect, only the unacknowledged suffix of a partially acknowledged item is replayed.
- **Submission receipts** — a transport can report how much of a multi-frame send actually reached the server, so a failure part-way through a batch keeps the acknowledged prefix instead of discarding it.
- **Extension seams** — `EncoderHooks`, `AckModelHooks`, `OpenFunc`, and an exported `WireStream` let a protocol instantiate the core over its own payload type.

Atomic proto/JSON behavior is unchanged: those encoders report one unit per item, never produce partial acknowledgments, and continue to resolve wire offsets through the existing path.

## Behavior fixes found while reviewing this change

- An acknowledgment that arrives while its `Send` is still outstanding, where that `Send` then fails, is no longer converted into a non-retryable protocol error. The core cannot know how much of that send reached the server, which is not a server violation — so the unusable ack is discarded and the retryable send failure drives recovery. Without this, a transient failure killed the stream permanently instead of reconnecting, and blamed the server for a violation it never committed.
- Rejecting an impossible submission receipt now also clears the buffered-ack state. Otherwise teardown waits forever for a completion event that was already consumed, hanging the receiver, the supervisor, and every `Close()`/`Terminate()` caller.
- An impossible submission receipt is now terminal rather than retryable. A receipt larger than the payload, or a short receipt from a `Send` that reported success, is a deterministic transport defect: reconnecting replays the same payload into the same defect, so the stream burned its whole retry budget before failing. Accepting a short receipt from a successful send would also record unsubmitted units as submitted, so it has to fail loudly instead.
- Durable progress refreshes the head item's lack-of-ack budget, including when a full ack promotes the next item to head. Previously only a partial ack landing inside the head refreshed it, so a boundary-aligned ack promoted an item still carrying the deadline it was given on entering flight. Once partial acks had carried the head past that deadline, completing it tore down a healthy stream on the very ack that proved the server alive. Acks are ordered, so a promoted item could not have been made durable sooner; a duplicate or stale ack still makes no progress, so a stalled server cannot postpone recovery.
- A definitive server rejection now survives the prefix-ACK teardown path. When a partial submission's prefix ACK ended the durability wait, the receiver was reaped without its result being read, so the server's status was dropped and the earlier send error was returned instead — recovery replayed against a server that rejects every attempt, and an auth rejection never invalidated its credential. The sender-first arbitration now also runs after the reap, restricted to definitive rejections so the forced abort's own retryable error cannot replace the cause.
- An acknowledgment that advances the watermark but resolves to no offset (landing in a gap below every submitted range) is rejected rather than silently raising the connection's ack watermark with no durable progress behind it.

## Performance

Resolving an acknowledgment while a `Send` is in flight no longer allocates or copies the outstanding window. `AckState` carries the active range beside the submitted slice (`Active`/`HasActive`, read through `NumRanges`/`RangeAt`) instead of materializing a combined slice, so the resolver walks both in submission order and ack processing is no longer O(n) in the in-flight window.

An A/B run against a real Delta table showed no throughput change for proto/JSON ingestion from the generalized ack model. That run predates the copy removal above, so it is a lower bound on the current state.

## Test plan

- [x] `cd purego && go test -race ./...` — 543 tests pass, with and without `-race`
- [x] `cd purego && go test -count=3 ./internal/stream` — no flakiness in the timing-sensitive tests
- [x] `cd purego && go vet ./...` and `gofmt -l` clean
- [x] `deadcode ./...` reports nothing

`durability_model_test.go` exercises the new generic seam through a protocol-neutral row-based fake: unit-resolution table cases, partial acks, replay of only the unacknowledged remainder, receipt-preserved prefix acks, and hook validation.

Every behavior fix above has a regression guard that was confirmed to fail without its fix, by running the new test against the unmodified core in a scratch worktree:

- `TestHookProtocolImpossibleReceiptIsTerminal` — both receipt shapes fail with `recovery exhausted after 4 attempt(s)`, showing the payload was replayed into the same defect.
- `TestHookProtocolPromotedItemGetsFreshAckBudget` — fails with `opened 2 connections, want 1`, showing the promoted item was torn down and replayed.
- `TestHookProtocolPrefixAckPreservesTerminalRecvStatus` — fails with `recovery exhausted after 4 attempt(s)` instead of surfacing the server's `Unauthenticated` status, and asserts the credential is invalidated exactly once.
- The recovery guard in `core_test.go` passes against the pre-change core but failed against this branch; the teardown guard hangs to the test timeout without its fix.

## Notes

- No public API change and no Arrow dependency. This is the generic-core foundation for the PureGo Arrow Flight path, split out so it can be reviewed on its own.
- No `NEXT_CHANGELOG.md` entry: the change is confined to `purego/internal/`. Partial acknowledgments, submission receipts, and the record-count ack path are all unreachable for the shipping proto/JSON protocols, and the recovery fix restores behavior `main` already has, so there is no user-visible difference relative to `main`.
- Now based directly on `main`. It was previously stacked on `purego-descriptor-cache-lifecycle`, which was closed without merging.

### DIFF
--- a/purego/internal/stream/buffer.go
+++ b/purego/internal/stream/buffer.go
@@ -1,12 +1,8 @@
 // Package stream is the generic ingestion core: offset assignment, send/recv
 // goroutines, ack watermark, Flush/WaitForOffset, and the recovery supervisor.
-// Protocol-specific behaviour (encoding, ack parsing, wire transport) is
-// injected through the encoder, ackModel, and wireStream interfaces, so
-// proto and JSON share one implementation.
-//
-// Arrow Flight will reuse these seams but not unchanged: buffer entries carry no
-// record count, and recovery replays whole entries, so a partially acknowledged
-// batch cannot be sliced. Both are core changes, not encoder changes.
+// Protocol-specific behaviour (encoding, ack parsing, payload slicing, and wire
+// transport) is injected through narrow hooks, so atomic offset protocols and
+// record-count protocols share one implementation.
 package stream
 
 import (
@@ -15,20 +11,24 @@ import (
 	"fmt"
 	"math"
 	"sync"
+	"time"
 )
 
 // defaultMaxInflight is the fallback backpressure cap used when a non-positive
 // value reaches newBuffer.
 const defaultMaxInflight = 1_000_000
 
-// item is one unit of work in the buffer: an already-encoded wire message
-// paired with the logical offset that identifies it for acknowledgment. Req is
-// the wire request type the core is instantiated with (encodedMsg for
-// proto/JSON; a Flight frame for Arrow).
+// item is one logical unit of work in the buffer: an already-encoded payload
+// paired with its SDK offset and protocol durability-unit count. Proto/JSON
+// payloads are atomic and always carry one unit; a record-count protocol can
+// carry multiple units and acknowledge a prefix.
 type item[Req any] struct {
-	offset  int64
-	payload Req
-	weight  int64
+	offset     int64
+	payload    Req
+	units      uint64
+	ackedUnits uint64
+	weight     int64
+	pendingAt  time.Time
 }
 
 type discardResult struct {
@@ -54,9 +54,9 @@ type capacityWaiter struct {
 // Concurrency model:
 //   - Many goroutines may call enqueue concurrently.
 //   - Exactly one sender goroutine calls next.
-//   - Exactly one receiver goroutine calls discardThrough as acks arrive.
-//   - The supervisor calls requeue and drain, but only while the sender is
-//     stopped — so next never runs concurrently with requeue or drain.
+//   - Exactly one receiver goroutine calls acknowledge as acks arrive.
+//   - The supervisor calls requeueWithSlicer and drain, but only while the
+//     sender is stopped — so next never runs concurrently with either.
 //
 // All state (queue, flight, capacity, cond) is private; the sender and receiver
 // interact only through these methods, never by touching the fields directly.
@@ -74,6 +74,7 @@ type buffer[Req any] struct {
 	usedBytes        int64
 	accountingReset  bool
 	waiters          *list.List
+	flightRevision   uint64
 }
 
 func newBuffer[Req any](maxInflight int, byteLimit int64) *buffer[Req] {
@@ -184,8 +185,40 @@ func (b *buffer[Req]) release(weight int64) {
 	b.mu.Unlock()
 }
 
-// append adds an item after reserve succeeds.
-func (b *buffer[Req]) append(offset int64, msg Req, weight int64) error {
+// reconcileReservation replaces an admission estimate with the actual retained
+// payload size. Shrinks are always applied immediately. An underestimated
+// reservation may grow only when the unreserved byte budget is already
+// available; it never waits while holding an item slot, which would deadlock if
+// several concurrent builders all underestimated at the count limit.
+func (b *buffer[Req]) reconcileReservation(estimated, actual int64) error {
+	if actual < 0 || actual > b.maxBufferedBytes {
+		return fmt.Errorf("%w: buffered payload weight %d exceeds limit %d",
+			ErrPayloadTooLarge, actual, b.maxBufferedBytes)
+	}
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if b.closed {
+		return errClosed
+	}
+	delta := actual - estimated
+	if delta > b.maxBufferedBytes-b.usedBytes {
+		return fmt.Errorf(
+			"%w: actual retained payload weight %d exceeds reserved estimate %d and remaining byte capacity",
+			ErrPayloadTooLarge,
+			actual,
+			estimated,
+		)
+	}
+	b.usedBytes += delta
+	if delta < 0 {
+		b.grantWaitersLocked()
+	}
+	return nil
+}
+
+// appendUnits adds an item with its protocol durability-unit count after reserve
+// succeeds.
+func (b *buffer[Req]) appendUnits(offset int64, msg Req, units uint64, weight int64) error {
 	b.mu.Lock()
 	if b.closed {
 		if !b.accountingReset {
@@ -196,15 +229,17 @@ func (b *buffer[Req]) append(offset int64, msg Req, weight int64) error {
 		b.cond.Broadcast()
 		return errClosed
 	}
-	b.queue = append(b.queue, item[Req]{offset: offset, payload: msg, weight: weight})
+	b.queue = append(b.queue, item[Req]{
+		offset: offset, payload: msg, units: units, weight: weight,
+	})
 	b.mu.Unlock()
 	b.cond.Signal()
 	return nil
 }
 
 // next blocks until a pending item is available and moves it to the in-flight
-// list, returning the item. The sender must later call discard (on ack) or
-// requeue (on reconnect) for every item returned by next.
+// list, returning the item. Every item it returns is later either acknowledged
+// or requeued for replay; see the concurrency model on buffer.
 //
 // Returns errClosed when the buffer has been closed and drained.
 // Returns ctx.Err() if ctx is cancelled while waiting.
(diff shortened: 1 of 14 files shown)


purego/internal/stream/core.go
```diff
@@ -882,20 +1014,57 @@ waitLoop:
...
+		// and replays the remainder.
+		var partial *partialSubmissionFailure
+		waitedForReceiver := false
+		if !receiverReported && errors.As(cause, &partial) {
+			waitedForReceiver = true
+			timer := time.NewTimer(cs.cfg.DrainTimeout)
+			waiting := true
+			for waiting {
+				select {
+				case acknowledged := <-ackProgress:
```
teodordelibasic-db, line 1039: A terminal receive status can be lost once `ackProgress` satisfies this wait. We
then close the transport and wait for `receiverDone`, but never consume the
receiver result published during that close, so `runOnce` returns the earlier
send error. A permanent server rejection can consequently be retried, and an
authentication rejection skips credential invalidation.

Could we preserve a non-cancellation receiver error after reaping the receiver,
using the same arbitration rule as the other sender-first paths?

```go
<-receiverDone
if !receiverReported {
        select {
        case receiverErr := <-receiverExitCh:
                if receiverErr != nil && !errors.Is(receiverErr, context.Canceled) {
                        cause = receiverErr
                }
        default:
        }
}
```

It would be good to cover a partial receipt followed by its prefix ACK, with
`Recv` returning `codes.Unauthenticated` when teardown closes the wire, and
assert that `Flush` returns `codes.Unauthenticated` rather than the send error.

purego/internal/stream/buffer.go
```diff
@@ -266,45 +340,96 @@ func (b *buffer[Req]) discardThrough(offset int64) discardResult {
...
+	progressed := result.count > 0
+	if resolution.PartialOffset >= 0 {
+		if resolution.PartialUnits > b.flight[0].ackedUnits {
+			b.flight[0].ackedUnits = resolution.PartialUnits
+			// Durable progress inside the item refreshes its lack-of-ack budget:
+			// a server acknowledging rows is demonstrably alive, so a batch
+			// larger than one timeout's worth of work must not be torn down. A
+			// duplicate or stale partial makes no progress and leaves the
+			// deadline untouched, so a stalled server cannot postpone recovery.
+			b.flight[0].pendingAt = time.Now()
```
teodordelibasic-db: Refreshing only the head item's `pendingAt` leaves every trailing item on its
original deadline. If items A and B enter flight together and partial ACKs keep A
alive past one timeout, completing A makes B the head with an already-expired
deadline. The timer read at
[`purego/internal/stream/core.go:1352`](https://github.com/databricks/zerobus-sdk/blob/6ea3d65ea69fed709ddc155d72f7cda9c0beb6ce/purego/internal/stream/core.go#L1352)
then tears down a healthy stream immediately, even though durable progress just
arrived.

Since partial progress is intentionally allowed to extend the head item here,
could we carry that extension forward when the next item becomes head, or base
the timeout on the last genuine durability progress instead?

A focused regression case can enqueue two row-count items, partially ACK A before
the timeout, fully ACK A after B's original deadline, and assert that B is not
immediately replayed on a second connection.

purego/internal/stream/core.go
```diff
@@ -1008,28 +1185,68 @@ func (cs *CoreStream[Req, Resp]) sender(
...
+		var receipt SubmissionReceipt
+		if receiptStream, ok := stream.(submissionReceiptStream[Req]); ok {
+			receipt, err = receiptStream.SendWithReceipt(it.payload)
+		} else {
+			err = stream.Send(it.payload)
+			if err == nil {
+				receipt.SubmittedUnits = it.units
+			}
+		}
+		if receipt.SubmittedUnits > it.units {
```
teodordelibasic-db, line 1214: An over-reported receipt or a short successful receipt is a deterministic adapter
contract violation, but these paths create ordinary `fmt.Errorf` values, which
the supervisor treats as retryable. The duplicate checks at
[`purego/internal/stream/core.go:1550`](https://github.com/databricks/zerobus-sdk/blob/6ea3d65ea69fed709ddc155d72f7cda9c0beb6ce/purego/internal/stream/core.go#L1550)
do the same. Recovery therefore reconnects and replays the same payload after an
impossible local result.

Could we classify both validation sites as non-retryable, for example:

```go
err = wrapValidation(fmt.Errorf("stream: invalid submission receipt: ..."))

return rejectSubmission(wrapValidation(fmt.Errorf(
        "stream: invalid submission receipt: ...",
)))
```

The existing `overReportingRowWire` is a deterministic regression seam. With
recovery enabled and a second wire available, the test can assert that the error
is terminal and no second connection receives a replay.

purego/internal/stream/core.go
```diff
@@ -1117,120 +1371,288 @@ func (cs *CoreStream[Req, Resp]) receiver(
...
+
+	// Scratch space for the submitted ranges plus the in-flight one. Models are
+	// forbidden from retaining Ranges, so one buffer can back every ack instead
+	// of allocating a fresh copy each time.
+	var ackScratch []SubmittedRange
+	ackState := func(includeSending bool) AckState {
+		ranges := submitted
+		limit := submittedUnits
+		if includeSending && sending != nil {
+			ackScratch = append(append(ackScratch[:0], submitted...), *sending)
```
teodordelibasic-db: The scratch buffer removes the per-ACK allocation, but this `append` still copies
all unacknowledged ranges whenever an ACK arrives while `Send` is active. Under
continuous ingestion `sending` is usually non-nil, so ACK processing remains
`O(n)` in the outstanding window and can consume substantial memory bandwidth at
the default in-flight limit.

Could we represent the active range separately in `AckState`, or otherwise let
the resolver consume the submitted slice and the one active range without
materializing a combined slice?

zlata-stefanovic-db: Yes, thank you! I went with representing the active range separately in AckState (now holds Active+HasActive boolean). It exposes NumRanges&RangeAt functions to be iterated over. The ResolveAcknowledgedUnits now loops over the submitted and active range in submission order without combining them into a merged slice.

purego/internal/stream/core.go
zlata-stefanovic-db: fixed now, made these non-retryable with wrapValidation, also added the suggested test

purego/internal/stream/buffer.go
zlata-stefanovic-db: Yes, I'll allow the partial progress to extend to the next item as well. Also another case I've noticed - if a partial ack extends progress of the current head, then the full ack of a previous item should refresh the new head's deadline as well, which we do not currently do. Now all progress refreshes the deadline.

Adding the regression cases too.

purego/internal/stream/core.go
zlata-stefanovic-db: Fixed as suggested

purego/internal/stream/core.go
```diff
@@ -882,20 +1014,57 @@ waitLoop:
 		default:
 		}
 	case errors.As(cause, &ps):
```
elenagaljak-db, line 1016: On `main` the pause case was an immediate hard abort (`stream.Close()`). Switching it to `gracefulTeardown`, plus `resolvePendingOnExit(!isPause)` now waiting up to `DrainTimeout` for the in-flight Send instead of aborting it, is a real behavior change on the shipping proto/JSON path — server `CloseStreamSignal` hits this.

Effects for proto/JSON:
- the in-flight Send is allowed to complete (and can become durable if an ack was already buffered) rather than being aborted and replayed;
- the server sees `END_STREAM` instead of an abrupt cancel;
- teardown can take up to ~2x `DrainTimeout` when a Send is slow/stuck — the receiver's `resolvePendingOnExit` wait gates `receiverDone`, and `gracefulTeardown` then waits on `receiverDone` again with its own budget.

It's bounded (no hang) and probably a fine change, but it's not behavior-neutral vs `main`, so the "no user-visible difference" rationale for skipping the changelog doesn't quite hold. Can we confirm it's intended and covered by a pause test that exercises a slow/stuck Send (not just `fakeRPC.CloseSend` closing Recv)?

purego/internal/stream/core.go
```diff
@@ -614,10 +709,21 @@ func (cs *CoreStream[Req, Resp]) GetUnackedBatches() ([][][]byte, error) {
 	items := cs.consolidateUnacked()
 	out := make([][][]byte, 0, len(items))
 	for _, it := range items {
+		payload := it.payload
+		if it.ackedUnits > 0 {
+			var err error
+			payload, err = cs.enc.slice(payload, it.ackedUnits)
```
elenagaljak-db, line 715: This slice-before-decode makes the `GetUnackedBatches` doc comment above (lines 691-696) inaccurate for the generic core: it still says the grouping is "the unit the server acks atomically" and reproduces "the original durability boundaries," but for a record-count protocol the returned group is now the unacked *suffix*, not the original ingest group. Still true for proto/JSON (ackedUnits is always 0), but since this PR introduces the divergence, worth updating the comment here — and the public `Stream.GetUnackedBatches` wording ("grouped by ingest call") too.

elenagaljak-db (commented):
Nice refactor!

purego/internal/stream/core.go
zlata-stefanovic-db: Good catch, fixed both. The internal comment now says an entry is safe to replay as one batch — exactly the ingest call for proto/JSON, only the unacked suffix when a protocol acks part of an item — and the public doc reads "grouped for replay: one group per ingest call, minus any prefix already made durable."

purego/internal/stream/core.go
zlata-stefanovic-db: Added two pause tests with a controlled Send: `TestCoreStreamPauseAwaitsSlowSendAndKeepsAck`
(Send completes after the pause deadline, the buffered ack becomes durable, nothing replays — fails against `main`'s abort) and `TestCoreStreamPauseStuckSendStaysBounded` (Send never returns, with and without a buffered ack: bounded abort, record replays).

elenagaljak-db (approved):

PR #699, opened 2026-08-07
[PureGo] Remove UC descriptor cache

### LINKED ISSUE #676: [PureGo] Improve UC descriptor cache lifecycle

zlata-stefanovic-db:

Removes the Unity Catalog descriptor cache from the PureGo SDK and documents the fetch-once-and-reuse contract in its place.

**WHAT:**

- Delete the cache from `purego/zerobus/sdk.go`: the `dynamicSchemaCache` field, `cachedDescriptor` struct, the 5-minute TTL and 128-entry constants, `getDynamicDescriptorFromCache`, `storeDynamicDescriptor`, the SHA-256 credential-scoped `dynamicSchemaCacheKey`, and the cache reset in `Close`. `fetchProtoDescriptor` now goes straight to UC, and `SDK.mu` guards only the open-stream set and the closed flag.
- Document the contract where a reader will actually hit it: the `FetchProtoDescriptorFromUC` godoc (with a reuse-across-streams snippet and an explicit warning that per-stream calls put a UC round-trip on the ingestion path), a new "Proto descriptors from Unity Catalog" section in the package godoc, and a callout in the README's dynamic-proto section.
- Move the descriptor fetch in `examples/dynamic/single` out of its `openStream` helper and into `main`, since bundling fetch into stream creation is exactly the pattern that now costs a round-trip per stream. The `batch` and `proto` examples already fetched once and only gain a comment.
- Correct the `NEXT_CHANGELOG.md` entry to the current API name and state the fetch-once semantics.

**WHY:**

Per [@andrijast-db's review feedback](https://github.com/databricks/zerobus-sdk/pull/678#issuecomment-5215843901): fetching the descriptor is now an explicit standalone call rather than a step inside stream creation, so a caller can fetch once and reuse the bytes for every stream on a table. That makes the cache's cost/benefit negative — it bought a TTL, an entry limit, a credential-scoped key, and a proposed `RefreshProtoDescriptorFromUC` escape hatch, in exchange for saving a round-trip that correct callers do not make in the first place.

Deleting it also resolves #676 by construction rather than by implementing the lifecycle machinery that issue proposed. There is no stale descriptor to invalidate because every call is fresh, no cache key to coalesce misses on, and no shared in-flight work whose cancellation needs isolating. OAuth token minting is still shared across streams via `auth.SharedTokenCache`, which is unaffected by this change.

This supersedes #678, which is now closed.

Note for reviewers: `FetchProtoDescriptor` remains as a `Deprecated:` alias for `FetchProtoDescriptorFromUC`. Both names landed in #666 and v0.1.0 has not shipped, so it is deprecating a name no user has seen. I left it alone to keep this PR scoped to the cache removal, but I am happy to delete it here if you would rather not carry it into the first release.

## How is this tested?

- `cd purego && go test -race ./...` — all packages pass.
- `cd purego && go vet ./...` and `gofmt -l .` — clean.
- `cd purego/examples && go build ./... && go vet ./...` — clean.

Test changes:

- `TestSDKFetchProtoDescriptor_CacheHit` became `TestSDKFetchProtoDescriptor_FetchesOnEveryCall`. It now asserts two UC schema requests for two fetches, pinning the no-caching behavior instead of the old "1 request" expectation, and still checks that repeated fetches return equivalent descriptors.
- Added `TestSDKFetchProtoDescriptorFromUC_RejectsClosedSDK`. The closed-SDK guard was previously only exercised through the deleted `storeDynamicDescriptor` test, so this keeps it covered against the public API.
- Deleted the four tests whose subject no longer exists: cache-hit counting, credential-scoped cache keys, expired-entry pruning, and the entry limit.

Closes #676

### DIFF
--- a/purego/NEXT_CHANGELOG.md
+++ b/purego/NEXT_CHANGELOG.md
@@ -11,8 +11,10 @@
 - Added opt-in wait-ready stream creation whose context bounds the complete
   first-open process; asynchronous creation now preserves context values while
   detaching caller cancellation.
-- Added `FetchProtoDescriptor` for building protobuf descriptors from Unity
-  Catalog table schemas.
+- Added `FetchProtoDescriptorFromUC` for building protobuf descriptors from
+  Unity Catalog table schemas. Each call issues a fresh Unity Catalog request:
+  fetch the descriptor once and reuse the returned bytes for every stream on
+  that table, and fetch again only when the table schema changes.
 - Added `Stream.IngestJSONOffset` and `Stream.IngestJSONRecordsOffset`. JSON
   streams queue JSON directly; proto streams convert it before ingestion.
 
--- a/purego/README.md
+++ b/purego/README.md
@@ -92,11 +92,19 @@ streams and convert it to protobuf on proto streams.
 Fetch the table schema explicitly, then create a regular proto stream. The
 stream can accept either protobuf bytes or JSON converted at ingest time.
 
+> **Fetch the descriptor once and reuse it.** `FetchProtoDescriptorFromUC`
+> performs a Unity Catalog request on every call and the SDK does not cache the
+> result. Call it once per table at startup, hold on to the returned bytes, and
+> pass them to `WithProto` for every stream you open on that table. Fetch again
+> only when the table schema changes. Calling it per stream — or per record —
+> puts a UC round-trip on your ingestion path.
+
 UC schema conversion rejects nullable array/map fields and collections that
 allow null elements or values because protobuf cannot preserve those
 distinctions. JSON ingestion also rejects explicit `null` collection fields.
 
 ```go
+// Fetch once at startup, then reuse `descriptor` for every stream on this table.
 descriptor, err := sdk.FetchProtoDescriptorFromUC(
     ctx,
     "catalog.schema.table",
--- a/purego/examples/dynamic/batch/main.go
+++ b/purego/examples/dynamic/batch/main.go
@@ -25,6 +25,7 @@ func main() {
 	defer sdk.Close()
 
 	ctx := context.Background()
+	// Fetch the descriptor once; reuse it for every stream on this table.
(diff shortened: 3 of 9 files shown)


danilonajkov-db (approved):
LGTM, I like this approach its better than us internally managing it

PR #728, opened 2026-08-12
[Python] Fix wait_for_offset regression; release 1.6.1

### LINKED ISSUE #726: [Python] wait_for_offset broken in 1.5.0+. PyO3 0.29 upgrade made timeout_sec required
### SDK

Python

### SDK Version

1.5.0

### Operating System

_No response_

## Problem
`ZerobusStream.wait_for_offset(offset)` raises on every call in 1.5.0 and 1.6.0: TypeError: ZerobusStream.wait_for_offset() missing 1 required positional argument: 'timeout_sec'

The SDK's own Python wrapper is the caller, so this breaks through the public API with no user error involved. `python/zerobus/sdk/sync/zerobus_sdk.py:101`: `return self._inner.wait_for_offset(offset)`

## Cause
Regression from the PyO3 0.20 → 0.29 bump in #652. The Rust function is unchanged between 1.4.0 and 1.6.0:
```rust
fn wait_for_offset(&self, py: Python, offset: i64, timeout_sec: Option<f64>) -> PyResult<()> {
    let _ = timeout_sec; // Timeout is handled internally by the Rust SDK
```

PyO3 dropped the rule where `Option<T>` arguments implicitly default to `None`, so the binding's text signature changed without any source change:
| Version | PyO3   | Binding signature                              |
| ------- | ------ | ---------------------------------------------- |
| 1.4.0   | 0.20.3 | `($self, offset, timeout_sec=None)`            |
| 1.5.0   | 0.29.0 | `($self, offset, timeout_sec)`                 |
| 1.6.0   | 0.29.0 | `($self, offset, timeout_sec)`                 |

## Same class of regression elsewhere
These also lost their defaults in 1.5.0 and are reachable if callers use the Rust bindings directly (the Python wrappers happen to always pass `options`, which is the only reason they aren't crashing too):
- `RecordAcknowledgment.wait_for_ack(_timeout_sec)`
- `ZerobusSdk.create_stream(..., options)`
- `ZerobusSdk.create_stream_with_headers_provider(..., options)`
`ZerobusArrowStream.wait_for_offset` (`sync/zerobus_sdk.py:187`) has the identical bug to the record-stream case.

## Suggested fix
Add `#[pyo3(signature = (offset, timeout_sec=None))]` to restore the defaults. This SDK already uses this attribute in 13 other places in `python/rust/src/`.

## Impact 
Production crash-loop on 1.6.0. Any consumer calling `wait_for_offset` is broken on 1.5.0+. Pinned to `<1.5.0` as a workaround.

### 1. Real usage (any Zerobus table)
```python
from zerobus.sdk.shared import RecordType, StreamConfigurationOptions, TableProperties
from zerobus.sdk.sync import ZerobusSdk

sdk = ZerobusSdk(ZEROBUS_ENDPOINT, WORKSPACE_URL)
stream = sdk.create_stream(
  CLIENT_ID,
  CLIENT_SECRET,
  TableProperties("catalog.schema.table"),
  StreamConfigurationOptions(record_type=RecordType.JSON),
)

offset = stream.ingest_records_offset([{"id": 1}])
stream.wait_for_offset(offset)   # <-- raises on 1.5.0 / 1.6.0, fine on <= 1.4.0
```

```text
TypeError: ZerobusStream.wait_for_offset() missing 1 required positional argument: 'timeout_sec'
File ".../zerobus/sdk/sync/zerobus_sdk.py", line 101, in wait_for_offset
  return self._inner.wait_for_offset(offset)
```

### 2. No credentials required
The arity mismatch is visible from the binding's own signature, so this needs only `pip install databricks-zerobus-ingest-sdk`:
```python
import inspect
from importlib.metadata import version
import zerobus._zerobus_core as core

print(version("databricks-zerobus-ingest-sdk"))
print(inspect.signature(core.sync.ZerobusStream.wait_for_offset))
```

```text
# 1.4.0
1.4.0
(self, /, offset, timeout_sec=None)

# 1.6.0
1.6.0
(self, /, offset, timeout_sec)
```

The wrapper at `zerobus/sdk/sync/zerobus_sdk.py:101` passes only `offset`, so the
second signature can never be satisfied through the public API.

### Expected behavior

`stream.wait_for_offset(offset)` should block until the given offset is acknowledged and return `None`, exactly as it did in 1.4.0 and every earlier 1.x release. It is the SDK's own public wrapper making this call, so it should never raise `TypeError` regardless of how the underlying binding is generated. 

More generally, upgrading the PyO3 dependency is an internal implementation detail and should not change the SDK's public Python API in a minor release. `timeout_sec` should remain optional with a default of `None`, particularly since the Rust implementation discards the argument entirely and handles the timeout internally. 

### Is it a regression?

Yes

### Logs

```shell

```

### Additional context

_No response_

elenagaljak-db:
## Summary

Fixes #726 — `ZerobusStream.wait_for_offset(offset)` raised `TypeError: missing 1 required positional argument: 'timeout_sec'` on **every** call through the public sync API on 1.5.0 and 1.6.0.

### Root cause
The PyO3 `0.20 → 0.29` upgrade (#652) dropped the rule where an `Option<T>` argument implicitly defaults to `None`. Without an explicit `#[pyo3(signature = ...)]`, `timeout_sec` silently became a **required** argument in the generated Python signature — with no change to any source line:

| Version | PyO3 | Binding signature |
| --- | --- | --- |
| 1.4.0 | 0.20.3 | `(offset, timeout_sec=None)` |
| 1.5.0 / 1.6.0 | 0.29.0 | `(offset, timeout_sec)` |

The SDK's own wrapper calls `self._inner.wait_for_offset(offset)`, so it could never satisfy the second signature.

### Fix
Restore the optional defaults with `#[pyo3(signature = ...)]` on the four affected **sync** bindings:
- `ZerobusStream.wait_for_offset` (the crash)
- `RecordAcknowledgment.wait_for_ack`
- `ZerobusSdk.create_stream`
- `ZerobusSdk.create_stream_with_headers_provider`

The async surface already carried the attribute and was unaffected. The Rust bodies are untouched (`wait_for_offset` already discards `timeout_sec`), so this is a pure signature restoration back to 1.4.0 behavior — additive, non-breaking.

### Why it shipped unnoticed
No example exercised the broken combination (sync **record** stream + explicit `wait_for_offset`). The JSON/proto examples use `flush()`; the Arrow examples call `wait_for_offset` on the Arrow stream, whose binding never had a `timeout_sec` argument. `tests/test_signatures.py` closes that coverage gap — 8 tests asserting the optional defaults survive, needing neither network nor credentials. Verified they fail on the unfixed binding and pass with the fix.

## Release v1.6.1
Patch on top of the current released line (1.6.0). Bumped the version in `rust/Cargo.toml`, `Cargo.lock`, and `zerobus/__init__.py`; moved `NEXT_CHANGELOG.md` into `CHANGELOG.md` and reset the template.

> Note on versioning: 1.6.1 (not 1.5.1) because this branch is on the already-released 1.6.0 codebase; a 1.5.1 would sort below 1.6.0 and `pip` would never hand it to users on the latest release. A 1.5.x backport, if wanted, is a separate branch off the `python/v1.5.0` tag.

## Test plan
- `python -m unittest tests.test_signatures tests.test_smoke` → 30 passing.
- Negative control: reverted the binding fix, rebuilt → the 5 sync signature tests fail with the exact reported `TypeError`; the 3 async tests stay green.

### DIFF
--- a/python/rust/src/sync_wrapper.rs
+++ b/python/rust/src/sync_wrapper.rs
@@ -50,6 +50,7 @@ pub struct RecordAcknowledgment {
 impl RecordAcknowledgment {
     /// Wait for the acknowledgment and return the offset ID.
     /// This method can only be called once.
+    #[pyo3(signature = (_timeout_sec = None))]
     pub fn wait_for_ack(&mut self, py: Python, _timeout_sec: Option<f64>) -> PyResult<i64> {
         if self.done {
             return Err(PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(
@@ -192,6 +193,7 @@ impl ZerobusStream {
     }
 
     /// Wait for a specific offset to be acknowledged
+    #[pyo3(signature = (offset, timeout_sec = None))]
     fn wait_for_offset(&self, py: Python, offset: i64, timeout_sec: Option<f64>) -> PyResult<()> {
         let _ = timeout_sec; // Timeout is handled internally by the Rust SDK
         let stream = self.inner.clone();
@@ -338,6 +340,7 @@ impl ZerobusSdk {
     }
 
     /// Create a new stream with OAuth authentication
+    #[pyo3(signature = (client_id, client_secret, table_properties, options = None))]
     fn create_stream(
         &self,
         py: Python,
@@ -369,6 +372,7 @@ impl ZerobusSdk {
     }
 
     /// Create a new stream with custom headers provider
+    #[pyo3(signature = (table_properties, headers_provider, options = None))]
     fn create_stream_with_headers_provider(
         &self,
         py: Python,
--- a/python/CHANGELOG.md
+++ b/python/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Version changelog
 
+## Release v1.6.1
+
+### Bug Fixes
+
+- Fix `ZerobusStream.wait_for_offset(offset)` raising `TypeError: missing 1 required positional argument: 'timeout_sec'` on the synchronous record stream (issue #726). The PyO3 0.20 → 0.29 upgrade dropped the implicit `None` default for `Option<T>` arguments, which silently made `timeout_sec` required. Restored the default via `#[pyo3(signature = ...)]`. The same fix restores optional arguments on `RecordAcknowledgment.wait_for_ack`, `ZerobusSdk.create_stream`, and `ZerobusSdk.create_stream_with_headers_provider`.
+
+### Documentation
+
+- Document `stream_paused_max_wait_time_ms` on `ArrowStreamConfigurationOptions` in the Arrow type stub so it matches the already-supported runtime option.
+
 ## Release v1.6.0
 
 ### New Features and Improvements
--- a/python/NEXT_CHANGELOG.md
+++ b/python/NEXT_CHANGELOG.md
@@ -10,8 +10,6 @@
 
 ### Documentation
 
-- Document `stream_paused_max_wait_time_ms` on `ArrowStreamConfigurationOptions` in the Arrow type stub so it matches the already-supported runtime option.
-
 ### Internal Changes
 
 ### Breaking Changes
--- a/python/rust/Cargo.lock
+++ b/python/rust/Cargo.lock
@@ -2820,7 +2820,7 @@ dependencies = [
 
 [[package]]
 name = "zerobus-python"
-version = "1.6.0"
+version = "1.6.1"
 dependencies = [
  "arrow-array",
  "arrow-ipc",
--- a/python/rust/Cargo.toml
+++ b/python/rust/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zerobus-python"
-version = "1.6.0"
+version = "1.6.1"
 edition = "2021"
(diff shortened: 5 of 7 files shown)


python/rust/src/sync_wrapper.rs
```diff
@@ -50,6 +50,7 @@ pub struct RecordAcknowledgment {
 impl RecordAcknowledgment {
     /// Wait for the acknowledgment and return the offset ID.
     /// This method can only be called once.
+    #[pyo3(signature = (_timeout_sec = None))]
```
teodordelibasic-db, line 53: Here we expose `_timeout_sec`, but the stub and example document `timeout_sec`. Thus `ack.wait_for_ack(timeout_sec=30)` still raises `TypeError`. The new test incorrectly locks in `_timeout_sec` at `python/tests/test_signatures.py:36`. Rename the Rust parameter and test the documented keyword.

teodordelibasic-db: As discussed offline, this would be a breaking change.

teodordelibasic-db (approved):

PR #729, opened 2026-08-12
[All] Correct SDK documentation and examples

teodordelibasic-db:

Correct README snippets, checked-in examples, and public API doc examples across Rust, Python, Java, Go, pure Go, TypeScript, C++, and .NET so they compile and demonstrate the current APIs. The changes fix API names and arguments, stream format selection, dependency setup, recovery and callback semantics, resource cleanup, and queue-then-flush ingestion. They also add the generated-message fixture needed to run the .NET Protobuf example.

The documentation had drifted as the public APIs evolved, leaving examples that either failed to compile or demonstrated incorrect recovery, durability, and throughput patterns. Keeping these examples aligned with the supported APIs prevents users from copying invalid or unnecessarily slow client code.

This PR documents current behavior. The following implementation issues are deferred:

- Python `_zerobus_core.pyi` still has wrong `get_unacked_records()` return types, `create_stream` argument order, async signatures, and is missing row-stream `get_unacked_batches()`.
- Python `stream_id` and `get_state()` still return placeholder values (`stream-placeholder-id` and constant `OPENED`).
- TypeScript custom header callbacks are invoked once and snapshotted; they do not refresh on recovery.
- Java JNI logs `AckCallback` exceptions but does not clear the pending Java exception, which can poison later callback operations.
- Java `setRecoveryBackoffMs` does not reject negative values.
- Java `NativeTestHelper` remains a public production class despite exposing test-only JNI symbols.
- Python `ingest_record_nowait` / `ingest_records_nowait` are documented as unsafe but do not emit a deprecation warning.
- A flush timeout can leave a Pure-Go or C++ stream active, so unacked retrieval fails until the stream is actually closed. Recovering that case needs an implementation change.
- `go/version.go` remains `1.4.0`. The user-agent string bumps with the next Go release, not this docs PR.

## How is this tested?

All runnable examples and executable documentation snippets were compiled or type-checked. Representative JSON, Protobuf, Arrow, batch, recovery, and custom-header paths were also exercised end to end.

The affected SDK formatters, linters, unit tests, example builds, and doc tests pass. The .NET suite passes 47 unit and 54 integration tests. TypeScript build, tests, and example type-check pass; its all-features Clippy check still reports three pre-existing warnings unrelated to these documentation changes. The C++ Arrow example was source-reviewed but not linked because Apache Arrow C++ was unavailable; the other C++ examples built and their JSON batch and custom-header paths were exercised.

### DIFF
--- a/dotnet/README.md
+++ b/dotnet/README.md
@@ -41,10 +41,10 @@ stream.Flush();
 
 ## Installation
 
-### NuGet
+### NuGet (when published)
 
 ```bash
-dotnet add package Databricks.Zerobus --version 0.5.1
+dotnet add package Databricks.Zerobus.Ingest.Sdk
 ```
 
 ### From Source
--- a/dotnet/README.md
+++ b/dotnet/README.md
@@ -4,8 +4,8 @@ High-performance .NET SDK for streaming data ingestion into Databricks Delta tab
 
 ## Requirements
 
-- **.NET 8** or **.NET 10**
-- **Rust toolchain** (for building the native `zerobus_ffi` library from source)
+- Consumers: .NET 8 or .NET 10
+- Building from source: .NET 10 SDK and a Rust toolchain
 
 ## Quick Start
 
@@ -31,19 +31,20 @@ using var stream = sdk.CreateJsonStream(
     clientSecret,
     options);
 
-// 4. Ingest records.
-long offset = stream.IngestRecord("""{"id": 1, "message": "Hello"}""");
-
-// 5. Wait for acknowledgment.
-stream.WaitForOffset(offset);
+// 4. Queue records, then confirm the whole run with one flush.
+for (int id = 1; id <= 100; id++)
+{
+    stream.IngestRecord($$"""{"id": {{id}}, "message": "Hello"}""");
+}
+stream.Flush();
 ```
 
 ## Installation
 
-### NuGet (when published)
+### NuGet
 
 ```bash
-dotnet add package Databricks.Zerobus.Ingest.Sdk
+dotnet add package Databricks.Zerobus --version 0.5.1
 ```
 
 ### From Source
@@ -187,13 +188,26 @@ If you use this untyped API, JSON streams must set `RecordType.Json` and use the
 Proto streams must provide `DescriptorProto` and use the byte-oriented overloads.
 
 ```csharp
-// JSON
-long offset = stream.IngestRecord("""{"field": "value"}""");
+// JSON stream
+var jsonOptions = options with { RecordType = RecordType.Json };
+using var jsonStream = sdk.CreateStream(
+    new TableProperties("catalog.schema.json_table"),
+    clientId,
+    clientSecret,
+    jsonOptions);
+long jsonOffset = jsonStream.IngestRecord("""{"field": "value"}""");
+jsonStream.WaitForOffset(jsonOffset);
 
-// Protobuf
+// Protobuf stream
+var protoOptions = options with { RecordType = RecordType.Proto };
 byte[] protoBytes = myMessage.ToByteArray();
-long offset = stream.IngestRecord(protoBytes);
-stream.WaitForOffset(offset);
+using var protoStream = sdk.CreateStream(
+    new TableProperties("catalog.schema.proto_table", descriptorProto),
+    clientId,
+    clientSecret,
+    protoOptions);
+long protoOffset = protoStream.IngestRecord(protoBytes);
+protoStream.WaitForOffset(protoOffset);
 ```
 
 #### `IngestRecords`
--- a/dotnet/README.md
+++ b/dotnet/README.md
@@ -219,8 +219,8 @@ string[] records = [
     """{"device": "sensor-001", "temp": 20}""",
     """{"device": "sensor-002", "temp": 21}""",
 ];
-long lastOffset = stream.IngestRecords(records);
-stream.WaitForOffset(lastOffset);
+long batchOffset = stream.IngestRecords(records);
+stream.WaitForOffset(batchOffset);
 ```
 
 #### `WaitForOffset` (sync)
@@ -241,14 +241,26 @@ stream.Flush();
 
 #### `GetUnackedRecords`
 
-Retrieves unacknowledged records after stream failure (call after close/failure only).
+Retrieves unacknowledged records after stream failure (call after close/failure only). A flush timeout can leave the stream active, in which case `GetUnackedRecords()` throws until the stream closes.
 
 ```csharp
-ReadOnlyMemory<byte>[] unacked = stream.GetUnackedRecords();
-foreach (var payload in unacked)
+try
 {
-    // Decode as UTF-8 if you know this stream ingests JSON.
-    Console.WriteLine($"record bytes: {payload.Length}");
+    stream.Flush();
+}
+catch (ZerobusException)
+{
+    // A flush timeout can leave the stream active. GetUnackedRecords
+    // requires a closed or failed stream.
+    try
+    {
+        var unacked = stream.GetUnackedRecords();
+        Console.WriteLine($"Failed to acknowledge {unacked.Length} records");
+    }
+    catch (ZerobusException retrieval)
+    {
+        Console.WriteLine($"Could not inspect unacked records (stream may still be active): {retrieval.Message}");
+    }
 }
 ```
 
--- a/go/zerobus.go
+++ b/go/zerobus.go
@@ -533,6 +533,9 @@ func (st *ZerobusStream) IngestRecordOffset(payload interface{}) (int64, error) 
 //	if err != nil {
 //	    log.Fatal(err)
 //	}
+//	if err := stream.Flush(); err != nil {
+//	    log.Fatal(err)
+//	}
 func (st *ZerobusStream) IngestRecordNowait(payload interface{}) error {
 	if st.ptr == nil {
 		return &ZerobusError{Message: "Stream has been closed", IsRetryable: false}
@@ -569,6 +572,12 @@ func (st *ZerobusStream) IngestRecordNowait(payload interface{}) error {
 //	    `{"field": "value1"}`,
 //	    `{"field": "value2"}`,
 //	})
+//	if err != nil {
+//	    log.Fatal(err)
+//	}
+//	if err := stream.Flush(); err != nil {
+//	    log.Fatal(err)
+//	}
 func (st *ZerobusStream) IngestRecordsNowait(records []interface{}) error {
 	if st.ptr == nil {
 		return &ZerobusError{Message: "Stream has been closed", IsRetryable: false}
@@ -638,6 +647,9 @@ func (st *ZerobusStream) IngestRecordsNowait(records []interface{}) error {
 //	}
 //	batchOffset, err := stream.IngestRecordsOffset(records)
 //	if err != nil {
+//	    log.Fatal(err)
+//	}
+//	if err := stream.Flush(); err != nil {
 //	    log.Fatal(err)
 //	}
 //	log.Printf("Batch ingested with offset: %d", batchOffset)
@@ -728,7 +740,7 @@ func (st *ZerobusStream) WaitForOffset(offset int64) error {
 //
 // Use this method to:
 //   - Retrieve unacknowledged records after stream failure for retry logic
-//   - Check which records weren't durably written after Close() fails
+//   - Inspect payloads that were not durably written after a failed Flush()
 //   - Implement custom retry strategies after stream errors
 //
 // Returns a slice where each element is either:
--- a/go/zerobus.go
+++ b/go/zerobus.go
@@ -783,7 +783,9 @@ func (st *ZerobusStream) GetUnackedRecords() ([]interface{}, error) {
 // Example:
 //
 //	for _, r := range records {
-//	    stream.IngestRecordOffset(r)
+//	    if _, err := stream.IngestRecordOffset(r); err != nil {
+//	        log.Printf("Ingest failed: %v", err)
+//	    }
 //	}
 //	if err := stream.Flush(); err != nil {
 //	    log.Printf("Flush failed: %v", err)
--- a/go/zerobus.go
+++ b/go/zerobus.go
@@ -6,13 +6,14 @@
 //
 // # Installation
 //
-// This package is a CGO wrapper around a Rust core. Tagged releases include
-// pre-built libraries, so consumers can install with:
+// This package is a CGO wrapper around a Rust core. Tagged releases and
+// checkouts that include lib/ archives do not need Rust. Consumers can install
+// with:
 //
 //	go get github.com/databricks/zerobus-sdk/go@v1.4.0
 //
 // Prerequisites for consumers: Go 1.21+, CGO enabled, a C compiler.
-// Rust and `go generate` are required only when building from source.
+// Rust and `go generate` are required only when rebuilding the FFI.
 //
 // # Quick Start
 //
@@ -98,7 +99,7 @@
 //
 // Errors are categorized as retryable or non-retryable:
 //
-//	ack, err := stream.IngestRecord(data)
+//	_, err := stream.IngestRecordOffset(data)
 //	if err != nil {
 //	    if zbErr, ok := err.(*zerobus.ZerobusError); ok {
 //	        if zbErr.Retryable() {
--- a/go/zerobus.go
+++ b/go/zerobus.go
@@ -740,7 +740,7 @@ func (st *ZerobusStream) WaitForOffset(offset int64) error {
 //
 // Use this method to:
 //   - Retrieve unacknowledged records after stream failure for retry logic
-//   - Inspect payloads that were not durably written after a failed Flush()
+//   - Inspect payloads that were queued but not acked, before Close()
 //   - Implement custom retry strategies after stream errors
 //
 // Returns a slice where each element is either:
--- a/go/zerobus.go
+++ b/go/zerobus.go
@@ -6,12 +6,13 @@
 //
 // # Installation
 //
-// This package requires a one-time build step to compile the Rust FFI layer:
+// This package is a CGO wrapper around a Rust core. Tagged releases include
+// pre-built libraries, so consumers can install with:
 //
-//	go get github.com/databricks/zerobus-sdk/go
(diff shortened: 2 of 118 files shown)


python/examples/sync_example_json.py
```diff
@@ -228,15 +229,15 @@ def main():
             # ack = stream.ingest_record(record_dict)  # Deprecated
             # offset = ack.wait_for_ack()  # Extra step needed
 
-            end_time = time.time()
```
nikolaobradovic-db, line 231: ingest_record_nowait() and ingest_records_nowait() spawn detached tasks and discard enqueue errors. flush() can complete before those tasks allocate offsets, so this example may report durability while losing submissions.

@teodordelibasic-db @elenagaljak-db I think we should remove nowait from examples and put it on a deprecation path.

python/README.md
```diff
@@ -327,56 +328,51 @@ stream = sdk.create_stream(client_id, client_secret, table_properties, options)
...
 
 try:
     for i in range(10000):
         stream.ingest_record_offset(record)
-    stream.flush()
-except NonRetriableException as e:
-    unacked = stream.get_unacked_records()  # Returns List[bytes]
+    stream.close()
+except ZerobusException as e:
+    unacked = list(stream.get_unacked_records())
```
nikolaobradovic-db: This catch also handles immediate enqueue errors that leave the stream active. Calling get_unacked_records() or recreate_stream() then fails and masks the original error.

Please separate enqueue failures from terminal failures, close before recovery, and protect the replacement stream with try/finally.

python/zerobus/sdk/shared/config.py
```diff
@@ -11,15 +11,16 @@
...
 # Add module-level documentation
 AckCallback.__doc__ = """
-Base class for record acknowledgment callbacks.
+Base class for logical ingest submission acknowledgment callbacks.
 
-Subclass this in Python to create custom callbacks that are invoked
-when records are acknowledged by the server or encounter errors.
+Subclass this in Python to create custom callbacks that are invoked once per
+logical ingest submission. A batch submission produces one callback, not one
+callback per record in the batch.
```
nikolaobradovic-db: Pre-queue validation, size, type, and closed-stream failures do not generate callbacks, so this does not apply to every attempted submission.
Please qualify it as one callback per successfully queued submission that later acknowledges or fails.

python/README.md
```diff
@@ -327,56 +328,51 @@ stream = sdk.create_stream(client_id, client_secret, table_properties, options)
...
+    new_stream = sdk.recreate_stream(stream)
     new_stream.flush()
     new_stream.close()
 ```
 
-Use `get_unacked_batches()` for batch-level retry:
+Use `get_unacked_batches()` to inspect the original batch grouping after the stream closes:
 
 ```python
-unacked_batches = stream.get_unacked_batches()  # Returns List[List[bytes]]
```
nikolaobradovic-db, line 377: This ranks per-record ingest_record_nowait() highest but omits the batch APIs. Batch ingestion amortizes the Python→Rust crossing and is the preferred hot path. Additionally, detached nowait submissions are not safely synchronized with flush.

Please include the batch APIs and recommend ingest_records_offset() plus one flush for reliable bulk ingestion.

nikolaobradovic-db: Table at 385

typescript/src/headers_provider.ts
```diff
@@ -42,24 +42,20 @@ export interface HeadersProvider {
  * );
  * ```
  *
- * **How to use custom authentication (PAT, etc.):**
```
nikolaobradovic-db, line 45: The exported interface requires async getHeaders(): Promise<...>, while this example and createStream() require synchronous getHeadersCallback. A class implementing the public interface cannot be passed to the documented API.

Please expose and document one provider type matching createStream().

typescript/src/lib.rs
```diff
@@ -1119,7 +1118,8 @@ impl ZerobusSdk {
     ///
     /// # Arguments
     ///
-    /// * `stream` - The failed or closed stream to recreate
+    /// * `stream` - The terminally failed stream to recreate. The TypeScript wrapper
+    ///   must not have been closed because `close()` releases its native handle.
```
nikolaobradovic-db, line 1121: LLM find:
TypeScript custom credentials cannot refresh
Affected: typescript/src/lib.rs:1024-1046

The callback runs once during stream creation and its result is stored in StaticHeadersProvider. Long-running streams therefore reuse the original token during recovery, so rotating or expiring custom credentials eventually break reconnection.

Keep the threadsafe callback in a HeadersProvider adapter and invoke it whenever the Rust core requests fresh headers. Support invalidation if required by the core contract.

typescript/README.md
```diff
@@ -682,11 +696,7 @@ try {
...
     console.error('Ingestion failed:', error);
 
-    // When stream fails, close it first
-    await stream.close();
-    console.log('Stream closed after error');
-
-    // Optional: Inspect what needs recovery (must be called on closed stream)
+    // Optional: Inspect what needs recovery after a terminal stream failure.
     const unackedBatches = await stream.getUnackedBatches();
     console.log(`Batches to recover: ${unackedBatches.length}`);
```
nikolaobradovic-db: getUnackedBatches() runs before cleanup protection, and this catch can also handle non-terminal failures. An inspection failure can leak the original stream and mask the ingestion error.

Please use an outer finally, recover only from confirmed terminal failures, and close every replacement stream in a nested finally.

typescript/examples/proto/batch.ts
```diff
@@ -105,9 +105,9 @@ async function main() {
 
         // 1. Auto-encoding: array of Message objects - SDK handles encoding
         const batch1 = [
-            AirQuality.create({ device_name: 'sensor-001', temp: 22, humidity: 65 }),
-            AirQuality.create({ device_name: 'sensor-002', temp: 23, humidity: 67 }),
-            AirQuality.create({ device_name: 'sensor-003', temp: 24, humidity: 69 })
+            AirQuality.create({ deviceName: 'sensor-001', temp: 22, humidity: 65 }),
+            AirQuality.create({ deviceName: 'sensor-002', temp: 23, humidity: 67 }),
+            AirQuality.create({ deviceName: 'sensor-003', temp: 24, humidity: 69 })
```
nikolaobradovic-db, line 110: typescript/examples/proto/batch.ts:113-116

Waiting here before queueing the next batch serializes the example into one server round trip per batch.

Please queue all demonstration batches first, then call flush() once or wait only on the final offset. The JSON batch example has the same issue.

typescript/examples/json/README.md
nikolaobradovic-db, line 65: typescript/examples/json/README.md:65-71

This first “Recommended” pattern immediately waits after one ingest. Although valid for strict low-volume confirmation, it is not the default pattern readers should copy.

Please show loop-then-flush() first and move this into a clearly labeled low-volume section. The Protobuf README has the same ordering issue.

typescript/README.md
```diff
@@ -160,6 +161,12 @@ try {
 } finally {
     await stream.close();
 }
+
+}
+
+main().catch((error) => {
+  console.error('Fatal error:', error);
+});
```
nikolaobradovic-db, line 170: This logs a fatal error but leaves the process exit status as zero, so copied CLI or CI code can report success after ingestion fails.

Please set process.exitCode = 1 or rethrow.

nikolaobradovic-db:
I don't see any changes to GO SDK while it was mentioned in PR desc. Maybe you audited it and there were no stale docs/examples?

python/zerobus/__init__.py
```diff
@@ -26,7 +26,7 @@
     ... )
     >>>
     >>> # New optimized API
-    >>> offset = stream.ingest_record_offset(b"data")
+    >>> offset = stream.ingest_record_offset('{"value": "data"}')
```
nikolaobradovic-db, line 29: python/zerobus/__init__.py:13

Callbacks represent logical submissions and can correspond to batches, so “Record acknowledged” is too specific.

Please use “Submission acknowledged” for consistency.

teodordelibasic-db:
Thanks. Fixed docs/examples issues, leaving code defects as follow-up:

Go / pure Go
- Recovery docs now require GetUnackedRecords() before Close().
- Consumer install no longer requires Rust or go generate. Go 1.21+ is the documented minimum.
- make build-go is documented as depending on the Rust FFI build.
- RecordAck.Await() is documented as blocking for durability; Offset() is the immediate handle.
- Concurrency guidance matches the API: one stream can be used from multiple goroutines.
- Copyable snippets call Flush().
- Pure-Go flush-timeout recovery only inspects unacked when retrieval succeeds.
- Pure-Go batch examples treat a batch as one callback event and wait for it before exit.
- Batch examples name the returned offset batchOffset.
- Documented minimum stays Go 1.21+. Example/test modules keep the CI toolchain versions (1.24/1.25); lowering them makes `go fmt` demand tidy.

Java
- Migration examples queue then flush/wait once.
- flush() is documented as a durability barrier, not a callback drain.
- skipNativeLibCheck is documented as not packaging JNI libs.
- macOS is documented as source-build only for the current JNI release set.
- Java 8 is runtime; source builds need JDK 11.
- Proto examples generate AirQualityProto.java with protoc; it is not checked in.
- GenerateProto no longer claims STRUCT support; JAR examples use zerobus-ingest-sdk 1.3.0.
- recoveryRetries default is 4; maxInflightRecords default is 1,000,000.
- Stream-builder snippets use try-with-resources.
- Thread-safety constraints are documented.
- Maven Central snippets no longer tell users to redeclare compile-scope transitives.

Python / TypeScript
- TypeScript omitting descriptorProto does not select JSON.
- Inflight default documented as 1,000,000, including the binding rustdoc.
- close() is still required; GC does not flush.

Rust / C++ / .NET / FFI
- Rust batch examples and primary rustdoc queue then flush/wait once.
- Removed ingest_record()/ingest_records() names from rustdoc and example READMEs.
- C++ recovery treats unacked retrieval failure as an active stream.
- .NET recovery and JSON example no longer treat a failed ingest as success; snippets include Flush().
- FFI README shows flush, close/free, SDK free, and error-string cleanup.
- C++ Arrow schema validation is documented at stream creation.
- generate_files README quoting is valid shell.
- .NET CONTRIBUTING.md points at src/Zerobus/Native/.

Deferred as code, not docs:

- TypeScript custom headers are snapshotted once; the callback is not invoked again on recovery.
- JNI does not clear pending AckCallback exceptions.
- Java setRecoveryBackoffMs javadoc claims it rejects negatives; the setter does not validate.
- Python _zerobus_core.pyi stubs (get_unacked_records() -> int, create_stream argument order, missing/wrong declarations).
- Python placeholder stream IDs and constant OPENED state.
- Putting nowait on a real deprecation path (DeprecationWarning). Examples already dropped it; docs call it unsafe.
- go/version.go is still 1.4.0. That bumps with the v1.5.0 release, not this docs PR.

teodordelibasic-db:
> I don't see any changes to GO SDK while it was mentioned in PR desc. Maybe you audited it and there were no stale docs/examples?

Go did have stale docs. Follow-up commits on this PR cover recovery-before-close, consumer install (no Rust / go generate), Flush() in snippets, Await() vs Offset(), concurrency, and the example go.mod / batchOffset nits.

python/examples/sync_example_json.py
teodordelibasic-db: Removed nowait from the examples. Docs now rank ingest_records_offset() + one flush() as the bulk path and call nowait unsafe (detached tasks, not synchronized with flush). Putting the APIs on a real deprecation path (DeprecationWarning) is a follow-up, not this docs PR.

python/README.md
teodordelibasic-db: Updated. The snippet now distinguishes that enqueue failures can leave the stream active, closes before get_unacked_records()/recreate_stream(), re-raises the original ingest error if unacked retrieval fails, and wraps the replacement stream in try/finally.

python/zerobus/sdk/shared/config.py
teodordelibasic-db: Qualified: one callback per successfully queued logical ingest submission that later acknowledges or fails. Pre-queue validation / size / type / closed-stream failures raise immediately and do not generate a callback.

python/README.md
teodordelibasic-db: Table now ranks ingest_records_offset() + one flush() first. nowait is listed as unsafe and is not recommended for bulk ingestion.

typescript/src/headers_provider.ts
teodordelibasic-db: The exported HeadersProvider is now the shape createStream() actually accepts: getHeadersCallback returning header tuples synchronously. The docs no longer show an async getHeaders() class.

typescript/src/lib.rs
teodordelibasic-db: Agreed, this is a real implementation bug. Out of scope for this docs PR. The headers-provider docs now say the callback is invoked once at stream creation and that token refresh is not wired through. Follow-up will keep the callback and invoke it whenever the core requests headers.

typescript/README.md
teodordelibasic-db: Updated: outer finally always closes the original stream, recovery only runs after a failure, and the replacement stream is closed in a nested finally.

typescript/examples/proto/batch.ts
teodordelibasic-db: Proto and JSON batch examples now queue every demonstration batch, then flush() once.

typescript/examples/json/README.md
teodordelibasic-db: JSON and Protobuf example READMEs now lead with loop-then-flush(). Immediate waitForOffset after a single ingest is labeled as a low-volume confirmation pattern, not the default.

typescript/README.md
teodordelibasic-db: main().catch now sets process.exitCode = 1.

python/zerobus/__init__.py
teodordelibasic-db: Changed to “Submission acknowledged”.

purego/examples/json/batch/main.go
zlata-stefanovic-db: Maybe we can delete "waits on the batch offset" here since that call was deleted. Can we update it to what it does now - flush once, then wait for the batch's single ack callback?

purego/examples/json/batch/main.go
```diff
@@ -20,18 +20,27 @@ import (
...
 
-func (o *ackObserver) OnAck(offset int64) { o.acked.Add(1) }
+func (o *ackObserver) OnAck(offset int64) {
+	o.offset.Store(offset)
+	o.acked.Add(1)
+}
 
 func (o *ackObserver) OnError(offset int64, err error) {
+	o.failed.Store(true)
 	log.Printf("record at offset %d failed: %v", offset, err)
```
zlata-stefanovic-db: Maybe we can say "batch at offset %d failed" or "logical submission at offset %d failed" like in the other changes in this PR

zlata-stefanovic-db (changes requested):

dotnet/README.md
```diff
@@ -31,19 +31,20 @@ using var stream = sdk.CreateJsonStream(
...
 ```
 
 ## Installation
 
-### NuGet (when published)
+### NuGet
 
 ```bash
-dotnet add package Databricks.Zerobus.Ingest.Sdk
+dotnet add package Databricks.Zerobus --version 0.5.1
```
elenagaljak-db: Why 0.5.1? Databricks.Zerobus is incorrect

go/zerobus.go
elenagaljak-db, line 745: The important note and the second bullet contradict

python/NEXT_CHANGELOG.md
```diff
@@ -10,6 +10,15 @@
 
 ### Documentation
 
+- Corrected README, example, and docstring snippets for record-format selection,
+  exception handling, recovery, iterator return values, custom headers, async
+  contexts, and durability-aware throughput measurement.
+- Documented that `get_unacked_records()` and `recreate_stream()` require a closed
+  stream, and that enqueue failures must be closed before recovery.
```
elenagaljak-db: This contradicts readme, which says that "An enqueue failure leaves the stream active". The failed record was never queued, so there is nothing to recover?

dotnet/README.md
```diff
@@ -31,19 +31,20 @@ using var stream = sdk.CreateJsonStream(
...
+for (int id = 1; id <= 100; id++)
+{
+    stream.IngestRecord($$"""{"id": {{id}}, "message": "Hello"}""");
+}
+stream.Flush();
 ```
 
 ## Installation
 
-### NuGet (when published)
```
zlata-stefanovic-db, line 43: Is it published yet? Maybe better to leave it like this?

purego/examples/json/batch/main.go
teodordelibasic-db: Updated. The header now says it flushes once, then waits for the batch's single ack callback.

purego/examples/json/batch/main.go
teodordelibasic-db: Changed to "batch at offset %d failed".

dotnet/README.md
teodordelibasic-db: Not published yet. Restored "NuGet (when published)" and dropped the version pin.

dotnet/README.md
teodordelibasic-db: Fixed. The package id is Databricks.Zerobus.Ingest.Sdk. Left it unpinned under "NuGet (when published)" since it is not on NuGet yet.

go/zerobus.go
teodordelibasic-db: Good catch. A flush timeout can leave the stream active, so that bullet was wrong. It now says to inspect queued-but-unacked payloads before Close(), matching the IMPORTANT note.

python/NEXT_CHANGELOG.md
teodordelibasic-db: Agreed. An enqueue failure never queued that payload, so there is nothing to recover for that call. The changelog now matches the README: close first only if you need to inspect records that were already accepted.

zlata-stefanovic-db (dismissed):

elenagaljak-db (dismissed):

nikolaobradovic-db (dismissed):

nikolaobradovic-db (approved):

PR #730, opened 2026-08-13
[Java] Preserve unacked data on close

danilonajkov-db:
## Summary
- retain closed Rust streams until Java caches unacknowledged recovery data
- propagate recovery failures instead of silently replacing recovery data with empty lists
- preserve recovery data even when stream close reports a flush failure

## Testing
Added a test, verified it fails before the changes and passes after

### DIFF
--- a/java/NEXT_CHANGELOG.md
+++ b/java/NEXT_CHANGELOG.md
@@ -13,6 +13,10 @@
 - Arrow builders now reject unsupported ACK callbacks instead of silently
   discarding them. Configuring `ackCallback` before calling `ArrowStreamBuilder.build()`
   throws `IllegalStateException`.
+- Fixed proto, JSON, and Arrow stream recovery losing unacknowledged data during `close()`. Closed
+  native streams now remain available until Java has cached their recovery records or batches, and
+  recreation releases retained source streams after copying their recovery data. Failed recreation
+  attempts also release their temporary native streams.
 
 ### Documentation
 
--- a/java/src/main/java/com/databricks/zerobus/BaseZerobusStream.java
+++ b/java/src/main/java/com/databricks/zerobus/BaseZerobusStream.java
@@ -117,23 +117,42 @@ abstract class BaseZerobusStream implements AutoCloseable {
   public void close() throws ZerobusException {
     long handle = nativeHandle;
     if (handle != 0) {
-      // Close the stream first (flushes pending records)
-      nativeClose(handle);
-
-      // Cache unacked records before destroying the handle (for recreateStream)
+      ZerobusException closeFailure = null;
       try {
-        cachedUnackedRecords = nativeGetUnackedRecords(handle);
-        cachedUnackedBatches = nativeGetUnackedBatches(handle);
-      } catch (Exception e) {
-        logger.warn("Failed to cache unacked records: {}", e.getMessage());
-        cachedUnackedRecords = new ArrayList<>();
-        cachedUnackedBatches = new ArrayList<>();
+        // Close the stream first (flushes pending records).
+        nativeClose(handle);
+      } catch (ZerobusException e) {
+        // Closing marks the Rust stream closed even when flushing fails. Keep going so its
+        // unacknowledged records can be recovered before the native handle is destroyed.
+        closeFailure = e;
       }
 
-      // Now destroy the handle
+      // Cache unacked records before destroying the handle (for recreateStream)
+      List<byte[]> unackedRecords;
+      List<EncodedBatch> unackedBatches;
+      try {
+        unackedRecords = nativeGetUnackedRecords(handle);
+        unackedBatches = nativeGetUnackedBatches(handle);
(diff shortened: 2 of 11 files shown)


nikolaobradovic-db (dismissed):

danilonajkov-db:
@elenagaljak-db 
```
Finding 1: recreateStream on a cache-failed close leaks the old native handle
Severity: High
File: BaseZerobusStream.java close(), ZerobusSdk.java recreateStream / recreateArrowStream
Issue: If nativeGetUnacked* fails, Java keeps nativeHandle != 0 so recovery can be retried. After nativeClose, Rust is_closed() is already true, so Java isClosed() is true. recreateStream only checks isClosed(), reads unacked data from the live native stream, creates a new stream, and never destroys the old handle. discardFailedRecreation() only destroys the new stream.
Impact: The documented recovery path is close() then recreateStream. After this PR, that path can recover data and leak the original native stream (gRPC tasks, landing zone, JNI handle) until process exit. There is no Java finalizer.
Suggested fix: After a successful native unacked read in recreate*, destroy the old handle (or call a finalizeRecovery() that caches + nativeDestroy). Alternatively, treat nativeHandle == 0 as the recreate precondition so callers must retry close() until it finishes destroy.
Finding 2: Deprecated recreateStream(ZerobusStream) still returns empty unacked data
Severity: High
File: ZerobusSdk.java (deprecated recreateStream)
Issue: Proto/JSON/Arrow recreate uses getUnackedRecords() / getUnackedBatches(), which talk to native while nativeHandle != 0. The deprecated path uses getCachedUnackedRecords(), which is still null after a cache failure and therefore returns new ArrayList<>().
Impact: Legacy callers that catch close() and call recreateStream still silently drop unacked records. The changelog says proto/JSON/Arrow recovery is fixed; this overload is not.
Suggested fix: Use getUnackedRecords() (native-or-cache) instead of getCachedUnackedRecords().
Finding 3: Java close/recreate contract is untested
Severity: Medium
File: rust/jni/src/arrow_stream/tests.rs, rust/jni/src/stream/tests.rs
Issue: Tests only cover close_native_stream leaving the Rust Option populated. They do not cover Java close() continuing after flush failure, throwing on cache failure, retrying close(), recreateStream after flush failure, or discardFailedRecreation.
Impact: The user-visible behavior (exception type, handle lifetime, recreate) can regress without failing these tests.
Suggested fix: Add Java tests for: flush-fail then cache then throw; cache-fail then retry close(); getUnacked* after failed close(); recreate* after flush-fail close(); recreate replay failure destroying only the new stream.
Finding 4: Tests assert close() returns Err, but Rust close is idempotent
Severity: Medium
File: arrow_stream/tests.rs, stream/tests.rs
Issue: Rust close() returns Ok(()) if is_closed is already true. Arrow waits for wait_for_offset to fail first; a terminal invalid_argument often closes the stream before close(). Then assert!(close_native_stream(...).is_err()) fails even when the stream was retained correctly. The proto test is better (close is the flush), but a 50ms delayed error can still close the stream first.
Impact: Flaky CI, or a false sense that "close failed" is required for recovery. Recovery should work whether close returned Ok or Err.
Suggested fix: Assert the stream is still Some and get_unacked_* has the record. Treat close's Result as either Ok (already closed) or Err (flush failed).
Finding 5: Cache failure is hidden behind the flush exception
Severity: Medium
File: BaseZerobusStream.close(), ZerobusArrowStream.close()
Issue: If both fail, the thrown exception is the flush error; cache failure is only addSuppressed. Comments say to retry recovery, but callers typically catch close and call recreateStream (Finding 1) or log and drop the stream (leak + data stuck in native with no finalizer).
Impact: Operators won't see that unacked data was not copied to Java. Dropping the object after a failed close leaks native memory and the only copy of the records.
Suggested fix: Prefer throwing the cache failure (flush as suppressed), or a dedicated "close completed, recovery incomplete, retry close()" error. Document that close() must be retried until nativeHandle == 0.
Finding 6: close() succeeding with empty unacked data is gone
Severity: Medium (intentional, but breaking)
File: BaseZerobusStream.close(), ZerobusArrowStream.close()
Issue: Previously, cache failure was swallowed, empty lists were stored, the handle was destroyed, and close() returned. Recreate then saw no data (silent loss). Now cache failure throws and skips destroy.
Impact: try-with-resources / close() can now throw in cases that used to return. Correct, but callers that assumed close() always released the native handle will leak if they don't retry.
Suggested fix: Changelog should say close() can now throw when unacked data cannot be copied, and the native handle stays alive until a later successful close().
```
1. Fixed in this PR
2. It is preexisting on a deprecated method, I wouldn't do it in this PR
3. We cannot really test it until we have a mock for java
4. This shouldn't be an issue now, Once we have a mock we can more easily inject failures and make the tests not timing dependent.
5. Cache failure cannot really happen, but we should add better tests later on.
6. It is internal state/handling of the SDK, there shouldnt be anything behaviorally different other than the bug fix.

elenagaljak-db (approved):
thanks for the fix

PR #736, opened 2026-08-14
[FFI] Complete Arrow shutdown before free returns

nikolaobradovic-db:

- Update zerobus_arrow_stream_free to wait until Arrow background work stops, request bodies reach EOF or are dropped, and retained batches are released before returning.
- Previously, tonic could retain an imported C Data batch after the stream handle was freed. Its release callback could then access producer state that had already been destroyed.
- The change adds request-body lifecycle tracking, supervisor reaper waiting, runtime-aware blocking shutdown, recurring warnings for stalled shutdown, and fail-stop handling when shutdown cannot complete safely. Native Rust Drop remains nonblocking; only C FFI destruction waits for complete shutdown.
- Go finalizers delegate blocking cleanup to a goroutine. Go and C++ documentation describes the updated destruction and concurrency behavior. No C ABI signatures change.

## How is this tested?

- Added deterministic C Data lifetime tests covering:
- release before free returns;
- transport-retained owners;
- release outside SDK mutexes;
- multi-thread Tokio runtime destruction.
- Added request-body EOF/fusion and supervisor-reaper regression tests.


rust/ffi/src/arrow.rs
```diff
@@ -291,13 +304,108 @@ pub extern "C" fn zerobus_sdk_create_arrow_stream_with_headers_provider(
...
+                    return;
+                }
+                match tokio::runtime::Handle::try_current() {
+                    Ok(handle)
+                        if handle.runtime_flavor()
+                            == tokio::runtime::RuntimeFlavor::MultiThread =>
+                    {
+                        tokio::task::block_in_place(|| abort_and_drop_arrow_stream(stream));
+                    }
+                    Ok(_) => abort_and_drop_arrow_stream_on_thread(stream),
```
elenagaljak-db, line 412: This blocking branch can deadlock for a direct FFI consumer calling free from a single-thread Tokio runtime when a C Data release callback waits for work on that same runtime. I reproduced this locally. Our wrappers call free, but they use IPC ingestion only, so they take the earlier nonblocking path. 
Just flagging this, myb its okay to leave it as is

nikolaobradovic-db: Comment is valid, thanks. I clarified the contract around invoking free from ST env.

elenagaljak-db (approved):

PR #738, opened 2026-08-14
[Rust] Fix Arrow failed-enqueue finalization

nikolaobradovic-db:

- Roll back Arrow logical offsets and record ranges when enqueueing fails with recovery disabled.
- Claim terminal admission before releasing the ingest lock, preventing concurrent close from masking the request-stream error.
- Await shared terminal finalization instead of racing a fixed local timeout.
- Serialize flush target snapshots with enqueue rollback so withdrawn offsets are never awaited.
- Add coverage for first-batch failure, acknowledged-prefix rollback, delayed finalization, and concurrent close.

## How is this tested?

- New concurrency regression test verified red without the admission claim and green with it.

### DIFF
--- a/rust/sdk/src/stream/arrow/mod.rs
+++ b/rust/sdk/src/stream/arrow/mod.rs
@@ -89,6 +89,7 @@ struct TestHooks {
     replay_send: TestBarrierGate,
     ack_applied: TestNotifyGate,
     ack_idle: TestNotifyGate,
+    failed_enqueue: TestBarrierGate,
     close_finalize: TestBarrierGate,
 }
 
@@ -172,7 +173,7 @@ pub struct ZerobusArrowStream {
     _last_ack_rx: watch::Receiver<Option<OffsetId>>,
     /// True once the stream is terminally closed and unacknowledged batches may be retrieved.
     is_closed: Arc<AtomicBool>,
-    /// Rejects new ingests as soon as terminal finalization owns admission.
+    /// Rejects new ingests once terminal finalization or a failed enqueue owns admission.
     admission_closed: Arc<AtomicBool>,
     /// Coordinates one resumable explicit-close request with the recovery supervisor.
     close: CloseCoordinator,
@@ -408,6 +409,8 @@ impl ZerobusArrowStream {
     /// # Errors
     ///
     /// * `StreamClosedError` - If the stream is closing or closed
+    /// * The terminal request-stream error if enqueueing fails with recovery disabled;
+    ///   the call waits for terminal finalization before returning that error
     /// * `InvalidArgument` - If the batch schema doesn't match the stream schema, or the
     ///   batch has zero rows (an empty batch carries no data to send or acknowledge)
     ///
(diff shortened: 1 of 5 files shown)


rust/sdk/src/stream/arrow/mod.rs
```diff
@@ -526,14 +529,30 @@ impl ZerobusArrowStream {
...
+                    let barrier = self.test_hooks.failed_enqueue.lock().await.take();
+                    if let Some(barrier) = barrier {
+                        barrier.reached.notify_one();
+                        barrier.proceed.notified().await;
+                    }
+                }
+                // Finalization must reacquire ingest_mutex before publishing its outcome.
+                drop(_guard);
+                self.request_send_failure.report();
+                return match self.wait_for_terminal_outcome().await {
```
teodordelibasic-db, line 550: Should we bound the wait with `flush_timeout_ms`?

```rust
return match timeout(
    Duration::from_millis(self.options.flush_timeout_ms),
    self.wait_for_terminal_outcome(),
)
.await
{
    Err(_) => Err(ZerobusError::StreamClosedError(
        tonic::Status::deadline_exceeded("Failed to send batch"),
    )),
    Ok(Err(error)) => Err(error),
    Ok(Ok(())) => Err(Self::terminal_error_or(&self.server_error_rx, || {
        "Failed to send batch".to_string()
    })),
};
```

`wait_for_offset_internal` is bounded by `flush_timeout_ms`, but this path loops until `CloseState::Finalized`. The empty-target [`flush()` branch](https://github.com/databricks/zerobus-sdk/blob/f99d283bf0c49ddc598967d4159aeb365fb09f47/rust/sdk/src/stream/arrow/mod.rs#L745) and the [`Open` plus `admission_closed` arm](https://github.com/databricks/zerobus-sdk/blob/f99d283bf0c49ddc598967d4159aeb365fb09f47/rust/sdk/src/stream/arrow/mod.rs#L881) of `close_internal` have the same unbounded wait. If the supervisor never reaches `finish()`, `ingest_batch` hangs a caller ingest loop.

The rustdoc says this call returns the request-stream error, but that is only true on `Err`. The `Ok(())` arm synthesizes `Status::internal("Failed to send batch")`, while the supervisor's send failure is `Status::unavailable` with `Flight request stream closed while sending`. That `Ok` arm looks unreachable after the `admission_closed` claim, but if we keep it, could we still surface whatever is on `server_error_rx` instead of a generic internal?

nikolaobradovic-db: We should not use flush_timeout_ms here because this is terminal lifecycle coordination, not an ACK/flush wait. Returning on timeout could expose an incompletely finalized stream, make get_unacked_batches() temporarily invalid, and hide the real request-stream error. The wait is cancellation-safe, and finalization continues independently through the supervisor/reaper.
It is possible to do it by adding yet another complex state machine, by aborting the supervisor and awaiting the reaper, which I would rather skip at the cost of having a hang in case the supervisor deadlocks(for now).
I will just make a small change to preserve the terminal error.

teodordelibasic-db (approved):

PR #741, opened 2026-08-16
[Rust][Python] Add first-class federated-token auth (external IdP / Entra ID)

### LINKED ISSUE #740: [Rust][Python] Add first-class external-IdP (Entra ID) token federation to the SDK
## Summary

Would like to add a first-class authentication mode that lets a client
authenticate a Zerobus stream with an external identity provider (for example
Microsoft Entra ID) instead of a Databricks OAuth client_id/client_secret. The
SDK exchanges the external IdP token for a Zerobus-scoped Databricks token using
the RFC 8693 token-exchange grant, then caches and refreshes it, the same way
the existing OAuth path does.

## Motivation

Several enterprise customers cannot use Databricks-managed OAuth secrets for
their security posture, so they cannot use Zerobus today. The two shapes they
ask for are:

1. [**Account-level federation:**](https://docs.databricks.com/aws/en/dev-tools/auth/oauth-federation-policy#account]) no Databricks-managed service principal. The
   identity is synced into Databricks via Automatic Identity Management (SCIM).
   This is the mode that satisfies "identity lives only in the external IdP."
2. [**Workload identity federation:**](https://docs.databricks.com/aws/en/dev-tools/auth/oauth-federation-policy#workload) a Databricks service principal with a
   client_id and no secret, with a federation policy attached.

The token exchange already works on the gRPC path today via the undocumented
`HeadersProvider` hook. This proposal makes it a first-class, opt-in auth mode
so callers do not have to hand-roll the exchange. The Zerobus service does not
change: it still validates a bearer token.

## Design

The whole design is client-side, in the SDK. The Zerobus service is untouched.

**Token exchange (shared request shaping).** The existing client-credentials
grant already builds a Zerobus-scoped request against the workspace
`/oidc/v1/token` endpoint: `scope=all-apis`, `resource=api://databricks/
workspaces/<id>/zerobusDirectWriteApi`, and table-scoped `authorization_details`
(the Unity Catalog privileges for the target table). The new token-exchange
grant reuses that exact request shaping and only differs in the grant-specific
parameters: `grant_type=urn:ietf:params:oauth:grant-type:token-exchange`, the
`subject_token` (the external IdP token), `subject_token_type`, and, for
workload identity federation, the service principal `client_id`. Keeping one
request-shaping path keeps the two grants at parity.

**New auth provider.** A new provider (implementing the existing
`HeadersProvider` trait, including its `invalidate()` hook) takes an async
callback that returns the current external IdP token. It performs the exchange,
caches the exchanged Databricks token in the existing per-SDK token cache, and
serves it until it nears expiry, then refreshes. The callback is invoked only
when a fresh token must be minted (a cache miss or a proactive refresh), not on
every request.

**One toggle for both modes.** An optional `client_id` selects the mode:
present for workload identity federation (the exchange names the service
principal), absent for account-level federation (the subject resolves to the
SCIM-synced identity). The cache keys on `(client_id-or-none, table)` so the two
modes, and distinct service principals, cache independently.

**Callback, not a bare token.** The provider takes a callback rather than a
pre-obtained token on purpose. An external IdP token is short-lived (about an
hour), so a bare token would strand the stream at the token's expiry with no way
to refresh. A caller that holds a static token can wrap it in a trivial
callback.

**Opt-in surface.** New builder methods select the federated mode; the existing
client-credentials and custom-headers paths are unchanged. Exposed through the
Python binding first for validation (`auth=FederatedToken(...)` on
`create_stream`, sync and async callbacks); other language bindings follow.

## REST insert path

The same exchanged token works on the REST insert endpoint
(`/zerobus/v1/tables/<table>/insert`), verified with an HTTP 200 insert. The
"Invalid token audience" error some users have seen is caused by minting the
token without the `zerobusDirectWriteApi` resource, not a server-side gap. This
is a note for callers hitting REST directly; the SDK itself is gRPC-only.

## Backward compatibility

Purely additive and opt-in. The `client_id`/`client_secret` and custom
headers-provider paths behave identically; the new behavior is reached only when
the caller selects the federated mode.

## Testing evidence (from the reference implementation)

- Unit tests for request-shaping parity with and without `client_id`, plus
  provider tests for caching, refresh, `invalidate()`, mode independence, and
  callback-error propagation.
- Live end-to-end validation of both modes on gRPC against a real workspace,
  including a run over a full token lifetime showing the exchanged token
  auto-refreshes before expiry with no interruption.

## Scope and phasing

This is the client-side first step. A server-side / control-plane exchange (so
REST and non-SDK clients benefit uniformly) is a reasonable later phase; it is a
larger cross-team change and is not part of this proposal.

## Related

- A detailed internal design doc is available to Databricks reviewers on request.

I have a working prototype and a reference implementation ready, and I am happy
to open the PR against this issue. Would appreciate agreement on the API surface
before review.

anilmenon14:
## Motivation

Enterprise customers who cannot use Databricks-managed OAuth secrets currently
cannot use Zerobus. This PR adds an opt-in authentication mode that federates an
external identity provider (for example Entra ID) token into a Zerobus-scoped
Databricks token, **client-side**, so those customers can stream without a
Databricks secret. The platform token exchange already works on gRPC via the
undocumented `HeadersProvider` hook; this makes it first-class instead of a
workaround. The Zerobus service is unchanged.

## What this changes

- **Rust core:** new `FederatedTokenProvider` (implements the existing
  `HeadersProvider` trait, including `invalidate()`) and an `IdpTokenSupplier`
  callback type. The client-credentials and token-exchange grants now share one
  request-shaping path in `default_token_factory.rs`, keeping them at parity.
  Opt-in `StreamBuilder::federated(...)` and `federated_with_client_id(...)`.
- **Python binding:** `auth=FederatedToken(idp_token_supplier=..., databricks_client_id=...)`
  on `create_stream` (sync and async), with the Python callback bridged across
  FFI (sync and async callbacks both supported). The `HeadersProvider.invalidate()`
  hook is now forwarded through the Python bridge.
- No existing signatures change behavior.

Resolves #740 

## The two supported modes

- [**Account-level federation:**](https://docs.databricks.com/aws/en/dev-tools/auth/oauth-federation-policy#account]) (`databricks_client_id` omitted): no Databricks
  service principal. The exchanged token's subject resolves to an identity
  synced into Databricks via Automatic Identity Management (SCIM). The exchange
  request omits `client_id`.
- [**Workload identity federation:**](https://docs.databricks.com/aws/en/dev-tools/auth/oauth-federation-policy#workload) (`databricks_client_id` set): a Databricks
  service principal with a client_id and no secret, with a federation policy
  attached. The exchange request names the service principal via `client_id`.

## Backward compatibility

The `client_id`/`client_secret` (OAuth) and `headers_provider` paths are
unchanged. The new behavior is reached only when the caller passes the new,
opt-in `auth=FederatedToken(...)` argument (or the `federated*` builder methods
in Rust).

## Testing evidence

- Rust: unit tests for request-shaping parity with/without `client_id`, plus
  end-to-end provider tests against a mock token endpoint (caching, `invalidate()`
  re-mint, mode independence, supplier-error propagation). All lib tests pass;
  `clippy` and `rustfmt` clean.
- Python: dispatch/routing and argument-validation tests. `black`, `isort`,
  `pycodestyle` clean.
- Live (both modes, against a real workspace, streaming to a UC table): a
  successful stream on gRPC for account-level and workload identity federation;
  caching confirmed (the IdP callback fires once across multiple streams from one
  SDK); a >1 hour soak showing the exchanged token auto-refreshes near the
  ~55-minute mark (token lifetime ~60 min minus the 300s cache buffer) with no
  interruption; and an async callback validated via the async SDK.

## REST insert path (resolved, not a limitation)

The same federated exchange token works on the REST insert endpoint
(`/zerobus/v1/tables/<table>/insert`), verified live with an HTTP 200 insert.

## Known limitations / follow-ups

- Other language bindings (TypeScript, Java, Go, C++) to follow.
- A server-side / control-plane exchange remains a reasonable later phase (it
  would cover non-SDK and REST callers uniformly); this PR is the client-side
  first step.
- Suggested maintainer test follow-up: the sync callback is unit-tested and
  live-verified; the async callback bridge and the Python FFI error mapping are
  verified live but not yet covered by automated unit tests, because both need a
  live token+gRPC endpoint or a fuller mock harness. Adding CI coverage for these
  two using the repo's test fixtures would be worthwhile.

## Housekeeping

- DCO sign-off on all commits; GPG-signed.
- `NEXT_CHANGELOG.md` updated (Rust core and Python).
- README and an `examples/` sample added for the new API.

### DIFF
--- a/python/NEXT_CHANGELOG.md
+++ b/python/NEXT_CHANGELOG.md
@@ -6,6 +6,18 @@
 
 ### New Features and Improvements
 
+- Added `FederatedToken` for external-IdP (for example Entra ID) authentication.
+  Pass `auth=FederatedToken(idp_token_supplier=..., databricks_client_id=...)`
+  to `create_stream` and the SDK exchanges the external IdP token for a
+  Zerobus-scoped Databricks token (RFC 8693 token exchange), caching and
+  refreshing it. Supports account-level federation (omit `databricks_client_id`,
+  identity synced via Automatic Identity Management) and workload identity
+  federation (set `databricks_client_id` to the service principal, no secret).
+  The `idp_token_supplier` callback may be synchronous or asynchronous. Existing
+  `client_id`/`client_secret` and `headers_provider` calls are unchanged.
+- Forwarded the `HeadersProvider.invalidate()` hook through the Python bridge, so
+  a custom provider can drop cached auth state when the server rejects a token.
+
 ### Bug Fixes
 
 ### Documentation
--- a/python/README.md
+++ b/python/README.md
@@ -287,6 +287,59 @@ asyncio.run(main())
 
 See the [`examples/`](examples/) directory for complete runnable examples.
 
+## Authentication
+
+`create_stream()` supports three authentication methods, in this precedence:
+`auth` (federation) > `headers_provider` > `client_id`/`client_secret`.
+
+### OAuth client credentials (default)
+
+```python
+stream = sdk.create_stream(
+    client_id="your-client-id",
+    client_secret="your-client-secret",
+    table_properties=table_properties,
+)
+```
+
+### External-IdP federation (e.g. Entra ID)
+
+Use `auth=FederatedToken(...)` to authenticate with an external identity
+provider instead of a Databricks OAuth secret. You provide a callback that
+returns the current external IdP token; the SDK exchanges it for a
+Zerobus-scoped Databricks token (RFC 8693 token exchange) and caches and
+refreshes that token for you. The callback may be synchronous (sync SDK) or
+asynchronous (async SDK).
+
+```python
+from zerobus import FederatedToken
+
+def get_idp_token():
+    # Return the current external IdP (e.g. Entra ID) access token.
+    ...
+
+# Account-level federation: no Databricks service principal. The identity is
+# synced into Databricks via Automatic Identity Management (SCIM).
+stream = sdk.create_stream(
+    table_properties=table_properties,
+    auth=FederatedToken(idp_token_supplier=get_idp_token),
+)
+
+# Workload identity federation: a Databricks service principal with a client_id
+# and no secret, with a federation policy attached.
+stream = sdk.create_stream(
+    table_properties=table_properties,
+    auth=FederatedToken(idp_token_supplier=get_idp_token, databricks_client_id="<sp-client-id>"),
+)
+```
+
+See [`examples/sync_example_federated.py`](examples/sync_example_federated.py) for a complete example.
+
+### Custom headers provider
+
+For advanced cases you can supply your own `HeadersProvider` via
+`headers_provider=`; see the [`HeadersProvider`](#headersprovider) reference.
+
 ## Configuration
 
 Configure stream behavior by passing a `StreamConfigurationOptions` object to `create_stream()`:
--- a/python/examples/sync_example_federated.py
+++ b/python/examples/sync_example_federated.py
@@ -1,0 +1,119 @@
+"""
+Synchronous Ingestion Example - Federated Authentication (external IdP / Entra ID)
+
+This example demonstrates streaming to Zerobus using an external identity
+provider (for example Microsoft Entra ID) instead of a Databricks OAuth
+client_id/client_secret. You supply a callback that returns the current external
+IdP token; the SDK exchanges it for a Zerobus-scoped Databricks token (RFC 8693
+token exchange), then caches and refreshes that token for you.
+
+Two federation modes, selected by `databricks_client_id`:
+  - Account-level federation (databricks_client_id omitted): no Databricks
+    service principal. The identity is synced into Databricks via Automatic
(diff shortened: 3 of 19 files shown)


bogdanb-db (approved):
auth workflow is LGTM! Wait for an approval from @teodordelibasic-db  or @elenagaljak-db

anilmenon14:
Thanks @bogdanb-db . 
@teodordelibasic-db , @elenagaljak-db , since I have done this PR from an external fork, the CI checks may not run right as described in this open PR #458  that aims to fix CI checks for external fork. In that case, since I have a Databricks ID, I could ask internally for access to a feature branch on the same repo to push to

anilmenon14:
Superseded by #760 reopened against main from an internal branch (Same two GPG-signed commits, same content) now that I have write access (external-fork CI was blocked, as in #458). Closing this one.

PR #742, opened 2026-08-17
[PureGo] Add ColumnsFromDescriptor

zlata-stefanovic-db:

Adds `ColumnsFromDescriptor` to the pure-Go public API. It returns the set of
*column names* a serialized protobuf descriptor declares — the bytes returned by
`FetchProtoDescriptorFromUC` or passed to `WithProto`:

```go
columns, err := zerobus.ColumnsFromDescriptor(descriptor)
if err != nil {
    log.Fatal(err)
}
if _, ok := columns["event_id"]; !ok {
    log.Fatal("table has no event_id column")
}
```

**Why:** a caller who fetches a descriptor from Unity Catalog currently has no
way to see which columns it declares without unmarshalling
`descriptorpb.DescriptorProto` themselves — importing protobuf internals to
answer a question the SDK already holds the data for.

Only top-level fields are columns; the fields of a nested message belong to that
column's value. The exported function is a thin wrapper over a private helper,
mirroring `FetchProtoDescriptorFromUC`, so failures surface as `*zerobus.Error`
with `Op: "ColumnsFromDescriptor"` and non-retryable — consistent with every
other public entry point.

Two documentation fixes ride along:

- The README claimed "Unknown JSON fields are ignored." They are rejected:
  `internal/dynamicproto` runs `protojson` with `DiscardUnknown` false, and
  `TestConverter_UnknownFieldRejected` has been asserting that rejection all
  along. Anyone reading that line would have planned around the wrong behavior.
- The package doc comment (54 lines) and `FetchProtoDescriptorFromUC`'s doc
  comment (19 lines) were long yet partial — the package doc documented four
  topics out of the README's eleven, and every section it did cover was already
  documented by symbol godoc on the same page (`WithWaitForReady` covers async
  creation in more detail, `HeadersProvider` covers auth, and the Quick start
  block was `ExampleStream_ingestThenFlush` verbatim). Both are now short and
  point at the README, which keeps the queue-then-`Flush` rule up front.

## How is this tested?

New `purego/zerobus/columns_test.go`, in the external `zerobus_test` package so
it exercises the public surface:

- Round-trip through the SDK's own `schema.DescriptorFromUCColumns` builder, so
  the result is checked against the descriptor the SDK itself produces from a UC
  schema.
- Nested-message and unnamed fields are excluded.
- Error paths (truncated bytes, empty bytes, a descriptor with no fields) return
  `*zerobus.Error` with the right `Op` and are classified non-retryable.

`ExampleColumnsFromDescriptor` was added to `example_test.go` and compiles as
part of the suite. Full purego suite passes under `-race` (553 tests), with
`go vet` and `gofmt` clean.

### DIFF
--- a/purego/zerobus/sdk.go
+++ b/purego/zerobus/sdk.go
@@ -1,5 +1,4 @@
-// Package zerobus is a pure-Go client for Zerobus ingestion. It speaks gRPC
-// directly, so it needs no cgo and no prebuilt native libraries.
+// Package zerobus is a pure-Go client for Zerobus ingestion.
 //
 // Create an SDK with New and a stream with CreateStream. Ingestion is
 // asynchronous: IngestRecordOffset returns as soon as the record is queued, so
--- a/purego/zerobus/sdk.go
+++ b/purego/zerobus/sdk.go
@@ -209,12 +209,8 @@ func (s *SDK) CreateStreamWithProvider(
 }
 
 // FetchProtoDescriptorFromUC returns a protobuf descriptor built from a Unity
-// Catalog table schema.
-//
-// Every call performs a fresh Unity Catalog request; the SDK does not cache the
-// result. Fetch once per table and reuse the bytes for every stream on it —
-// calling this per stream, or worse per record, puts a UC round-trip on the
-// ingestion path. Fetch again only when the table schema changes.
+// Catalog table schema. Every call performs a fresh Unity Catalog request; the
+// SDK does not cache the result.
 func (s *SDK) FetchProtoDescriptorFromUC(
 	ctx context.Context,
 	tableName, clientID, clientSecret string,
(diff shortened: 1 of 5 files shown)


purego/zerobus/sdk.go
```diff
@@ -254,21 +212,9 @@ func (s *SDK) CreateStreamWithProvider(
...
-//	    stream, err := sdk.CreateStream(ctx, table, clientID, clientSecret,
-//	        zerobus.WithProto(descriptor))
-//	    // ...
-//	}
-//
-// Calling it per stream, or worse per record, adds a Unity Catalog round-trip
-// to the hot path. Fetch again only when the table schema changes.
+// result. Fetch once per table and reuse the bytes for every stream on it —
+// calling this per stream, or worse per record, puts a UC round-trip on the
+// ingestion path. Fetch again only when the table schema changes.
```
elenagaljak-db: nit: maybe this guideline should just be in the readme, and the comment should remain as explanation of the method?

zlata-stefanovic-db: Removed it from here, left the explanation in the readme

elenagaljak-db (approved):

PR #758, opened 2026-08-18
[PureGo] Add Arrow IPC payload

zlata-stefanovic-db:

Adds `internal/arrowproto` with the Arrow IPC payload type and the typed encoding
path that produces it. This is the **first of five changes split out of #732**,
which carried the whole encoder at +3386 lines. Nothing in the package is
reachable from a public API yet.

A payload has to survive a reconnect, and a caller's `RecordBatch` cannot: it
points at buffers the caller still owns and may release the moment `ingest`
returns. So a payload materializes when it is built, as a canonical
self-contained Arrow IPC stream carrying schema, dictionaries, exactly one
non-empty batch, and the end marker. The core can then hold it across a
reconnect without pinning anything of the user's.

The two pieces worth review attention:

- **Schema ownership goes through IPC.** `New` serializes the caller's schema and
  reads it back rather than copying fields, because a shallow field copy still
  aliases mutable pointer-backed types such as `DictionaryType` — a caller
  mutating one after construction would change what the stream encodes.
- **`Slice` is a genuine row-range slice.** A partially acknowledged batch
  replays only its unacknowledged suffix rather than duplicating an acknowledged
  prefix, which is the same client-slicing choice the Rust SDK made, so the core
  does not depend on the server deduplicating a replayed prefix. This became
  expressible only because #682 generalized the core from "one offset = one
  atomic ack" to protocol-neutral durability units. `Slice` reconciles the
  payload's row count against the decoded batch before slicing, since a bound
  past the decoded extent panics inside Arrow and that panic would escape
  recovery into the caller's goroutine.

**Admission sizes a batch from the rows it covers, not the whole buffers it
points at.** A slice shares its parent's buffers, so measuring buffer lengths
charges a ten-row slice of a 20,000-row batch for the entire parent. That is not
merely conservative: the core reserves the estimate *before* encoding, and
`reserve` rejects any single weight above the buffered-bytes limit, so an
over-estimate is refused outright rather than corrected by the reconciliation
that follows materialization.

Slicing also rebases only the top-level node — a struct's fields stay parallel to
the parent's offset, and a list's values are reached through its offsets buffer —
so the children have to be charged from the parent's row window as well, and they
are the dominant term for both layouts. A row window is therefore threaded
through the whole sizing recursion rather than each node reporting its own
length. Fixed-width, variable-width, list, large-list, map, fixed-size-list,
struct, and dictionary layouts are all charged for the rows they cover;
list-view, union, and run-end-encoded still fall back to whole buffers.

A batch's custom metadata is charged separately, because arrow-go writes it into
the IPC message header rather than into any column buffer, so a walk over the
columns cannot see it at all. The 64 KiB slop hides small metadata, but a batch
carrying more than that would otherwise be admitted for less than it
materializes — and since `reconcileReservation` grows a short reservation only
out of already-free budget and never waits, that surfaces as a spurious
`ErrPayloadTooLarge` rather than as backpressure.

**Dependencies.** Adds `github.com/apache/arrow-go/v18`, the first dependency
this module has taken beyond grpc and protobuf. arrow-go requires
`google.golang.org/grpc` v1.82.0, which raises this module's grpc minimum from
v1.81.1.

**Deliberately not here**, each following as its own change:

1. Raw IPC bytes as an input, with the flatbuffer preflight that sums declared
   uncompressed buffer sizes so a compression bomb is rejected at admission
   instead of expanding past the limit once Arrow materializes it.
2. Flight frame encoding: the 2 MiB chunk plan measured by binary searching
   actual encoded protobuf size, and the frame emitter.
3. Flight acknowledgment and batch metadata parsing in `internal/transport`.
4. `stream.EncoderHooks` and the stream-core wiring.

`EncodeSchemaIPC`/`DecodeSchemaIPC` are exported in #732 but consumed only by
tests there, so this change keeps an unexported `encodeSchemaIPC` for the
admission base and they get exported by the change that consumes them.

## How is this tested?

10 unit tests in `internal/arrowproto` (13 cases counting subtests), none of
which need a server or a network — every test runs in-process against Arrow. `go
test -race ./...` passes across the module (566 tests, 10 packages), as does
`CGO_ENABLED=0 go test ./...`, and `gofmt`, `go vet`, and `go mod tidy -diff` are
clean.

Coverage worth calling out:

- The typed round trip runs on `memory.NewCheckedAllocator` and asserts zero
  outstanding allocations after the caller's batch is released, so the payload
  demonstrably holds no reference to the batch, its columns, or their buffers.
  It also asserts `IPCBytes` hands back a copy rather than the retained slice.
- Schema exactness includes metadata, so a batch matching field-for-field but
  differing in schema metadata is rejected rather than silently encoded.
- Mutating a `DictionaryType` after `New` does not change the owned schema.
- `Slice` produces the correct suffix, leaves the original payload unchanged, and
  rejects both a fully acknowledged prefix and a zero prefix. The core only ever
  slices a partially acknowledged payload, so either bound means its accounting
  is wrong and should surface rather than quietly yield an empty or unchanged
  payload. A payload whose row count drifted from its bytes returns an error
  instead of panicking.
- LZ4 and Zstd both round-trip 2,000 rows and each come out smaller than
  uncompressed; an unsupported compression value is rejected at construction.
- Admission covers the materialized payload in every case tested, and a ten-row
  slice of a 20,000-row batch is charged under a hundredth of its parent — for
  string, list, and struct columns, the last two being the layouts whose children
  a per-node measurement charges in full. Head and tail slices cost the same, so
  the estimate does not depend on where the slice sits.
- A separate case pins that a nested child's window follows its parent's offset,
  using a struct whose final rows carry 256 KiB strings: sizing the wrong rows
  under-reserves there rather than merely looking plausible on uniform data.
- A one-row batch carrying 512 KiB of custom metadata has to raise the estimate
  by at least the metadata it carries, and the payload it materializes has to
  stay within that estimate. Asserting only the second half would pass on the
  slop alone, so the growth is asserted too.

Not covered here because it needs the changes that follow: Flight frame
chunking, compression-bomb rejection at admission, and anything end-to-end
against a Flight server.

## Known Rust/PureGo differences (non-blocking for this PR)

These are current implementation differences, not regressions in this change:

- **Backpressure model**
  - PureGo Arrow payload admission is byte-based (size estimate + reservation).
  - Rust Arrow stream admission is count-based (`max_inflight_batches` semaphore).
  - Consequence: under very large or highly compressible inputs, the two SDKs may
    apply backpressure at different points.

- **Compressed IPC expansion guard**
  - PureGo preflights IPC metadata and charges declared uncompressed buffer sizes
    before materialization.
  - Rust currently materializes IPC directly in `ingest_ipc_batch`.
  - Consequence: PureGo may reject some compressed payloads earlier than Rust.

- **Dictionary wire policy (future transport behavior)**
  - Rust Flight encoding defaults to dictionary hydration.
  - PureGo payload contract preserves canonical IPC with schema/dictionary/batch.
  - Consequence: once Flight framing is compared end-to-end, dictionary handling
    may differ unless explicitly aligned in transport-layer follow-ups.


elenagaljak-db (approved):
Lgtm

PR #761, opened 2026-08-20
[Rust] Add persistent stream proto

elenagaljak-db:
## What changes are proposed in this pull request?

This PR adds the in-development protobuf contract used by persistent Zerobus ingestion streams. It contains only the wire-level API.

The contract adds:

- `PersistentStream`, a bidirectional RPC whose first message creates a durable stream or resumes one by `stream_id`.
- `ResumeIngestStreamRequest`, including the protobuf descriptor and required `record_type` used to validate a resumed stream against its original configuration.
- `ResumeIngestStreamResponse.last_committed_offset`, the server watermark from which the client continues at `last_committed_offset + 1`.
- Persistent envelopes for create, resume, ingestion, acknowledgements, and server close signals.
- The server-side `RetireStream` contract; a public Rust retirement API is intentionally deferred.

The shapes and field numbers match the current Shinkansen server contract. The PR also records this API-contract addition in `rust/NEXT_CHANGELOG.md`.

### Stack

1. **#761 — protobuf contract**
2. #762 — gRPC transport and recovery engine
3. #763 — public Rust API
4. #764 — stateful mock and integration tests
5. #765 — documentation and runnable example

## How is this tested?

The protobuf is compiled by the Rust SDK build. The complete create, ingest, flush, and resume flow is exercised in #764.

PR #762, opened 2026-08-20
[Rust] Add persistent gRPC transport

elenagaljak-db:

This PR implements the internal gRPC transport and recovery machinery for persistent streams. It deliberately does not expose a new public SDK type; that surface is added in #763.

The sender, receiver, connection, and supervisor tasks are shared between ephemeral and persistent streams through a typed transport seam. Persistent streams use the dedicated RPC while ephemeral behavior remains unchanged.

Persistent behavior includes:

- Create once, then resume the same `stream_id` after recovery.
- Stable logical offsets on the persistent wire protocol.
- New ingestion beginning at `last_committed_offset + 1` after process-level resume.
- Landing-zone reconciliation for the lost-ack race where the server committed records before the client disconnected.
- Local completion of reconciled waiters and callbacks, with only offsets above the watermark resent.
- Existing backpressure, flush, close, callback, credential-refresh, and initial-retry behavior shared across transports.

The implementation remains behind `eos`. Its internal transport/recovery addition is recorded in `rust/NEXT_CHANGELOG.md`.

### Stack

1. #761 — protobuf contract
2. **#762 — gRPC transport and recovery engine**
3. #763 — public Rust API
4. #764 — stateful mock and integration tests
5. #765 — documentation and runnable example

## How is this tested?

- `cargo check -p databricks-zerobus-ingest-sdk --features eos`
- `cargo test -p databricks-zerobus-ingest-sdk --features eos --lib` — 156 tests passed
- End-to-end persistent coverage is added in #764.

### DIFF
--- a/rust/sdk/Cargo.toml
+++ b/rust/sdk/Cargo.toml
@@ -77,7 +77,7 @@ internal-arrow-c-data = [
 ]
 # Zero-copy protobuf parser.
 zeroparser = ["dep:self_cell", "dep:prost-build"]
-# Persistent (Eos) streams; in development, API is unstable
+# Persistent streams with exactly-once semantics (EoS); API is in development
 eos = []
 testing = ["dep:futures"]
 # Test-only deterministic seams (barriers/notifies) for the Arrow stream. Zero footprint
--- a/rust/sdk/Cargo.toml
+++ b/rust/sdk/Cargo.toml
@@ -77,7 +77,7 @@ internal-arrow-c-data = [
 ]
 # Zero-copy protobuf parser.
 zeroparser = ["dep:self_cell", "dep:prost-build"]
-# Persistent streams with exactly-once semantics (EoS); API is in development
+# Persistent streams with exactly-once semantics (EoS); API is in development.
 eos = []
 testing = ["dep:futures"]
 # Test-only deterministic seams (barriers/notifies) for the Arrow stream. Zero footprint
--- a/rust/sdk/Cargo.toml
+++ b/rust/sdk/Cargo.toml
@@ -77,6 +77,8 @@ internal-arrow-c-data = [
 ]
 # Zero-copy protobuf parser.
 zeroparser = ["dep:self_cell", "dep:prost-build"]
+# Persistent (Eos) streams; in development, API is unstable
+eos = []
 testing = ["dep:futures"]
 # Test-only deterministic seams (barriers/notifies) for the Arrow stream. Zero footprint
 # unless enabled; never enabled by FFI/production builds.
--- a/rust/sdk/src/landing_zone.rs
+++ b/rust/sdk/src/landing_zone.rs
@@ -85,25 +85,22 @@ impl<T: Clone> LandingZone<T> {
         all_items
     }
 
-    /// Removes and returns items from the front of the unobserved queue while
-    /// `pred` holds, stopping at the first item that fails it.
+    /// Removes a prefix of observed items while `should_remove` returns true.
     ///
-    /// Used by the persistent-stream resume path to reconcile the retained tail
-    /// against the server's committed watermark: it removes the prefix of
-    /// records the server has already durably stored (offset <= resume watermark)
-    /// before re-sending the rest. Only the unobserved queue is consulted:
-    /// callers reset observation (`reset_observe`) first, so every retained item
-    /// lives in the queue in offset order. One semaphore permit is released per
-    /// removed item, mirroring `remove_observed`.
-    pub fn remove_front_while<F: FnMut(&T) -> bool>(&self, mut pred: F) -> Vec<T> {
+    /// Observed items stay in FIFO order. Stops at the first item that should
+    /// remain. Releases the corresponding backpressure permits.
+    pub fn remove_observed_prefix(&self, mut should_remove: impl FnMut(&T) -> bool) -> Vec<T> {
         let mut state = self.state.lock().expect("Lock poisoned");
         let mut permits = self.permits.lock().expect("Lock poisoned");
         let mut removed = Vec::new();
-        while let Some(front) = state.queue.front() {
-            if !pred(front) {
+        while let Some(front) = state.observed_items.front() {
+            if !should_remove(front) {
                 break;
             }
-            let item = state.queue.pop_front().expect("front just checked");
+            let item = state
+                .observed_items
+                .pop_front()
+                .expect("front existed before pop");
             permits.pop_front();
             removed.push(item);
         }
@@ -305,6 +302,34 @@ mod tests {
             result,
             Err(LandingZoneError::RemovingNonObservedElement)
         ));
+    }
+
+    #[tokio::test]
+    async fn test_remove_observed_prefix_preserves_fifo_suffix() {
+        let lz = LandingZone::new(4);
+        for value in 1..=4 {
+            lz.add(value).await;
+        }
+        for _ in 0..3 {
+            lz.observe().await;
+        }
+
+        assert_eq!(lz.remove_observed_prefix(|value| *value <= 2), vec![1, 2]);
+        assert_eq!(lz.reset_observe(), 1);
+        assert_eq!(lz.observe().await, 3);
+        assert_eq!(lz.observe().await, 4);
+    }
+
+    #[tokio::test]
+    async fn test_remove_observed_prefix_never_removes_unsent_items() {
+        let lz = LandingZone::new(2);
+        lz.add(1).await;
+        lz.add(2).await;
+        lz.observe().await;
+
+        assert_eq!(lz.remove_observed_prefix(|_| true), vec![1]);
+        assert_eq!(lz.len(), 1);
+        assert_eq!(lz.observe().await, 2);
     }
 
     #[tokio::test]
--- a/rust/sdk/src/landing_zone.rs
+++ b/rust/sdk/src/landing_zone.rs
@@ -85,6 +85,31 @@ impl<T: Clone> LandingZone<T> {
         all_items
     }
 
+    /// Removes and returns items from the front of the unobserved queue while
+    /// `pred` holds, stopping at the first item that fails it.
+    ///
+    /// Used by the persistent-stream resume path to reconcile the retained tail
+    /// against the server's committed watermark: it removes the prefix of
+    /// records the server has already durably stored (offset ≤ resume watermark)
+    /// before re-sending the rest. Only the unobserved queue is consulted:
+    /// callers reset observation (`reset_observe`) first, so every retained item
+    /// lives in the queue in offset order. One semaphore permit is released per
+    /// removed item, mirroring `remove_observed`.
+    pub fn remove_front_while<F: FnMut(&T) -> bool>(&self, mut pred: F) -> Vec<T> {
+        let mut state = self.state.lock().expect("Lock poisoned");
+        let mut permits = self.permits.lock().expect("Lock poisoned");
+        let mut removed = Vec::new();
+        while let Some(front) = state.queue.front() {
+            if !pred(front) {
+                break;
+            }
+            let item = state.queue.pop_front().expect("front just checked");
+            permits.pop_front();
+            removed.push(item);
+        }
+        removed
+    }
+
     /// Adds an item to the queue.
     ///
     /// This method will block if the maximum number of inflight requests has been reached,
--- a/rust/sdk/src/landing_zone.rs
+++ b/rust/sdk/src/landing_zone.rs
@@ -90,7 +90,7 @@ impl<T: Clone> LandingZone<T> {
     ///
     /// Used by the persistent-stream resume path to reconcile the retained tail
     /// against the server's committed watermark: it removes the prefix of
-    /// records the server has already durably stored (offset ≤ resume watermark)
+    /// records the server has already durably stored (offset <= resume watermark)
     /// before re-sending the rest. Only the unobserved queue is consulted:
     /// callers reset observation (`reset_observe`) first, so every retained item
     /// lives in the queue in offset order. One semaphore permit is released per
--- a/rust/sdk/src/stream/grpc/connection.rs
+++ b/rust/sdk/src/stream/grpc/connection.rs
@@ -3,7 +3,7 @@
 //! This module is transport-specific: it opens the bidirectional gRPC stream
 //! used by `ZerobusStream`. It handles both stream kinds through the transport
 //! seam (`super::transport`): ephemeral streams over the `EphemeralStream` RPC
-//! and, behind the `eos` feature, persistent streams over `PersistentStream`.
+//! and persistent streams over `PersistentStream`.
 //! The Arrow Flight transport has its own equivalent under `stream/arrow/`.
 
 use std::sync::Arc;
@@ -15,9 +15,11 @@ use tonic::transport::Channel;
 use tracing::{debug, error, info, instrument, warn};
 
 use super::supervisor::StreamInitInfo;
-use super::transport::{self, GrpcConnectionMode, InboundStream, Opened, OutboundSink};
+use super::transport::{
+    self, GrpcConnectionMode, InboundStream, Opened, OutboundSink, StreamOpenParams,
+};
 use crate::databricks::zerobus::zerobus_client::ZerobusClient;
-use crate::databricks::zerobus::{CreateIngestStreamRequest, RecordType};
+use crate::databricks::zerobus::RecordType;
 use crate::{HeadersProvider, TableProperties, ZerobusError, ZerobusResult};
 
 /// A freshly opened stream connection: the outbound sink, the inbound response
@@ -119,10 +121,8 @@ impl super::ZerobusStream {
         record_type: RecordType,
         kind: &GrpcConnectionMode,
     ) -> ZerobusResult<StreamConnection> {
-        let (sink, request_body) = transport::make_outbound(kind);
-        let mut request = tonic::Request::new(request_body);
-
-        let stream_metadata = request.metadata_mut();
+        let outbound = transport::make_outbound(kind);
+        let mut stream_metadata = tonic::metadata::MetadataMap::new();
         let headers = headers_provider.get_headers().await?;
         for (key, value) in headers {
             match key {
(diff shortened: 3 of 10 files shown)


rust/sdk/Cargo.toml
```diff
@@ -75,6 +75,8 @@ internal-arrow-c-data = [
 ]
 # Zero-copy protobuf parser.
 zeroparser = ["dep:self_cell", "dep:prost-build"]
+# Persistent (Eos) streams; in development, API is unstable
```
teodordelibasic-db: nit: `(exactly-once semantics / EoS)` just to be clearer maybe.

rust/sdk/src/landing_zone.rs
```diff
@@ -85,6 +85,31 @@ impl<T: Clone> LandingZone<T> {
         all_items
     }
 
+    /// Removes and returns items from the front of the unobserved queue while
+    /// `pred` holds, stopping at the first item that fails it.
+    ///
+    /// Used by the persistent-stream resume path to reconcile the retained tail
+    /// against the server's committed watermark: it removes the prefix of
+    /// records the server has already durably stored (offset ≤ resume watermark)
```
teodordelibasic-db: nit: Let's use `<=` maybe. 😄

rust/sdk/src/stream/grpc/transport.rs
```diff
@@ -0,0 +1,386 @@
...
+                    payload: Some(payload),
+                })
+                .await
+                .map_err(|_| Self::open_failed())
+            }
+            // The sink and kind are always constructed together by
+            // `make_outbound`, so a mismatch is unreachable. Only present when
+            // more than one variant exists (i.e. `eos` is enabled).
+            #[cfg(feature = "eos")]
+            _ => Err(Self::open_failed()),
```
teodordelibasic-db: Hm can the `OutboundSink` and `TransportKind` enums be merged somehow so we don't have to do this double matching? I don't have a specific idea in mind.

rust/sdk/src/stream/grpc/transport.rs
```diff
@@ -0,0 +1,386 @@
...
+                tx.send(PersistentStreamRequest {
+                    payload: Some(payload),
+                })
+                .await
+                .map_err(|_| Self::send_failed())
+            }
+        }
+    }
+
+    fn send_failed() -> ZerobusError {
```
teodordelibasic-db: nit: The naming matrix of these two methods and their helper error methods are a bit mixed up I'd say.

`send_open` - `open_failed`.
`send_ingest` - `send_failed`, I'd say the second one here should be `ingest_failed`. I get the sentiment behind `send_failed`, we are _sending_ a batch, but WDYT?

rust/sdk/src/stream/grpc/transport.rs
```diff
@@ -0,0 +1,386 @@
...
+        ZerobusError::StreamClosedError(tonic::Status::internal("Failed to send record"))
```
teodordelibasic-db: nit: Probably pre-existing, but `"Failed to send batch"` is more suitable.

rust/sdk/src/stream/grpc/transport.rs
```diff
@@ -0,0 +1,386 @@
...
+                let payload = batch.into_request_payload(offset_id);
+                tx.send(EphemeralStreamRequest {
+                    payload: Some(payload),
+                })
+                .await
+                .map_err(|_| Self::send_failed())
+            }
+            #[cfg(feature = "eos")]
+            OutboundSink::Persistent(tx) => {
+                let payload = ingest_payload_to_persistent(batch.into_request_payload(offset_id));
```
teodordelibasic-db: Can we have a `batch.into_persistent_request_payload` or something like that?

rust/sdk/src/stream/grpc/transport.rs
```diff
@@ -0,0 +1,386 @@
...
+    /// The server will close the current session after the given duration.
+    Close(CloseStreamSignal),
+    /// Any other payload (create/resume responses arrive only while opening and
+    /// are handled there; seeing one mid-stream is unexpected).
+    Other,
+}
+
+/// The inbound half of a stream: wraps the concrete tonic response stream and
+/// yields normalized `InboundMessage`s.
+pub(super) enum InboundStream {
```
teodordelibasic-db, line 202: Similar question as https://github.com/databricks/zerobus-sdk/pull/762/changes#r3851952011 if it makes sense, although here the sentiment is not as strong since we have no double matching. In general looks good regarding division of responsibilities, but I feel a bit we have two many enums where one variant is `Ephemeral` and the other is `Persistent`.

rust/sdk/src/stream/grpc/supervisor.rs
```diff
@@ -190,6 +223,31 @@ impl ZerobusStream {
             // 2. Reset landing zone.
             let resent_records = landing_zone_recovery.observed_count();
             let resent_batches = landing_zone_recovery.reset_observe();
+
+            // Resume alignment (persistent streams). The server reports how far
```
teodordelibasic-db: We should remove the `Decision 6` part and in general comment can probably be shorter.

rust/sdk/src/stream/grpc/connection.rs
```diff
@@ -1,51 +1,55 @@
...
+use crate::databricks::zerobus::{CreateIngestStreamRequest, RecordType};
+use crate::{HeadersProvider, OffsetId, TableProperties, ZerobusError, ZerobusResult};
+
+/// A freshly opened stream connection: the outbound sink, the inbound response
+/// stream, the server-assigned `stream_id`, and (persistent resume only) the
+/// committed-offset watermark to resume from.
+pub(super) struct StreamConnection {
+    pub(super) sink: OutboundSink,
+    pub(super) inbound: InboundStream,
+    pub(super) stream_id: String,
```
teodordelibasic-db: These two last fields are kind of duplicated with `StreamInitInfo`.

rust/sdk/src/stream/grpc/transport.rs
```diff
@@ -0,0 +1,386 @@
...
+                        Ok(Opened {
+                            stream_id,
+                            last_committed_offset: None,
+                        })
+                    }
+                    other => Err(Self::unexpected_open(&other)),
+                }
+            }
+            #[cfg(feature = "eos")]
+            InboundStream::Persistent(s) => {
```
teodordelibasic-db, line 252: Mismatched setup responses currently succeed. Both `CreateStreamResponse` and `ResumeStreamResponse` are accepted regardless of whether the client requested a create or resume. The fallback in [`connection.rs:169`](https://github.com/databricks/zerobus-sdk/blob/12c4957c53dd41edcba6970eac90b14f1e0f783e/rust/sdk/src/stream/grpc/connection.rs#L169) then lets a resume adopt the ID from an unexpected create response, while a create can accept an unexpected resume response without receiving a stream ID.

Could we preserve the response variant and validate it against the requested operation, for example:

```rust
enum Opened {
    Created { stream_id: String },
    Resumed { last_committed_offset: Option<i64> },
}
```

The create path should only accept `Opened::Created`, and the resume path should only accept `Opened::Resumed`.

rust/sdk/src/stream/grpc/supervisor.rs
teodordelibasic-db, line 109: Maybe we should also note the idempotency hole here when create persistent stream fails but client gets no response from server?

rust/sdk/src/stream/grpc/receiver.rs
```diff
@@ -86,98 +91,94 @@ impl ZerobusStream {
...
+                                    ZerobusError::StreamClosedError(tonic::Status::internal(
+                                        "Missing ack offset in server response",
+                                    ));
+                                let _ = server_error_tx.send(Some(error.clone()));
+                                return Err(error);
+                            }
+                        };
+                        let mut last_logical_acked_offset = -2;
+                        let mut map = oneshot_map.lock().await;
+                        for _offset_to_ack in (last_acked_offset + 1)..=durability_ack_up_to_offset
```
teodordelibasic-db, line 128: Let's reject an ACK lower than `last_acked_offset` or higher than the greatest sent wire offset before removing anything from `landing_zone` and close the stream on either protocol violation. We added such checks recently to the Arrow Flight SDK. It's an edge case for a malformed server, but worth just filling all the gaps.

rust/sdk/src/stream/grpc/mod.rs
```diff
@@ -186,20 +247,37 @@ impl ZerobusStream {
...
+        // watch to the watermark: everything up to it is already durable, so a
+        // `flush()` / `wait_for_offset()` targeting an already-committed offset
+        // must resolve immediately. Without this the watch stays `None` and, on a
+        // fresh-process resume with no new ingests, `flush()` (and the `close()`
+        // that calls it) would block until the flush timeout and then error,
+        // since the server never re-acks offsets it committed in a prior session.
+        // Safe because the constructor returns before any user ingest, so no real
+        // ack can race this initial value.
+        if let Some(watermark) = init_info.last_committed_offset {
+            logical_offset_id_generator.set_next(watermark + 1);
```
teodordelibasic-db: Let's use checked arithmetic here. A peer can return `i64::MAX` for this `int64` field, at which point `watermark + 1` panics when overflow checks are enabled and can wrap to a negative offset otherwise.

```rust
let next_offset = watermark.checked_add(1).ok_or_else(|| {
    ZerobusError::UnexpectedStreamResponseError(
        "Persistent stream offset space is exhausted".to_string(),
    )
})?;
logical_offset_id_generator.set_next(next_offset);
```

rust/sdk/Cargo.toml
elenagaljak-db: Fixed

rust/sdk/src/landing_zone.rs
elenagaljak-db: Fixd

rust/sdk/src/stream/grpc/transport.rs
elenagaljak-db: I removed the (sink, mode) double match. OutboundConnection now keeps the matching sender and request stream together, and open_rpc constructs the corresponding sink.

rust/sdk/src/stream/grpc/transport.rs
elenagaljak-db: I renamed to ingest_failed

rust/sdk/src/stream/grpc/transport.rs
elenagaljak-db: Yep, I updated anyway

rust/sdk/src/stream/grpc/transport.rs
elenagaljak-db: Added EncodedBatch::into_persistent_request_payload, matching the ephemeral conversion helper.

rust/sdk/src/stream/grpc/transport.rs
elenagaljak-db: I removed OutboundRequestStream and pairing the outbound sender/request stream in OutboundConnection. I kept InboundStream because tonic gives us different concrete response-stream types, and Opened because we need to preserve whether the server acknowledged create or resume for validation.

rust/sdk/src/stream/grpc/supervisor.rs
elenagaljak-db: Yep, removed

rust/sdk/src/stream/grpc/connection.rs
elenagaljak-db: Removed

rust/sdk/src/stream/grpc/transport.rs
elenagaljak-db: Fixed

rust/sdk/src/stream/grpc/supervisor.rs
elenagaljak-db: Added a comment documenting the remaining hole: if create succeeds server-side but its response is lost, retrying may create an orphan because no client-supplied ID exists yet. Resume remains idempotent.

rust/sdk/src/stream/grpc/receiver.rs
elenagaljak-db: Added validation before touching the landing zone

rust/sdk/src/stream/grpc/mod.rs
elenagaljak-db: Done

rust/sdk/Cargo.toml
```diff
@@ -75,6 +75,8 @@ internal-arrow-c-data = [
 ]
 # Zero-copy protobuf parser.
 zeroparser = ["dep:self_cell", "dep:prost-build"]
+# Persistent streams with exactly-once semantics (EoS); API is in development
```
teodordelibasic-db: nit: Period at the end.

rust/sdk/src/stream/grpc/supervisor.rs
```diff
@@ -54,7 +54,7 @@ impl ZerobusStream {
         options: StreamConfigurationOptions,
         // Mutated only on the persistent recovery path (flip create → resume);
         // that mutation is `eos`-gated, so `mut` is otherwise unused.
-        #[cfg_attr(not(feature = "eos"), allow(unused_mut))] mut kind: GrpcConnectionMode,
+        mut kind: GrpcConnectionMode,
```
teodordelibasic-db, line 56: Why did we remove the feature gating for mutability, also is the comment above now stale?

rust/sdk/src/stream/grpc/mod.rs
```diff
@@ -186,20 +245,42 @@ impl ZerobusStream {
...
+        // watch to the watermark: everything up to it is already durable, so a
+        // `flush()` / `wait_for_offset()` targeting an already-committed offset
+        // must resolve immediately. Without this the watch stays `None` and, on a
+        // fresh-process resume with no new ingests, `flush()` (and the `close()`
+        // that calls it) would block until the flush timeout and then error,
+        // since the server never re-acks offsets it committed in a prior session.
+        // Safe because the constructor returns before any user ingest, so no real
+        // ack can race this initial value.
+        if let Some(watermark) = init_info.last_committed_offset {
+            let next_offset = watermark.checked_add(1).ok_or_else(|| {
```
teodordelibasic-db: The supervisor and callback task are spawned before the `checked_add` and if that call errors out their join handles are dropped which detaches them instead of cancelling. We can cancel the tasks before returning the error:

```rust
let Some(next_offset) = watermark.checked_add(1) else {
    cancellation_token.cancel();
    supervisor_task.abort();
    if let Some(task) = callback_handler_task.as_ref() {
        task.abort();
    }
    return Err(ZerobusError::UnexpectedStreamResponseError(
        "Persistent stream offset space is exhausted".to_string(),
    ));
};
```

rust/sdk/src/stream/grpc/supervisor.rs
```diff
@@ -176,20 +200,55 @@ impl ZerobusStream {
...
             }
 
-            // 2. Reset landing zone.
+            // A dropped persistent connection can lose ACKs for records the
+            // server already committed. During automatic recovery only, resolve
+            // the committed observed prefix locally before resetting the
+            // remaining observed records for resend. A first-process resume has
+            // an empty landing zone and only seeds the receiver watermark.
+            let mut initial_last_acked_offset: OffsetId = -1;
+            if let Some(watermark) = last_committed_offset {
```
teodordelibasic-db, line 233: Let's just leave a `TODO` immediately before this watermark handling so the missing validation is not lost? It should record that automatic resume must validate the normalized watermark against the previous ACK and highest successfully sent wire offset before reconciling, including `None` as the no-commit position `-1`.

```rust
// TODO: Validate the resumed watermark against retained ACK and sent-offset
// bounds before reconciling; treat None as -1 and fail closed if out of range.
```

rust/sdk/src/stream/grpc/receiver.rs
```diff
@@ -86,98 +91,102 @@ impl ZerobusStream {
...
-                                Some(offset) => offset,
-                                None => {
-                                    error!("Missing ack offset in server response");
-                                    let error =
-                                        ZerobusError::StreamClosedError(tonic::Status::internal(
-                                            "Missing ack offset in server response",
-                                        ));
-                                    let _ = server_error_tx.send(Some(error.clone()));
-                                    return Err(error);
+                            landing_zone.observed_count(),
```
teodordelibasic-db, line 121: Let's leave a `TODO` at the `observed_count()` argument noting that ACK bounds need the highest successfully sent wire offset rather than a landing-zone count. `observed_count()` counts records even though a batch has one wire offset, and counting observed batches alone is also insufficient because [`observe`](https://github.com/databricks/zerobus-sdk/blob/a02b3363fbe2d9c7b40a1c253ce3408ea0e1f78a/rust/sdk/src/landing_zone.rs#L165) runs before the send succeeds.

```rust
// TODO: Bound ACKs by the highest successfully sent wire offset, not landing-zone
// counts; batches use one offset and observation precedes a successful send.
```

rust/sdk/src/stream/grpc/mod.rs
```diff
@@ -186,20 +245,42 @@ impl ZerobusStream {
...
+        // offsets rather than restarting at 0. Also seed the last-received-offset
+        // watch to the watermark: everything up to it is already durable, so a
+        // `flush()` / `wait_for_offset()` targeting an already-committed offset
+        // must resolve immediately. Without this the watch stays `None` and, on a
+        // fresh-process resume with no new ingests, `flush()` (and the `close()`
+        // that calls it) would block until the flush timeout and then error,
+        // since the server never re-acks offsets it committed in a prior session.
+        // Safe because the constructor returns before any user ingest, so no real
+        // ack can race this initial value.
+        if let Some(watermark) = init_info.last_committed_offset {
```
teodordelibasic-db, line 271: Let's leave a `TODO` before this fallible post-initialization block noting that the supervisor and optional callback task must be cleaned up if watermark validation fails. Dropping their `JoinHandle`s detaches them rather than cancelling them.

```rust
// TODO: Validate the resume watermark before spawning tasks, or cancel and abort
// the spawned tasks on post-initialization errors so their handles are not detached.
```

teodordelibasic-db (dismissed):
Shape looks good, let's just leave a couple of TODO comments for non-blocking hardening work for some edge cases.

rust/sdk/Cargo.toml
elenagaljak-db: Fixed

rust/sdk/src/stream/grpc/supervisor.rs
elenagaljak-db: Removed the comment. The code was getting really mess with the eos feature gate

rust/sdk/src/stream/grpc/supervisor.rs
elenagaljak-db: Added

rust/sdk/src/stream/grpc/receiver.rs
elenagaljak-db: Added

rust/sdk/src/stream/grpc/mod.rs
elenagaljak-db: Added

teodordelibasic-db (approved):

PR #781, opened 2026-08-24
[CI] Simplify cross-SDK Rust testing

teodordelibasic-db:

This simplifies cross-SDK CI around a single dependency model: wrapper source builds consume the Rust SDK from the repository path.

The workflows no longer patch Cargo manifests at runtime or expose a `use_local_sdk` mode. Cross-SDK jobs also skip wrappers already selected by their own path filters, avoiding duplicate test matrices when a pull request changes both Rust and a wrapper. Go and .NET retain a narrowly scoped input controlling their published-FFI-artifact smoke tests.

This PR should merge after #780, which makes the TypeScript path dependency permanent. The branches modify separate files and can merge without a textual conflict.

## How is this tested?

- Parsed all seven modified workflow files as YAML.
- Verified reusable-workflow callers use only declared inputs.
- Verified the diff has no whitespace errors.

Each edited reusable workflow is selected by its workflow path filter in this PR, so the pull request CI run will exercise the affected SDK job matrices.

### DIFF
--- a/.github/workflows/ci-cpp.yml
+++ b/.github/workflows/ci-cpp.yml
@@ -1,10 +1,6 @@
 name: C++ CI
 
 on:
-  # No use_local_sdk input: the C++ build always compiles the Rust core from
-  # local source (BuildRustFfi.cmake runs `cargo build` in ../rust/ffi, which
-  # depends on ../rust/sdk by path), so there is no released-vs-local mode to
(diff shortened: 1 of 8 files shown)


elenagaljak-db (dismissed):
lgtm

elenagaljak-db (approved):

PR #794, opened 2026-08-31
[All] Expose per-stream connection option

danilonajkov-db:
## Summary

- Expose the Rust connection-per-stream setting through C FFI, JNI, Go, C++, .NET, Java, Python, and TypeScript.
- Keep dedicated connections enabled by default and allow shared HTTP/2 multiplexing as an opt-out.
- Document the throughput tradeoff in every included SDK without changing examples.
- PureGo is intentionally excluded and will be handled separately.
- Rebuild the five checked-in Go FFI archives from the same local Rust source; no prerequisite Rust, FFI, or JNI release is required.

## Motivation

HTTP/2 multiplexes logical streams over one TCP connection. For high-throughput workloads over the public internet, packet loss and TCP retransmissions can cause head-of-line blocking across every stream on that connection, reducing aggregate throughput.

Dedicated connections isolate streams from this shared loss domain. For many smaller, low-throughput streams, shared multiplexing remains recommended to reduce connection overhead.

## Testing

- Rust FFI: 90 tests passed
- Rust JNI: 4 tests passed
- Go cgo: full unit and mock-server suites passed against rebuilt FFI
- C++: build and all 14 tests passed
- Python: native extension built and all 174 tests passed
- TypeScript: native module built and the connection-option test passed; 30 of 31 full-suite tests passed, with the remaining packaging test blocked by the pre-existing unavailable `protobufjs@^8.7.1` dependency
- Live staging ingestion: Rust, Python, TypeScript, Go cgo, C, C++, Java, and .NET each passed with one default stream, two default/dedicated streams, and two explicitly shared streams

### DIFF
--- a/dotnet/README.md
+++ b/dotnet/README.md
@@ -85,6 +85,16 @@ using var sdk = ZerobusSdk.CreateBuilder()
     .Build();
 ```
 
+`ConnectionPerStream` applies to JSON/protobuf streams. Arrow Flight streams
+already use dedicated connections and are unaffected.
+
+HTTP/2 multiplexes logical streams over one TCP connection. On high-throughput
+workloads over the public internet, packet loss and TCP retransmissions can
+cause head-of-line blocking across every stream on that connection, reducing
+aggregate throughput. Dedicated connections isolate that loss. For many
+smaller, low-throughput streams, call `ConnectionPerStream(false)`; shared
+multiplexing is recommended there to reduce connection overhead.
+
 #### `CreateJsonStream`
 
 Creates a JSON-only stream with OAuth 2.0 client credentials authentication.
--- a/dotnet/tests/Zerobus.Tests/ZerobusSdkBuilderTests.cs
+++ b/dotnet/tests/Zerobus.Tests/ZerobusSdkBuilderTests.cs
@@ -72,6 +72,16 @@ public class ZerobusSdkBuilderTests
         Assert.That(returned, Is.SameAs(builder));
     }
 
+    [Test]
+    public void ConnectionPerStream_ReturnsSameBuilderInstance()
+    {
+        using var builder = ZerobusSdk.CreateBuilder();
+
+        var returned = builder.ConnectionPerStream(false);
+
+        Assert.That(returned, Is.SameAs(builder));
+    }
+
     // -------------------------------------------------------------------------
     // Null argument guards
     // -------------------------------------------------------------------------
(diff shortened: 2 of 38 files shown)


rust/jni/src/sdk.rs
```diff
@@ -52,7 +52,7 @@ impl NativeSdkHandle {
 ///
 /// # JNI Signature
 /// ```java
-/// private static native long nativeCreate(String serverEndpoint, String unityCatalogEndpoint, String applicationName);
+/// private static native long nativeCreate(String serverEndpoint, String unityCatalogEndpoint, String applicationName, boolean connectionPerStream);
 /// ```
 #[no_mangle]
 pub extern "system" fn Java_com_databricks_zerobus_ZerobusSdk_nativeCreate<'local>(
```
teodordelibasic-db: It's maybe better to leave the three-argument `nativeCreate` as-is and add a new native method for the boolean, the same way `nativeCreateStreamWithHeadersProvider` was added? A Maven bump ships matching classes and natives, so this is hygiene, not a client break.

```java
private static native long nativeCreate(
    String serverEndpoint, String unityCatalogEndpoint, String applicationName);

private static native long nativeCreateWithConnectionPerStream(
    String serverEndpoint,
    String unityCatalogEndpoint,
    String applicationName,
    boolean connectionPerStream);
```

dotnet/tests/Zerobus.Tests/ZerobusSdkBuilderTests.cs
```diff
@@ -72,6 +72,16 @@ public void DisableTls_ReturnsSameBuilderInstance()
         Assert.That(returned, Is.SameAs(builder));
     }
 
+    [Test]
+    public void ConnectionPerStream_ReturnsSameBuilderInstance()
```
teodordelibasic-db: This never calls `Build()`, so the new P/Invoke is not exercised. Could we build after setting the option?

```csharp
[Test]
public void ConnectionPerStream_Build_Succeeds()
{
    var builder = ZerobusSdk.CreateBuilder()
        .Endpoint("https://zerobus.databricks.com")
        .UnityCatalogUrl("https://workspace.databricks.com")
        .ConnectionPerStream(false);

    using var sdk = builder.Build();
    Assert.That(sdk, Is.Not.Null);
}
```

rust/ffi/NEXT_CHANGELOG.md
teodordelibasic-db, line 15: The addition to `New Features and Improvements` may be more suitably placed here.

dotnet/README.md
```diff
@@ -85,6 +85,16 @@ using var sdk = ZerobusSdk.CreateBuilder()
     .Build();
 ```
 
+`ConnectionPerStream` applies to JSON/protobuf streams. Arrow Flight streams
```
teodordelibasic-db: Maybe makes more sense to entirely drop the second sentence for Arrow as .NET SDK has no Arrow support currently.

teodordelibasic-db (commented):
LGTM, mostly nits.

teodordelibasic-db (approved):
LGTM

PR #795, opened 2026-08-31
[Rust] Isolate connections per stream

danilonajkov-db:
## Summary

- Give every JSON/protobuf stream a dedicated gRPC connection by default.
- Add `connection_per_stream(bool)`; setting it to `false` retains shared HTTP/2 multiplexing.
- Leave Arrow Flight unchanged because it already creates dedicated connections.

## Motivation

On high-throughput workloads over the public internet, packet loss and TCP retransmissions can cause head-of-line blocking across every HTTP/2 stream sharing one connection, reducing aggregate throughput. Dedicated connections isolate that loss.

For many smaller, low-throughput streams, shared multiplexing remains recommended to reduce connection overhead.

## Testing

- `cd rust && make test`
- Live staging ingestion passed with one default stream, two default/dedicated streams, and two explicitly shared streams.


rust/README.md
```diff
@@ -826,6 +826,19 @@ Also accepts `0` or `no`.
 
 ## Configuration Options
 
+### SDK builder options
+
+| Method | Default | Description |
+|--------|---------|-------------|
+| `connection_per_stream(bool)` | `true` | Give each JSON/protobuf stream a dedicated gRPC connection. Pass `false` to multiplex streams over one shared HTTP/2 connection. Arrow Flight streams are unaffected because they already use dedicated connections. |
```
teodordelibasic-db, line 837: Maybe we should list out a few preexisting options as well.

teodordelibasic-db (dismissed):
LGTM

rust/README.md
danilonajkov-db: yep done

teodordelibasic-db (approved):

PR #811, opened 2026-09-01
[All] Add avro record type to proto surface

irinatomic-db:
## PR stack

- [avro/proto-surface](https://github.com/databricks/zerobus-sdk/pull/811) <- _this PR_
    - [avro/rust-core](https://github.com/databricks/zerobus-sdk/pull/827)
        - avro/java
        - avro/typescript

Adds the Avro record type to the proto surface (wire contract only) so later PRs can build Avro ingestion on ephemeral streams. No public SDK API and no behavior change yet.

- Proto (`rust/sdk/zerobus_service.proto`): 
   - reserve 3 (ARROW_IPC on the service - this SDK models Arrow as a separate Arrow Flight stream, not a RecordType) 
   - add AVRO = 4; 
   - add `AvroRecordBatch`, `CreateIngestStreamRequest.avro_schema_json`, `IngestRecordRequest.avro_encoded_record`, and `IngestRecordBatchRequest.avro_batch`.
- purego: regenerate `internal/zerobuspb` bindings from the schema. The commit also resyncs pre-existing drift (proto doc-comments + gofmt) and adds a generate CI job that fails if the committed bindings are stale = the gate purego was missing.
- Rust: set `avro_schema_json`: None at the one `CreateIngestStreamRequest` construction site so codegen compiles.

## How is this tested?

```
# Rust - proto compiles + no regressions
cd rust
cargo build -p databricks-zerobus-ingest-sdk
cargo test  -p databricks-zerobus-ingest-sdk

# purego - bindings match the schema (the new CI gate), plus build/vet/test
cd ../purego/internal/zerobuspb
go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.36.11
go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.6.2
go generate ./...
cd ../../..
git diff --exit-code -- purego/internal/zerobuspb/   # clean = bindings in sync

cd purego
go build ./...
go vet ./...
go test ./...
```

## Code diff
Most of the code diff is generated (`purego/internal` files): +1005 -87

### DIFF
--- a/rust/ffi/src/stream.rs
+++ b/rust/ffi/src/stream.rs
@@ -117,6 +117,11 @@ async fn build_stream_from_parts(
             base.compiled_proto(desc)
         }
         RecordType::Json => base.json(),
+        RecordType::Avro => {
+            return Err(ZerobusError::InvalidArgument(
+                "Avro record type is not supported".to_string(),
+            ))
+        }
         RecordType::Unspecified => {
             return Err(ZerobusError::InvalidArgument(
                 "Record type is not specified".to_string(),
--- a/rust/sdk/zerobus_service.proto
+++ b/rust/sdk/zerobus_service.proto
@@ -68,7 +68,6 @@ enum RecordType {
   RECORD_TYPE_UNSPECIFIED = 0;
   PROTO = 1;
   JSON = 2;
-  // 3 is ARROW_IPC on the Zerobus service. 
   reserved 3;
   AVRO = 4;
 }
@@ -131,7 +130,7 @@ message CreateIngestStreamRequest {
   // This descriptor defines the structure of the records being ingested.
   // It must be compatible with the target table's schema.
   //
-  // This is a required field for all stream creation requests.
+  // Required only for streams with record_type = PROTO; ignored otherwise.
   optional bytes descriptor_proto = 3;
 
   // Record type that will be accepted in the stream.
@@ -195,7 +194,7 @@ message IngestRecordBatchRequest {
   optional int64 offset_id = 1;
 
   // Batch of serialized records.
-  // The batch can contain multiple records encoded as either protobuf or JSON.
+  // The batch can contain multiple records encoded as protobuf, JSON, or Avro.
   oneof batch {
     // Batch of protobuf-encoded records. Each record must be serialized according to
     // the protobuf descriptor provided in the CreateIngestStreamRequest.
--- a/rust/sdk/zerobus_service.proto
+++ b/rust/sdk/zerobus_service.proto
@@ -68,6 +68,9 @@ enum RecordType {
   RECORD_TYPE_UNSPECIFIED = 0;
   PROTO = 1;
   JSON = 2;
+  // 3 is ARROW_IPC on the Zerobus service. 
+  reserved 3;
+  AVRO = 4;
 }
 
 /*
@@ -90,6 +93,19 @@ message JsonRecordBatch {
  */
 message ProtoEncodedRecordBatch {
   // Array of protobuf-encoded records.
+  repeated bytes records = 1;
+}
+
+/*
+ * Batch of Avro-encoded records.
+ *
+ * Each element is a single raw Avro binary datum (no framing). The writer
+ * schema is supplied once at stream creation via
+ * CreateIngestStreamRequest.avro_schema_json.
+ */
+message AvroRecordBatch {
+  // Array of raw Avro binary datums. Each datum must be encoded according to
+  // the writer schema provided in CreateIngestStreamRequest.avro_schema_json.
   repeated bytes records = 1;
 }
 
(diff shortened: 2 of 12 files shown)


danilonajkov-db (dismissed):
LGTM, just check with Teodor about the CI

rust/sdk/zerobus_service.proto
```diff
@@ -68,6 +68,9 @@ enum RecordType {
   RECORD_TYPE_UNSPECIFIED = 0;
   PROTO = 1;
   JSON = 2;
+  // 3 is ARROW_IPC on the Zerobus service. 
+  reserved 3;
+  AVRO = 4;
```
teodordelibasic-db, line 72: Ah, this is a preexisting issue but has only now surfaced/come to my attention. Adding this is a potential breaking change for some clients. The code that would break is not useful in any way I think, but it's still breaking, for example:

```rust
use databricks_zerobus_ingest_sdk::databricks::zerobus::RecordType;
use databricks_zerobus_ingest_sdk::StreamConfigurationOptions;
fn describe(opts: &StreamConfigurationOptions) -> &'static str {
  match opts.record_type {
    RecordType::Proto => "proto",
    RecordType::Json => "json",
    RecordType::Unspecified => "unspecified",
  }
}
```

After this PR this match would break. This is true also for some interactions with other generated types from this file. Due to this you had to change wrapper SDKs in this change as well.

IMO we shouldn't treat this as a regular API breaking change that requires a major version bump, but instead just note it in `NEXT_CHANGELOG.md` and then handle it later the proper way during a major version bump so that we don't have to think about it anymore. I've created an issue for this long term fix: [[Rust] Hide generated gRPC types so additive proto changes are not crate-breaking](https://github.com/databricks/zerobus-sdk/issues/822).

Let's discuss with other folks offline as well. cc: @davidtosovic-db

rust/sdk/zerobus_service.proto
```diff
@@ -68,6 +68,9 @@ enum RecordType {
   RECORD_TYPE_UNSPECIFIED = 0;
   PROTO = 1;
   JSON = 2;
+  // 3 is ARROW_IPC on the Zerobus service.
```
teodordelibasic-db: No need for this comment, we can either just say `reserved 3;` or actually copy the proto as is, but I lean towards just having `reserved 3;`.

rust/sdk/zerobus_service.proto
teodordelibasic-db: Preexisting as well, but could you update this to say it's required only for streams with record_type equal to PROTO.

teodordelibasic-db: This is also outdated now: `either protobuf or JSON`.

rust/ffi/src/stream.rs
```diff
@@ -117,6 +117,11 @@ async fn build_stream_from_parts(
             base.compiled_proto(desc)
         }
         RecordType::Json => base.json(),
+        RecordType::Avro => {
```
teodordelibasic-db, line 120: This arm's only use is so that the new prost variant compiles. Callers never construct `RecordType::Avro`: [`c_record_type`](https://github.com/databricks/zerobus-sdk/blob/e0d6cc5b09682083f7309f910727d8acd81aa849/rust/ffi/src/common.rs#L183) maps 1 to Proto, 2 to Json, and everything else (including 4) to Unspecified, so a C/Go/C++ caller passing `4` gets "Record type is not specified" instead of this error. JNI is also in a bad state as `is_json` is only proto vs json.

Could we map 4 so this arm actually runs?

```rust
pub(crate) fn c_record_type(value: i32) -> RecordType {
    match value {
        1 => RecordType::Proto,
        2 => RecordType::Json,
        4 => RecordType::Avro,
        _ => RecordType::Unspecified,
    }
}
```

And in `test_c_record_type_mapping`: `assert_eq!(c_record_type(4), RecordType::Avro);`

typescript/src/lib.rs
```diff
@@ -1086,6 +1086,11 @@ impl ZerobusSdk {
 
                 let builder = match record_type {
                     RustRecordType::Json => builder.json(),
+                    RustRecordType::Avro => {
```
teodordelibasic-db, line 1092: Same as the FFI comment. The mapping a few lines up is `Some(0) => Json`, `Some(1) => Proto`, `_ => Proto`, so a raw `record_type: 4` silently opens a proto stream if a descriptor is present.

```rust
let record_type = match opts.record_type {
    Some(0) => RustRecordType::Json,
    Some(1) => RustRecordType::Proto,
    Some(4) => RustRecordType::Avro,
    _ => RustRecordType::Unspecified,
};
```

irinatomic-db: Fixed. But since the TS `RecordType` is 00-based I used Avro = 2.

rust/sdk/zerobus_service.proto
irinatomic-db: Updated to "protobuf, JSON, or Avro".

rust/ffi/src/stream.rs
irinatomic-db: Added `4 => RecordType::Avro` to `c_record_type`

teodordelibasic-db (approved):
LGTM

PR #820, opened 2026-09-01
[Rust] Add pluggable stats exporter for Arrow streams

flaviofcruz:

This proposes a different implementation compared to https://github.com/databricks/zerobus-sdk/pull/790?timeline_per_page=5. Instead of adding a new `take_offset_details`, we introduce a mechanism where a user of the SDK can set a stats exporter. The stats exporter can receive different kinds of SDK events such as BatchSent/BatchAcked/Reconnected which can be consumed by the exporter. As a basic exporter we have a channel exporter which can be used by clients to drain the events and decide what to do with them.

The only major complication on the PR is how we handle `BatchSent`: because we need uncompressed and wired bytes, these are computed in different places so they need to be coordinated. Ditto for the graceful shutdown case, we have to push `BatchSent` on that scenario.

Non arrow flight path doesn't support stats exporting for now.

Why: this is required so that clients can understand how many bytes are put on the wire.

## How is this tested?

Added a few tests.


rust/sdk/src/builder/stream_builder.rs
```diff
@@ -328,12 +332,28 @@ impl<'a> StreamBuilder<'a> {
...
+    /// The exporter receives [`StreamStat`](crate::StreamStat) events (batch
+    /// acknowledgements with byte sizes, failures, reconnects). Its `record` runs
+    /// inline on the stream's IO tasks, so keep it lightweight — use
+    /// [`channel_exporter`](crate::channel_exporter) to offload heavier work.
+    ///
+    /// Supported only by Arrow Flight streams: configuring it and then calling
+    /// [`build`](Self::build) (a non-Arrow stream) is rejected. Use
+    /// [`build_arrow`](Self::build_arrow).
+    #[cfg(feature = "arrow-flight")]
+    pub fn stats_exporter(mut self, exporter: Arc<dyn crate::stats::StatsExporter>) -> Self {
```
teodordelibasic-db: Could we keep `Arc<dyn StatsExporter>` as an internal storage detail and accept the concrete exporter by value here?

```rust
pub fn stats_exporter<E>(mut self, exporter: E) -> Self
where
    E: StatsExporter + 'static,
{
    self.stats_exporter = Some(Arc::new(exporter));
    self
}
```

The encoder and supervisor can still share it internally through `Arc`, while client code becomes `.stats_exporter(MyExporter)` and does not need to perform allocation or type erasure itself. Callers that need shared observable state can put that sharing inside their exporter. The `'static` bound belongs on this registration method because the SDK moves the exporter into spawned tasks; it does not need to be a supertrait of `StatsExporter`.

WDYT?

rust/sdk/src/stream/arrow/connection.rs
teodordelibasic-db, line 555: `BatchSent` is documented as encode/send time, but a finished batch stays in the slot until the next pull, EOF, or graceful close. After ingest plus `flush()`, the last batch has no event. Rotation then drops that slot even though the batch already acked and will not be replayed.

Could we emit when this `poll_next` returns `Pending` (encoder idle, frames already out), keep EOS as a backstop, and not flush on cancel?

```rust
match encoded.as_mut().poll_next(cx) {
    Poll::Ready(None) => {
        if let Some(exporter) = &stats_eof {
            flush_sent(exporter.as_ref(), &pending_eof);
        }
        state.finish();
        Poll::Ready(None)
    }
    Poll::Pending => {
        if let Some(exporter) = &stats_eof {
            flush_sent(exporter.as_ref(), &pending_eof);
        }
        Poll::Pending
    }
    result => result,
}
```

On cancel, skip an in-progress split that has not gone idle. A test that keeps `batch_tx` open, pulls schema plus data, then times out the next `next()` so `Pending` is polled, should see `BatchSent` with no shutdown.

rust/sdk/src/stats.rs
```diff
@@ -0,0 +1,142 @@
...
+/// Byte-size and count stats for an encoded batch.
+///
+/// Payload only — the offset that identifies the batch is carried separately on
+/// [`StreamStat::BatchSent`].
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct BatchStats {
+    /// Number of rows in the batch.
+    pub records: u64,
+    /// Bytes actually placed on the wire (Arrow IPC frame, after IPC compression
```
teodordelibasic-db: `wire_bytes` is FlightData payload after IPC compression (`data_header + data_body + app_metadata` on data messages), not NIC bytes. It skips protobuf, gRPC, HTTP/2/TLS, the schema message, and `flight_descriptor`. `uncompressed_bytes` is in-memory Arrow slice size, not uncompressed IPC, and `unwrap_or(0)` zeros a column on error.

Could we rename or document them as FlightData payload vs input buffers, and use the same fallback as the splitter?

```rust
.map(|col| {
    let data = col.to_data();
    data.get_slice_memory_size()
        .unwrap_or_else(|_| data.get_buffer_memory_size()) as u64
})
```

rust/sdk/src/stream/arrow/acks.rs
teodordelibasic-db: `last_ack_tx.send` runs before `record(BatchAcked)`, so `wait_for_offset` / `flush` can return before the event is visible, even for an inline exporter.

Does it make sense to record first?

```rust
if let Some(exporter) = self.stats_exporter {
    for offset in acked_offsets {
        exporter.record(StreamStat::BatchAcked { offset });
    }
}
if let Some(offset) = max_acked_offset {
    let _ = self.last_ack_tx.send(Some(offset));
    self.close.observe_ack(offset, applied_at);
}
```

rust/sdk/src/stats.rs
```diff
@@ -0,0 +1,142 @@
...
+    /// The stream reconnected after a transient failure.
+    Reconnected {
+        /// 1-based recovery attempt that succeeded.
+        attempt: u32,
+    },
+}
+
+/// Sink for stream telemetry. Register with `.stats_exporter(...)` on the builder.
+///
+/// `record` runs inline on the stream's hot / IO tasks, so it must be cheap and
```
teodordelibasic-db: `BatchSent` `record` runs inside tonic's request-body `poll_next`, so a blocking or allocating exporter stalls ingest. The same poll does `pending.lock().unwrap()`, so poison panics the request body.

Could we document `record` as atomics or `try_send` only, and recover from poison?

```rust
fn flush_sent(exporter: &dyn StatsExporter, pending: &StdMutex<Option<EncodeInProgress>>) {
    let done = pending.lock().unwrap_or_else(|e| e.into_inner()).take();
    // ...
}
```

rust/sdk/src/stream/arrow/batch.rs
```diff
@@ -215,7 +215,7 @@ pub(super) fn rebuild_pending_for_replay(
         let end_record = cumulative_records + record_count;
         cumulative_records = end_record;
 
-        replay.push(batch.clone());
+        replay.push((pending_batch.offset_id, batch.clone()));
```
teodordelibasic-db: Replay keeps the original `offset_id` and emits a second `BatchSent` for the unacked suffix, with different `records` and byte sizes. Summing by offset then double-counts, or reports retry size as the original ingest.

Could we document the second event as a retransmit? An `is_replay` flag (or `attempt`) would let a bytes-sent meter keep counting and a per-batch size keep the original.

rust/sdk/src/stats.rs
```diff
@@ -0,0 +1,142 @@
...
+        /// Byte-size and row-count stats for the batch.
+        stats: BatchStats,
+    },
+    /// A batch was durably acknowledged by the server. Pure durability signal —
+    /// byte sizes are reported by [`BatchSent`](Self::BatchSent).
+    BatchAcked {
+        /// Offset assigned to the batch by `ingest_batch`.
+        offset: OffsetId,
+    },
+    /// The stream reconnected after a transient failure.
```
teodordelibasic-db: It feels a bit off that the variant is documented as recovery after a transient failure, but it also fires on planned rotation, always as `attempt: 1`. Maybe we should add a `reason` field or similar?

rust/sdk/src/stats.rs
```diff
@@ -0,0 +1,142 @@
...
+    pub uncompressed_bytes: u64,
+}
+
+/// A telemetry event emitted by a stream. All variants are owned and `Copy`, so an
+/// event can be forwarded across a channel directly (see [`ChannelExporter`]).
+///
+/// An *offset* identifies one ingest call (one batch), not one row — a batch of
+/// 1000 rows has a single offset. Byte sizes are therefore per batch.
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
teodordelibasic-db: `Copy` means every field must stay a number or a small enum. A later failure event that carries a `ZerobusError` cannot be `Copy`. Adding it would break every caller that copies events today.

Maybe it makes sense to drop `Copy` and keep `Clone`?`BatchStats` is `#[non_exhaustive]` with no constructor, so it also needs `BatchStats::new(...)`.

```rust
#[non_exhaustive]
#[derive(Debug, Clone, PartialEq, Eq)]
pub enum StreamStat { /* ... */ }

impl BatchStats {
    pub fn new(records: u64, wire_bytes: u64, uncompressed_bytes: u64) -> Self {
        Self { records, wire_bytes, uncompressed_bytes }
    }
}
```

rust/sdk/src/builder/stream_builder.rs
flaviofcruz: I like this proposal since it requires no allocs from the client

rust/sdk/src/stats.rs
flaviofcruz: Fair enough. Update comments and renamed field to `approximate_wire_bytes`.

rust/sdk/src/stats.rs
flaviofcruz: Done

rust/sdk/src/stats.rs
flaviofcruz: separated with a enum to know whether there was a reconnection event vs transient failure

rust/sdk/src/stream/arrow/batch.rs
flaviofcruz: added `attempt` which tracks the current attempt. note that there's a corner case where we might not encode the whole batch and in that case, we start with `attempt = 1`, documented in the exporter.

rust/sdk/src/stats.rs
flaviofcruz: Good point, done.

rust/sdk/src/stream/arrow/acks.rs
flaviofcruz: Indeed, done!

rust/sdk/src/stats.rs
flaviofcruz: Also added a new `frame_stats` which also includes dictionary encoded size and also nested arrays

rust/sdk/src/stream/arrow/connection.rs
flaviofcruz: Pending can still drop stats if the stream is cancelled. How about just counting when we encode the full batch? See above.

flaviofcruz (commented):
Also update the arrow example to include use of the new API.

rust/sdk/src/stream/arrow/batch.rs
teodordelibasic-db: Sounds good.

rust/sdk/src/stream/arrow/connection.rs
```diff
@@ -377,16 +460,38 @@ impl ZerobusArrowStream {
...
+        batch_rx: mpsc::Receiver<BatchItem>,
         table_properties: &ArrowTableProperties,
         options: &ArrowStreamConfigurationOptions,
+        stats_exporter: Option<Arc<dyn StatsExporter>>,
         #[cfg(feature = "test-hooks")] test_hooks: Arc<super::TestHooks>,
     ) -> ZerobusResult<(FlightRequestStream, RequestBodyControl)> {
         let ipc_write_options = make_ipc_write_options(options.ipc_compression)?;
         let schema = Arc::clone(&table_properties.schema);
-        let batch_stream = tokio_stream::wrappers::ReceiverStream::new(batch_rx);
+        // When an exporter is registered, track the batch being encoded in a single
```
teodordelibasic-db, line 471: Personal nit: Maybe we can remove this comment or move it to a doc comment, feels weird having only one multi line comment in the middle of the method.

rust/sdk/src/stream/arrow/connection.rs
```diff
@@ -395,12 +500,16 @@ impl ZerobusArrowStream {
                 .enumerate()
                 .map(move |(idx, result)| {
                     result.map(|mut flight_data| {
+                        // Frame 0 is the schema; only data frames get offsets and batch stats.
```
teodordelibasic-db, line 503: Same nit, can maybe be in the doc comment.

rust/sdk/src/stream/arrow/connection.rs
```diff
@@ -425,6 +534,8 @@ impl ZerobusArrowStream {
                 if state.test.poll_shutdown_barrier(cx).is_pending() {
                     return Poll::Pending;
                 }
+                // Completed batches were already reported with their final frame.
```
teodordelibasic-db, line 537: ditto.

rust/sdk/src/stream/arrow/acks.rs
```diff
@@ -127,8 +129,9 @@ impl DrainState {
 pub(super) enum AckProcessOutcome {
     Stopped,
     Recovery {
-        error: ZerobusError,
         drained: bool,
+        /// Why recovery is needed; carries the triggering error for a transient failure.
```
teodordelibasic-db, line 133: nit: Unnecessary comment.

rust/sdk/src/stream/arrow/acks.rs
```diff
@@ -601,10 +631,18 @@ impl AckProcessor {
                 request,
                 outcome: terminal_error.map_or(Ok(()), Err),
             }),
-            None => Ok(AckProcessOutcome::Recovery {
-                error: terminal_error.unwrap_or_else(Self::rotation_error),
-                drained: true,
-            }),
+            None => {
+                // No terminal error means the drain completed a clean server-requested
```
teodordelibasic-db, line 635: nit: Unnecessary comment.

rust/sdk/src/stream/arrow/acks.rs
```diff
@@ -257,17 +261,22 @@ impl AckProgress<'_> {
                 .fetch_max(acked_records, Ordering::AcqRel);
             let effective_acked_records = previous_acked_records.max(acked_records);
             let mut max_acked_offset: Option<OffsetId> = None;
+            // Collect newly-acked offsets; BatchAcked is emitted after the lock is dropped.
```
teodordelibasic-db, line 264: nit: Unnecessary comment.

rust/sdk/src/stream/arrow/connection.rs
```diff
@@ -20,18 +20,100 @@ use tonic::metadata::MetadataValue;
...
+/// Accounts for one data frame, emitting `BatchSent` and clearing the slot on the final frame.
+fn record_sent_frame(
+    exporter: &dyn StatsExporter,
+    pending: &StdMutex<Option<EncodeInProgress>>,
+    frame: &FlightData,
+) {
+    let wire = (frame.data_header.len() + frame.data_body.len() + frame.app_metadata.len()) as u64;
+    // IPC row counts identify the final split frame without decoding the body.
+    // Dictionary messages contribute bytes but do not advance the row count.
+    let (records, uncompressed) = frame_stats(frame).unwrap_or_default();
```
teodordelibasic-db, line 87: Kind of an open semantic question.

`frame_stats` returning `None` becomes `(0, 0)` rows and bytes. Completion is `remaining_records` hitting 0, so that data frame adds `wire` but never finishes the slot. There is no EOS or cancel flush anymore, and the next pulled batch overwrites the slot. The encoder already finished the batch; telemetry did not.

Unlikely on frames this encoder just wrote, but `frame_stats` can still return `None` after it has identified a record batch, for example if `buffers()` or a compressed-length prefix cannot be read. That offset can then have `BatchAcked` and no `BatchSent`.

Could we keep record-batch length for completion when byte accounting fails, and only zero `uncompressed_bytes`?

```rust
let bytes = batch
    .buffers()
    .and_then(|buffers| {
        buffers.iter().try_fold(0_u64, |total, buffer| {
            // same length / compression-prefix fold as today
            Some(total)
        })
    })
    .unwrap_or(0);
Some((rows, bytes))
```

rust/sdk/src/builder/stream_builder.rs
teodordelibasic-db, line 406: Docs say `validate()` performs the same checks as `build()` without opening a stream. `.json().stats_exporter(...).validate()` returns `Ok(())`, then `build()` rejects it.

Could we run the same check in `validate()`?

```rust
#[cfg(feature = "arrow-flight")]
if self.stats_exporter.is_some()
    && !matches!(self.format, Some(FormatConfig::Arrow(_)))
{
    return Err(ZerobusError::InvalidArgument(
        "stats_exporter is only supported for Arrow Flight streams; use .build_arrow()"
            .into(),
    ));
}
```

A unit test next to `build_rejects_stats_exporter_on_non_arrow_stream` can assert `.json().stats_exporter(...).validate()` returns that `InvalidArgument`.

rust/docs/stats-exporter-design.md
teodordelibasic-db, line 1: Not sure if we should commit this? The file is also outdated after the change in the builder signature.

rust/sdk/src/stats.rs
```diff
@@ -0,0 +1,210 @@
...
+            // Full or closed: drop. Stats are best-effort; never block ingest.
+            self.dropped.fetch_add(1, Ordering::Relaxed);
+        }
+    }
+}
+
+/// Creates a [`ChannelExporter`] with a bounded buffer of `capacity` events and
+/// returns it with the receiver to drain elsewhere.
+pub fn channel_exporter(capacity: usize) -> (Arc<ChannelExporter>, mpsc::Receiver<StreamStat>) {
+    let (tx, rx) = mpsc::channel(capacity.max(1));
```
teodordelibasic-db, line 157: The docs say the buffer is `capacity` events. `capacity.max(1)` makes `channel_exporter(0)` a one-slot channel.

Could we reject zero, or document the coercion?

rust/sdk/src/stats.rs
```diff
@@ -0,0 +1,210 @@
...
+
+/// Sink for stream telemetry. Register with `.stats_exporter(...)` on the builder.
+///
+/// `record` runs **inline inside the stream's IO poll loop** (e.g. tonic's
+/// request-body `poll_next`), so it must return immediately: do only atomic updates
+/// or a non-blocking `try_send`. Blocking, heavy allocation, or awaiting stalls
+/// ingestion for every stream sharing that task. Use [`ChannelExporter`] to hand
+/// stats to another task for anything heavier.
+pub trait StatsExporter: Send + Sync {
+    /// Handle one telemetry event. Must not block, allocate heavily, or panic.
```
teodordelibasic-db, line 118: Docs say `record` must not panic. Poison on the slot mutex is recovered; a panic inside the user's `record` still unwinds tonic's request-body `poll_next`.

Does it make sense to do `catch_unwind` around `record` and count it as dropped, or do you think the documented contract is enough?

```rust
let _ = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
    exporter.record(event);
}));
```

teodordelibasic-db (dismissed):
LGTM, a couple of nits and open questions, feel free to ping for reapproval. Not sure we should commit the design file.

rust/examples/arrow/README.md
```diff
@@ -99,6 +105,26 @@ stream.close().await?;
 - **Schema validation**: Each `RecordBatch` must exactly match the client schema
   configured on the stream; the server validates that schema against the target
 
+## Telemetry
+
+The example registers `channel_exporter(1024)` with `.stats_exporter(...)` and
```
teodordelibasic-db: The API now takes `NonZeroUsize`. The example `main.rs` and the SDK README already use `NonZeroUsize::new(1024).unwrap()`. A reader who copies this caption gets a type error.

Could we match the working call?

```rust
use std::num::NonZeroUsize;

channel_exporter(NonZeroUsize::new(1024).unwrap())
```

rust/sdk/src/stream/arrow/connection.rs
```diff
@@ -376,17 +468,40 @@ impl ZerobusArrowStream {
 
     /// Builds a request body that can be stopped at a FlightData boundary and observed
```
teodordelibasic-db, line 469: The first paragraph says it builds a request body that can be stopped at a FlightData boundary. The next paragraph starts again with “Builds the Flight request stream…”. Could we merge them into one?

```rust
/// Builds a request body that can be stopped at a FlightData boundary and observed
/// reaching EOF without dropping the surrounding HTTP/2 stream. The stream is a
/// schema frame (index 0) followed by per-batch data frames, each tagged with its
/// wire offset. When an exporter is registered, per-batch stats accumulate in a
/// single slot and `BatchSent` is emitted with the batch's final frame, because
/// cancellation or drop may arrive before another poll.
```

rust/sdk/src/stats.rs
```diff
@@ -0,0 +1,241 @@
...
```
teodordelibasic-db: `dispatch` already catches a panic, logs, and drops the event. The trait line still says must not panic.

Could we say panics are caught and the event is dropped, and keep the “do not block” rule?

```rust
/// Handle one telemetry event. Must not block or allocate heavily.
/// A panic is caught by the SDK, logged, and the event is dropped.
fn record(&self, stat: StreamStat);
```

teodordelibasic-db (commented):
Just a few documentation nits and I am fine to merge.

rust/sdk/src/stats.rs
flaviofcruz: Good idea, the clients can define any exporter they want which can panic.

teodordelibasic-db (approved):
LGTM

rust/examples/arrow/README.md
flaviofcruz: Fixed

rust/sdk/src/stats.rs
flaviofcruz: Done.

PR #826, opened 2026-09-03
[Rust] Harden multiplexed stream lifecycle

danilonajkov-db:
## Summary

This PR prepares the existing multiplexed stream for the stream-builder API in the next PR in the stack. It does not introduce the builder API; `MultiplexedStream` remains behind the `testing` feature.

- Reject new ingestion after a mux operation observes a terminal lane failure, preserving the typed error. Detection is lazy; ingestion does not scan every lane.
- Keep acknowledgment waits scoped to the lane that owns the message. A sibling failure does not fail a healthy lane's wait.
- Flush every lane, including when the mux is already poisoned, and wait for all flush attempts before returning an error.
- Close with one concurrent all-lane flush, then finalize lanes and drain callbacks concurrently. Retain the completed flush result and terminal errors so an interrupted close can resume cleanup without flushing again.
- Leave healthy lanes running after poison until explicit close or drop. `is_closed()` reports mux-level close/poison state, not background lane health.

## Why

A sibling failure should not make a healthy, acknowledged record appear failed. Flush should still give healthy lanes a chance to drain, and shutdown time should not multiply by the number of lanes or repeat a completed flush on retry. Unacknowledged records remain available for recovery.

## Shared gRPC teardown

This PR also changes teardown internals used by ordinary `ZerobusStream`; it is not entirely mux-local. Supervisor shutdown waits up to one second, then at most 100 ms after abort. Shutdown outcomes and callback task handles are retained for resumable mux cleanup.

Ordinary stream public method signatures are unchanged. The shutdown bound applies to the supervisor wait, not the entire close operation: flush and callback draining have their own timeout settings, and synchronous user code cannot be forcibly interrupted by Tokio.

This is the prerequisite PR in the three-PR stack: lifecycle fixes → stream builder → documentation/examples.

## Testing

- `cargo test --workspace --quiet`
- `cargo test -p tests --test multiplexed_stream_tests --quiet` (22 tests)
- Three focused close regressions: bounded supervisor wait with blocked synchronous code, cancelled mux close without re-flushing, and terminal-error preservation across cancelled cleanup
- Clippy with warnings denied, including all-feature and no-default-feature checks
- `cargo fmt --all --check`

### DIFF
--- a/rust/NEXT_CHANGELOG.md
+++ b/rust/NEXT_CHANGELOG.md
@@ -12,9 +12,16 @@
 
 ### Bug Fixes
 
+- Fixed gRPC stream teardown to preserve unacknowledged records, return the
+  terminal stream error from `close()`, and reliably cancel and reap background
+  tasks.
+
 ### Documentation
 
 ### Internal Changes
+
+- Hardened multiplexed-stream failure propagation, acknowledgment wakeups,
+  capacity routing, and bounded concurrent shutdown.
 
 ### Breaking Changes
 
--- a/rust/NEXT_CHANGELOG.md
+++ b/rust/NEXT_CHANGELOG.md
@@ -21,6 +21,37 @@
 
 ### Breaking Changes
 
+- **Adding the Avro record type extends the publicly re-exported gRPC types, which
+  can break compilation for some downstream crates.** The additions
+  (`RecordType::AVRO`, `CreateIngestStreamRequest.avro_schema_json`,
+  `IngestRecordRequest.avro_encoded_record`, `IngestRecordBatchRequest.avro_batch`, and
+  the new `AvroRecordBatch`) are wire-compatible — no runtime or behavior change. But
+  because the crate re-exports the prost-generated types
+  (`databricks_zerobus_ingest_sdk::databricks::zerobus`), code that matches or builds
+  them exhaustively will no longer compile. The builder API (`.json()`,
+  `.compiled_proto()`, `.dynamic_proto()`) and the other-language SDKs are unaffected.
+  The long-term fix (hiding these generated types) is tracked in
+  [#822](https://github.com/databricks/zerobus-sdk/issues/822).
+
+  Migration — every fix is additive:
+  - Exhaustive `match` on `RecordType` with no `_` arm → add `_ => { … }` (or a
+    `RecordType::Avro` arm). (E0004)
+  - Exhaustive `match` on the `ingest_record_request::Record` or
+    `ingest_record_batch_request::Batch` oneof → add a `_ => { … }` arm. (E0004)
+  - Full-field struct literal of `CreateIngestStreamRequest` → add
+    `..Default::default()`. (E0063)
+  - Full-field struct pattern/destructuring of `CreateIngestStreamRequest`
+    (e.g. `let CreateIngestStreamRequest { table_name, descriptor_proto, record_type } = req`)
+    → add `..`. (E0027)
+
+  Unaffected: `x == RecordType::Json`, any `match` that already has a `_` arm, and
+  struct literals/patterns that already use `..`.
+
 ### Deprecations
 
 ### API Changes
+
+- Added the Avro wire surface to the generated gRPC types (`RecordType::AVRO`,
+  `CreateIngestStreamRequest.avro_schema_json`, `IngestRecordRequest.avro_encoded_record`,
+  `IngestRecordBatchRequest.avro_batch`, `AvroRecordBatch`). See **Breaking Changes** for
+  the downstream-compilation impact and migration.
--- a/rust/NEXT_CHANGELOG.md
+++ b/rust/NEXT_CHANGELOG.md
@@ -12,16 +12,12 @@
 
 ### Bug Fixes
 
-- Fixed gRPC stream teardown to preserve unacknowledged records, return the
-  terminal stream error from `close()`, and reliably cancel and reap background
-  tasks.
-
 ### Documentation
 
 ### Internal Changes
 
-- Hardened multiplexed-stream failure propagation, acknowledgment wakeups,
-  capacity routing, and bounded concurrent shutdown.
+- Updated multiplexed-stream failure handling to reject new ingestion and wake
+  pending message waits after a lane fails, and to close lanes concurrently.
 
 ### Breaking Changes
 
--- a/rust/sdk/src/multiplexed_stream.rs
+++ b/rust/sdk/src/multiplexed_stream.rs
@@ -19,71 +19,25 @@
 //! [`get_unacked_batches`](MultiplexedStream::get_unacked_batches).
 
 use futures::future::join_all;
+use futures::stream::FuturesUnordered;
+use futures::StreamExt;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Arc, OnceLock};
 use std::time::Duration;
 use tokio_util::sync::CancellationToken;
 use tracing::{error, info, warn};
 
-use crate::stream::StreamShutdownHandle;
+use crate::stream::StreamFailureHandle;
 use crate::{
     AckCallback, EncodedBatch, EncodedRecord, OffsetId, ZerobusError, ZerobusResult, ZerobusStream,
 };
 
 const CAPACITY_WAIT_TIMEOUT: Duration = Duration::from_secs(30);
 
-/// Ensures a cancelled poison future still shuts sub-stream tasks down, but
-/// only after all ingests admitted before poison have left the admission
-/// section. Normal completion disarms the fallback after signaling directly.
-struct PoisonCleanupGuard {
-    admission: Arc<tokio::sync::RwLock<()>>,
-    shutdown_handles: Option<Vec<StreamShutdownHandle>>,
-    runtime: tokio::runtime::Handle,
-}
-
-impl PoisonCleanupGuard {
-    fn new(
-        admission: Arc<tokio::sync::RwLock<()>>,
-        shutdown_handles: Vec<StreamShutdownHandle>,
-    ) -> Self {
-        Self {
-            admission,
-            shutdown_handles: Some(shutdown_handles),
-            runtime: tokio::runtime::Handle::current(),
-        }
-    }
-
-    fn finish(mut self) {
-        for handle in self.shutdown_handles.take().unwrap_or_default() {
-            handle.signal();
-        }
-    }
-}
-
-impl Drop for PoisonCleanupGuard {
-    fn drop(&mut self) {
-        let Some(shutdown_handles) = self.shutdown_handles.take() else {
-            return;
-        };
-
-        warn!("Mux poison cleanup was cancelled; finishing sub-stream shutdown in background");
-        let admission = Arc::clone(&self.admission);
-        drop(self.runtime.spawn(async move {
-            // This is a barrier only. Readers that were already admitted have
-            // no further await between their final checks and enqueue.
-            {
-                let _admission = admission.write().await;
-            }
-            for handle in shutdown_handles {
-                handle.signal();
-            }
-        }));
-    }
-}
-
 /// Number of bits reserved for the stream index.
 /// 6 bits supports up to 64 sub-streams.
 const STREAM_BITS: u32 = 6;
+pub(crate) const MAX_STREAMS: usize = 1 << STREAM_BITS;
 const OFFSET_MASK: i64 = (1i64 << (64 - STREAM_BITS)) - 1;
 
 /// Opaque identifier returned by ingest methods on MultiplexedStream.
@@ -91,6 +45,8 @@ const OFFSET_MASK: i64 = (1i64 << (64 - STREAM_BITS)) - 1;
 ///
 /// Unlike a `ZerobusStream` offset, `MessageId` values are not ordered — pass
 /// them to [`MultiplexedStream::wait_for_message_id`] to await acknowledgment.
+/// A message ID is meaningful only to the mux that produced it; using it with
+/// another mux is caller error and is not detected at runtime.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct MessageId(i64);
 
@@ -107,7 +63,7 @@ impl std::fmt::Display for MessageId {
 
 impl MessageId {
     pub(crate) fn new(stream_index: usize, sub_offset: OffsetId) -> Self {
-        debug_assert!(stream_index < (1 << STREAM_BITS));
+        debug_assert!(stream_index < MAX_STREAMS);
         debug_assert!((0..=OFFSET_MASK).contains(&sub_offset));
         Self(((stream_index as i64) << (64 - STREAM_BITS)) | (sub_offset & OFFSET_MASK))
     }
@@ -164,9 +120,9 @@ pub(crate) fn multiplexed_ack_callback(
     callback: Arc<dyn AckCallback<MessageId>>,
 ) -> Arc<dyn AckCallback> {
     assert!(
-        stream_index < (1 << STREAM_BITS),
+        stream_index < MAX_STREAMS,
         "MultiplexedStream supports at most {} sub-streams",
-        1 << STREAM_BITS
+        MAX_STREAMS
     );
     Arc::new(MultiplexedAckCallbackAdapter {
         stream_index,
@@ -183,44 +139,43 @@ pub struct MultiplexedStream {
     round_robin_counter: AtomicUsize,
     is_closed: AtomicBool,
     closed_token: CancellationToken,
-    failure_context: OnceLock<String>,
+    failure: OnceLock<ZerobusError>,
     admission: Arc<tokio::sync::RwLock<()>>,
 }
 
 impl MultiplexedStream {
-    /// Creates a multiplexed stream over the given sub-streams.
-    ///
-    /// Ingest waits up to 30 seconds for capacity on its selected sub-stream.
-    ///
-    /// # Panics
-    ///
-    /// Panics if `streams` is empty or holds more than 64 sub-streams.
     pub fn new(streams: Vec<ZerobusStream>) -> Self {
         assert!(
             !streams.is_empty(),
             "MultiplexedStream requires at least one sub-stream"
         );
         assert!(
-            streams.len() <= (1 << STREAM_BITS),
+            streams.len() <= MAX_STREAMS,
             "MultiplexedStream supports at most {} sub-streams",
-            1 << STREAM_BITS
+            MAX_STREAMS
         );
         Self {
             streams,
             round_robin_counter: AtomicUsize::new(0),
             is_closed: AtomicBool::new(false),
             closed_token: CancellationToken::new(),
-            failure_context: OnceLock::new(),
+            failure: OnceLock::new(),
             admission: Arc::new(tokio::sync::RwLock::new(())),
         }
     }
 
     #[allow(clippy::result_large_err)]
     fn check_closed(&self) -> ZerobusResult<()> {
+        if let Some(error) = self.failure_error() {
+            return Err(error);
+        }
         if self.is_closed_fast() {
             return Err(ZerobusError::InvalidStateError(
                 "MultiplexedStream is closed".to_string(),
             ));
+        }
+        if let Some(idx) = self.first_closed_lane() {
+            return Err(self.lane_error(idx));
         }
         Ok(())
     }
@@ -229,13 +184,39 @@ impl MultiplexedStream {
         self.is_closed.load(Ordering::Relaxed)
     }
 
+    fn first_closed_lane(&self) -> Option<usize> {
+        self.streams.iter().position(ZerobusStream::is_closed)
+    }
+
+    fn failure_error(&self) -> Option<ZerobusError> {
+        self.failure.get().cloned()
+    }
+
+    fn lane_error(&self, idx: usize) -> ZerobusError {
+        self.streams[idx].terminal_error().unwrap_or_else(|| {
+            ZerobusError::InvalidStateError(format!(
+                "MultiplexedStream sub-stream {idx} closed unexpectedly"
+            ))
+        })
+    }
+
+    async fn ensure_open(&self) -> ZerobusResult<()> {
+        if self.failure.get().is_some() || self.is_closed_fast() {
+            return self.check_closed();
+        }
+        if let Some(idx) = self.first_closed_lane() {
+            let error = self.lane_error(idx);
+            self.shutdown_on_failure(idx, &error).await;
+            return Err(error);
+        }
+        Ok(())
+    }
+
     async fn shutdown_on_failure(&self, trigger_index: usize, cause: &ZerobusError) {
-        if self.is_closed.swap(true, Ordering::Relaxed) {
+        if self.failure.set(cause.clone()).is_err() {
             return;
         }
-        let _ = self
-            .failure_context
-            .set(format!("sub-stream {trigger_index} failed: {cause}"));
+        self.is_closed.store(true, Ordering::Release);
         self.closed_token.cancel();
 
         error!(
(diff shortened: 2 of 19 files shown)


rust/sdk/src/stream/grpc/acks.rs
```diff
@@ -51,19 +51,18 @@ impl ZerobusStream {
                         "Stream is not caught up to any offset yet. Waiting for the first offset."
                     );
                 }
-                if self.is_closed.load(Ordering::Relaxed) {
```
danilonajkov-db: acquire here is better as it guarantees that concurrent readers read the last value of terminal_error. Relaxed can only provide atomicity

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -211,32 +163,53 @@ impl MultiplexedStream {
...
     async fn shutdown_on_failure(&self, trigger_index: usize, cause: &ZerobusError) {
-        if self.is_closed.swap(true, Ordering::Relaxed) {
+        if self.is_closed_fast() || self.failure.set(cause.clone()).is_err() {
             return;
         }
-        let _ = self
-            .failure_context
-            .set(format!("sub-stream {trigger_index} failed: {cause}"));
+        self.is_closed.store(true, Ordering::Relaxed);
         self.closed_token.cancel();
```
teodordelibasic-db, line 205: This cancel only runs if some mux call already saw the dead lane (`ensure_open`, ingest error, flush, or a wait on that lane). A sibling can fail in the background with none of those. Then [`is_closed()`](https://github.com/databricks/zerobus-sdk/blob/bad47265e8fbf19b6b78fe16edaf69227750c854/rust/sdk/src/multiplexed_stream.rs#L523) is already true, but a `wait_for_message_id` or capacity wait on a healthy lane stays blocked until its own ack or the five-minute flush timeout.

The wakeup test waits on the failed id first ([`multiplexed_stream_tests.rs:734`](https://github.com/databricks/zerobus-sdk/blob/bad47265e8fbf19b6b78fe16edaf69227750c854/rust/tests/src/multiplexed_stream_tests.rs#L734)), and that call is what wakes the other wait. So the changelog line about waking pending waits after a lane fails is not true for a background death.

Could we select on each lane's terminal token from `wait_for_message_id` / `reserve_capacity` and call `shutdown_on_failure` as soon as any lane is terminal?

```rust
_ = self.streams[i].terminal_cancelled() => {
    self.shutdown_on_failure(i, &self.streams[i].last_error().unwrap_or_else(|| {
        ZerobusError::InvalidStateError(format!(
            "MultiplexedStream sub-stream {i} is closed",
        ))
    })).await;
    return Err(self.closed_error());
}
```

Could we add a test that waits only on the healthy id, waits until `mux.is_closed()` is true with no wait or flush on the failed id, and asserts the healthy wait finishes well under the flush timeout?

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -211,32 +163,53 @@ impl MultiplexedStream {
...
     fn is_closed_fast(&self) -> bool {
         self.is_closed.load(Ordering::Relaxed)
     }
 
+    fn first_closed_stream(&self) -> Option<usize> {
+        self.streams.iter().position(ZerobusStream::is_closed)
+    }
+
+    async fn ensure_open(&self) -> ZerobusResult<()> {
+        self.check_closed()?;
```
teodordelibasic-db: `ensure_open` poisons with `InvalidStateError("MultiplexedStream sub-stream {idx} is closed")` instead of the lane's `StreamClosedError`. `OnceLock` keeps that first value, so we never surface `PermissionDenied`. `InvalidStateError` is never retryable, and `is_auth_rejection()` will not match, so we also skip token invalidation. The tests keep `PermissionDenied` only because they wait or flush the failed lane first ([`multiplexed_stream_tests.rs:749`](https://github.com/databricks/zerobus-sdk/blob/bad47265e8fbf19b6b78fe16edaf69227750c854/rust/tests/src/multiplexed_stream_tests.rs#L749)).

Could we clone the lane's terminal error before `shutdown_on_failure`? If `close()` already ran, leave the stored cause alone.

```rust
if let Some(idx) = self.first_closed_stream() {
    let error = self.streams[idx].last_error().unwrap_or_else(|| {
        ZerobusError::InvalidStateError(format!(
            "MultiplexedStream sub-stream {idx} is closed",
        ))
    });
    self.shutdown_on_failure(idx, &error).await;
    return Err(self.closed_error());
}
```

Could we extend `test_public_is_closed_reports_closed_substream` so that after `mux.is_closed()` becomes true with no wait on that lane, the next `ingest_record` is `StreamClosedError` with `PermissionDenied`, not `InvalidStateError`?

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -490,10 +439,12 @@ impl MultiplexedStream {
...
             )));
         }
-        match self.streams[idx]
-            .wait_for_offset(message_id.sub_offset())
-            .await
-        {
+        let wait_result = tokio::select! {
+            biased;
+            result = self.streams[idx].wait_for_offset(message_id.sub_offset()) => result,
+            _ = self.closed_token.cancelled() => return Err(self.closed_error()),
```
teodordelibasic-db: This arm fails the wait without checking whether this id was already acked. Lane `wait_for_offset` re-reads the watermark before failing on close ([`acks.rs:54`](https://github.com/databricks/zerobus-sdk/blob/bad47265e8fbf19b6b78fe16edaf69227750c854/rust/sdk/src/stream/grpc/acks.rs#L54)); we do not. The new test then releases the healthy ack only after that error ([`multiplexed_stream_tests.rs:739`](https://github.com/databricks/zerobus-sdk/blob/bad47265e8fbf19b6b78fe16edaf69227750c854/rust/tests/src/multiplexed_stream_tests.rs#L739)).

`close()` has the same issue: it cancels the token ([`multiplexed_stream.rs:471`](https://github.com/databricks/zerobus-sdk/blob/bad47265e8fbf19b6b78fe16edaf69227750c854/rust/sdk/src/multiplexed_stream.rs#L471)) before the parallel flush, so in-flight waits fail while close is still waiting for those acks. A caller who retries that `MessageId` can duplicate a durable record. The error is also the sibling's `StreamClosedError`, so retry and auth checks describe the wrong record.

Could we keep waiting on this lane's `wait_for_offset` unless this lane is terminal?

```rust
_ = self.closed_token.cancelled() => {
    if !self.streams[idx].is_closed() {
        return self.streams[idx]
            .wait_for_offset(message_id.sub_offset())
            .await;
    }
    return Err(self.closed_error());
}
```

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -446,14 +397,17 @@ impl MultiplexedStream {
     /// state, the mux is poisoned. The first flush error is returned;
     /// additional ones are logged.
     pub async fn flush(&self) -> ZerobusResult<()> {
-        self.check_closed()?;
-        let results = join_all(self.streams.iter().map(|s| s.flush())).await;
+        self.ensure_open().await?;
```
teodordelibasic-db: `ensure_open` returns before any lane flush if a lane is already closed. If a lane dies mid-loop, we return on the first terminal error and drop the rest ([`multiplexed_stream.rs:408`](https://github.com/databricks/zerobus-sdk/blob/bad47265e8fbf19b6b78fe16edaf69227750c854/rust/sdk/src/multiplexed_stream.rs#L408)). A second `flush` hits `check_closed` and skips drain again.

The docs still say flush waits until every queued record is acked ([`multiplexed_stream.rs:393`](https://github.com/databricks/zerobus-sdk/blob/bad47265e8fbf19b6b78fe16edaf69227750c854/rust/sdk/src/multiplexed_stream.rs#L393)). `MessageId`s are unordered, so flush is the only all-lanes durability barrier. The new test locks in not waiting for a gated healthy ack ([`multiplexed_stream_tests.rs:806`](https://github.com/databricks/zerobus-sdk/blob/bad47265e8fbf19b6b78fe16edaf69227750c854/rust/tests/src/multiplexed_stream_tests.rs#L806)). A caller who treats `Err` as "these records failed" and retries will duplicate.

Could we keep ingest rejected after poison, but still wait on healthy lanes with the existing flush timeout?

```rust
for (i, stream) in self.streams.iter().enumerate() {
    if stream.is_closed() {
        continue;
    }
    flushes.push(async move { (i, stream.flush().await) });
}
// collect healthy results, then shutdown_on_failure if any lane is terminal
```

rust/sdk/src/multiplexed_stream.rs
teodordelibasic-db, line 526: A background lane close makes this true without storing `failure` or cancelling `closed_token`. The new test treats that as desired ([`multiplexed_stream_tests.rs:1245`](https://github.com/databricks/zerobus-sdk/blob/bad47265e8fbf19b6b78fe16edaf69227750c854/rust/tests/src/multiplexed_stream_tests.rs#L1245)).

[`get_unacked_records`](https://github.com/databricks/zerobus-sdk/blob/bad47265e8fbf19b6b78fe16edaf69227750c854/rust/sdk/src/multiplexed_stream.rs#L532) always `close()`s first, so `if mux.is_closed() { mux.get_unacked_records().await }` tears down healthy lanes that are still ingesting. Ingest can still succeed until something calls `ensure_open`.

Could `is_closed()` mean mux-terminal (poison or `close()`), and a separate `has_failed_lane()` cover the eager scan?

```rust
pub fn is_closed(&self) -> bool {
    self.is_closed_fast()
}

pub fn has_failed_lane(&self) -> bool {
    self.streams.iter().any(ZerobusStream::is_closed)
}
```

teodordelibasic-db: This comment says the per-stream `close` flush is cheap because we already drained. That is wrong when the first flush times out: [`ZerobusStream::close`](https://github.com/databricks/zerobus-sdk/blob/bad47265e8fbf19b6b78fe16edaf69227750c854/rust/sdk/src/stream/grpc/close.rs#L97) flushes again and can wait the same `flush_timeout_ms` a second time. Lanes already close in parallel, so the extra mux flush is redundant. `get_unacked_records` goes through this; default timeout is 300s, so recovery can take about ten minutes.

Could we run only the lane `close` futures in parallel?

```rust
self.is_closed.store(true, Ordering::Relaxed);
self.closed_token.cancel();

let close_results = join_all(
    self.streams
        .iter_mut()
        .map(|stream| stream.close()),
)
.await;
```

rust/NEXT_CHANGELOG.md
```diff
@@ -23,6 +23,9 @@
 
 ### Internal Changes
 
+- Updated multiplexed-stream failure handling to reject new ingestion and wake
```
teodordelibasic-db: This line says we reject new ingest, wake pending waits, and close lanes concurrently. It does not say that `flush` / `wait_for_message_id` can return a sibling failure while other lanes keep running, or that poison no longer flushes or shuts those lanes down. With `connection_per_stream` true, those lanes keep TCP connections and server stream tokens until `close` or `Drop`.

Mux is still `feature = "testing"`, so this is not a public break yet, but anyone on the testing API gets a silent contract change. Could we describe the actual poison / flush / wait / close behavior here?

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -246,32 +219,12 @@ impl MultiplexedStream {
...
-                .iter()
-                .map(ZerobusStream::shutdown_handle)
-                .collect(),
-        );
-
         // Drain any readers already admitted before `is_closed` was set. The
         // write lock is only a barrier: readers arriving after it is released
         // will observe the closed state and reject the ingest.
         {
             let _admission = self.admission.write().await;
```
irinatomic-db: Now that `shutdown_on_failure` doesn't tear down the sub-streams anymore (no more cleanup guard / flush / signal), is this `admission.write()` barrier still doing anything? Readers already re-check `check_closed()` under the read lock.

rust/NEXT_CHANGELOG.md
irinatomic-db: `MultiplexedStream` is publicly re-exported, and after this change ingest/flush return the sub-stream's real error (`StreamClosedError`) instead of `InvalidStateError` - the test updates show that. 

If a client was matching on that we should probably add it to the CHANGELOG.md

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -12,76 +12,28 @@
 //! [`wait_for_message_id`](MultiplexedStream::wait_for_message_id) without
 //! needing to know which sub-stream handled the record.
 //!
-//! The mux is poisoned on the first unrecoverable sub-stream error: remaining
-//! sub-streams are flushed and further ingest calls fail. Any records still
-//! buffered can be recovered via
+//! The mux stops accepting new records after the first unrecoverable
+//! sub-stream error. Healthy sub-streams remain alive until the mux is closed
```
irinatomic-db: Maybe we should add to the doc that after a poison the healthy lanes keep their gRPC/supervisor tasks running until an explicit `close()` (or drop).

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -211,32 +163,53 @@ impl MultiplexedStream {
...
+
     fn is_closed_fast(&self) -> bool {
         self.is_closed.load(Ordering::Relaxed)
     }
 
+    fn first_closed_stream(&self) -> Option<usize> {
+        self.streams.iter().position(ZerobusStream::is_closed)
+    }
+
+    async fn ensure_open(&self) -> ZerobusResult<()> {
```
irinatomic-db: `ensure_open` scans every sub-stream via `first_closed_stream`, and `ingest_record`/`ingest_records` end up calling it twice per record (directly + again through `reserve_capacity`). 

The old `check_closed` was a single atomic load. Probably minor since each `is_closed` is cheap, but on the hot path with up to 64 lanes it's ~2×N loads per record - worth scanning once, or leaning on` handle_ingest_error` to catch a dead lane lazily?

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: I'm not sure if we want to immediately cancel all wait_for_message_ids. That stream is healthy and an ack will be returned (most likely). This would just probably increase the duplicate rate as clients will then resend the record. WDYT?

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: makes sense, thanks

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: true, will fix

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: yep, ill add a check for all streams

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: yep makes sense to have is_closed() report mux-terminal state only.

danilonajkov-db: That is correct, but I kept the two phases intentionally. The initial flush captures an already-failed steams’s error, which its idempotent close() could otherwise hide. Avoiding the second flush safely would require a separate internal close-after-flush operation.

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: true i can remove the RwLock and write barrier

rust/NEXT_CHANGELOG.md
danilonajkov-db: MultiplexedStream remains behind the testing feature and is not yet a stable released API, so this is recorded under Internal Changes rather than as a breaking change.

rust/sdk/src/multiplexed_stream.rs
teodordelibasic-db: `join_all` now waits for every lane when flush itself discovers the failure. This `ensure_open` still skips that work if a earlier mux call already poisoned, or if a lane is already closed. Then healthy records stay in flight, and a caller who retries them can duplicate.

The new test only covers discovery during flush ([`multiplexed_stream_tests.rs:809`](https://github.com/databricks/zerobus-sdk/blob/f771e4c3981899abc49df034fd87d253fae150f2/rust/tests/src/multiplexed_stream_tests.rs#L809)). Could we skip `ensure_open` here, always flush live lanes, then poison? The changelog already describes that all-lane barrier.

```rust
pub async fn flush(&self) -> ZerobusResult<()> {
    let results = join_all(self.streams.iter().map(ZerobusStream::flush)).await;
    // keep the existing first_terminal / first_error handling
}
```

Could we add a test that poisons via `wait_for_message_id` on the failed id (healthy ack still gated), then asserts `flush` stays pending until that ack?

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -385,14 +326,14 @@ impl MultiplexedStream {
...
-    async fn handle_ingest_error(
+    fn handle_ingest_error(
         &self,
         e: ZerobusError,
         stream: &ZerobusStream,
         idx: usize,
     ) -> ZerobusError {
         if stream.is_closed() {
-            self.shutdown_on_failure(idx, &e).await;
+            self.shutdown_on_failure(idx, &e);
```
teodordelibasic-db: `ensure_open` now clones `terminal_error()`, but this path stores whatever the lane returned. If the lane closed during reserve or enqueue, that is often `StreamClosedError(Internal("Stream closed"))` from [`check_open`](https://github.com/databricks/zerobus-sdk/blob/f771e4c3981899abc49df034fd87d253fae150f2/rust/sdk/src/stream/grpc/ingest.rs#L185). `OnceLock` keeps it, so later ingest loses `PermissionDenied` and the retry/auth bits.

The capacity-wait test accepts any `StreamClosedError` and never checks the next ingest. Could we use `terminal_error()` here too, and return that?

```rust
if stream.is_closed() {
    let cause = stream.terminal_error().await.unwrap_or(e);
    self.shutdown_on_failure(idx, &cause);
    return cause;
}
```

Could we extend `test_concurrent_ingest_waiting_for_capacity_fails_after_mux_poison` so the next `ingest_record` is `StreamClosedError` with `PermissionDenied`?

rust/sdk/src/multiplexed_stream.rs
teodordelibasic-db: `join_all` now waits for every lane when flush itself discovers the failure. This `ensure_open` still skips that work if an earlier mux call already poisoned, or if a lane is already closed. Then healthy records stay in flight, and a caller who retries them can duplicate.

The new test only covers discovery during flush ([`multiplexed_stream_tests.rs:809`](https://github.com/databricks/zerobus-sdk/blob/f771e4c3981899abc49df034fd87d253fae150f2/rust/tests/src/multiplexed_stream_tests.rs#L809)). Could we skip `ensure_open` here, always flush live lanes, then poison? That matches the all-lane barrier already in the changelog.

```rust
pub async fn flush(&self) -> ZerobusResult<()> {
    let results = join_all(self.streams.iter().map(ZerobusStream::flush)).await;
    // keep the existing first_terminal / first_error handling
}
```

Could we add a test that poisons via `wait_for_message_id` on the failed id (healthy ack still gated), then asserts `flush` stays pending until that ack?

rust/sdk/src/multiplexed_stream.rs
teodordelibasic-db: `ensure_open` now clones `terminal_error()`, but this path stores whatever the lane returned. If the lane closed during reserve or enqueue, that is often `StreamClosedError(Internal("Stream closed"))` from [`check_open`](https://github.com/databricks/zerobus-sdk/blob/f771e4c3981899abc49df034fd87d253fae150f2/rust/sdk/src/stream/grpc/ingest.rs#L185). `OnceLock` keeps it, so later ingest loses `PermissionDenied` and the retry/auth bits.

The capacity-wait test accepts any `StreamClosedError` and never checks the next ingest. Could we use `terminal_error()` here too, and return that? Same typed-error rule as lazy `ensure_open`.

```rust
if stream.is_closed() {
    let cause = stream.terminal_error().await.unwrap_or(e);
    self.shutdown_on_failure(idx, &cause);
    return cause;
}
```

Could we extend `test_concurrent_ingest_waiting_for_capacity_fails_after_mux_poison` so the next `ingest_record` is `StreamClosedError` with `PermissionDenied`?

rust/tests/src/multiplexed_stream_tests.rs
```diff
@@ -660,6 +660,8 @@ mod multi_stream_tests {
 
         mux.close().await?;
         assert!(mux.is_closed());
+        // A ready acknowledgment wins even though close also cancelled the mux token.
```
teodordelibasic-db: This comment is leftover from when `wait_for_message_id` selected on `closed_token`. The wait is lane-scoped now, which is the agreed behavior. Could we update the comment so it does not say close cancelled the mux token?

rust/sdk/src/multiplexed_stream.rs
```diff
@@ -426,7 +367,7 @@ impl MultiplexedStream {
         I: IntoIterator<Item = T>,
         T: Into<PreparedInput>,
     {
-        self.check_closed()?;
+        self.ensure_open().await?;
```
teodordelibasic-db: `ensure_open` runs before the empty check, so an empty batch can return a typed lane error instead of `Ok(None)`. The lazy-error test uses that as a probe, which fits lazy observation. Could we say so on the method docs (it currently says it returns `None` for an empty batch)? Not asking to skip `ensure_open` for empty payloads.

rust/sdk/src/multiplexed_stream.rs
teodordelibasic-db: It can happen that an error is returned due to `check_closed` being called first and an already poisoned or closed mutex. We can just update the comment slighly.

rust/sdk/src/stream/grpc/acks.rs
teodordelibasic-db: This comment still says unacked landing-zone records can come from `signal_shutdown` on a poisoned mux. Poison does not shut healthy lanes; only `Drop` calls `signal_shutdown`. Could we update the comment to match?

rust/sdk/src/multiplexed_stream.rs
teodordelibasic-db: Makes sense, agreed, we can leave as is.

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: Updated since my earlier reply, on second thought having an internal method close-after-flush is actually useful. It tears down everything without actually flushing. Mux close now flushes all lanes once, then finalizes them concurrently without flushing again.

rust/sdk/src/multiplexed_stream.rs
danilonajkov-db: Fixed, including the already-poisoned case. Flush waits for every lane’s flush attempt before returning the error. Added a regression test with that exact case

rust/sdk/src/stream/grpc/close.rs
teodordelibasic-db: Maybe we can be a bit more detailed in these comments regarding releasing resources, since on line 31 we say the method does that, but that is not a guarantee, even for `Ok()`.  Maybe something like this:

```rust
/// `Ok(())` if flush succeeded. Teardown then waits up to one second, plus
/// 100ms after abort. A blocked synchronous credentials callback can outlive
/// that budget; `Ok` does not mean that task has exited.
```

rust/NEXT_CHANGELOG.md
```diff
@@ -19,10 +19,21 @@
 
 ### Bug Fixes
 
+- Bounded the post-abort supervisor wait during gRPC stream close, including
+  when a synchronous credentials callback is blocked.
 ### Documentation
```
teodordelibasic-db, line 25: Blank line above.

rust/sdk/src/stream/grpc/acks.rs
teodordelibasic-db: This comment says the supervisor sends the real error before `is_closed = true`. [`fail_stream`](https://github.com/databricks/zerobus-sdk/blob/aca2cee8bc31c6619294c1fa90c534da3b6958d7/rust/sdk/src/stream/grpc/supervisor.rs#L347) stores `is_closed` first, then the watch value, then cancels `terminal_token`. Mux `terminal_error` waits on that token, so it is safe. Could we update the comment to match?

```rust
// fail_stream sets is_closed, then server_error_rx, then cancels
// terminal_token. This path can still see is_closed before the final
// watch value.
```

teodordelibasic-db (dismissed):
LGTM, just a couple of doc nits.

teodordelibasic-db (approved):
LGTM

PR #827, opened 2026-09-03
[Rust] Add Avro record format (avro feature)

irinatomic-db:
## PR stack

    - [avro/rust-core](https://github.com/databricks/zerobus-sdk/pull/827) <- _this PR_
        - avro/java
        - avro/typescript

Adds the Avro record format to the Rust SDK, behind the off-by-default `avro` feature (Beta). Ephemeral streams only; server support is pending. Default builds are unchanged.

Declare the writer schema at stream creation with `.avro(schema_json)`, then ingest either a **record object** the stream encodes against the schema (`AvroRecord(AvroValue)`) or a **pre-encoded datum** (`AvroBytes`) — both via the existing `ingest_record_offset` / `ingest_records_offset`. `AvroValue` is `apache_avro::types::Value` (re-exported); it can represent every Avro type, so the record path is exhaustive.

Implementation notes:
- `apache-avro` is an optional dep, pulled only by the `avro` feature. It requires Rust 1.85; default builds are unaffected.
- Ingest accepts `impl Into<PreparedInput>` (a `#[doc(hidden)]` payload type). A blanket `From<T: Into<EncodedRecord>>` keeps every existing caller unchanged; `AvroRecord` carries an un-encoded `AvroValue`.
- The writer schema is parsed once at `build()` and stored in `TableProperties` next to the JSON wire form, so it's reused on ingest and survives stream recovery.
- CI compiles, lints, and tests the feature via `--features avro` steps in `ci-rust.yml` (the default `--workspace` / `make lint` runs don't touch it).

## How is this tested?

```
cd rust

# With the feature (unit + doctests incl. AvroRecord/AvroBytes cases)
cargo test  -p databricks-zerobus-ingest-sdk --features avro
cargo clippy -p databricks-zerobus-ingest-sdk --features avro --all-targets

# Default (feature off) — no regressions
cargo test -p databricks-zerobus-ingest-sdk

# Whole workspace builds/tests (ffi/jni not broken)
cargo build --workspace
cargo test  --workspace --no-run

cargo fmt --all --check

# Excluded example builds standalone
cd examples/avro && cargo build --examples
```


rust/.gitignore
```diff
@@ -1,2 +1,4 @@
 target/
 .cargo/
+# examples/avro is a non-workspace crate; its lockfile is not tracked.
+examples/avro/Cargo.lock
```
danilonajkov-db, line 5: weird that we need this

rust/examples/avro/src/main.rs
```diff
@@ -0,0 +1,59 @@
+//! Avro ingestion (Beta) with pre-encoded `AvroBytes` records.
```
danilonajkov-db: maybe add another example or in this same one how ingest looks like for AvroRecord instead of just bytes

danilonajkov-db (dismissed):

rust/.gitignore
elenagaljak-db: +1 on the weird

rust/Cargo.toml
```diff
@@ -9,6 +9,9 @@ members = [
     "tools/*",
     "tests",
 ]
+# Excluded: as a member it would force the sdk `avro` feature across the workspace
+# and break ffi/jni. Build it standalone with `cargo build` inside examples/avro.
+exclude = ["examples/avro"]
 resolver = "2"
```
elenagaljak-db: How do we resolve this problem for other features?

rust/sdk/src/builder/stream_builder.rs
```diff
@@ -54,6 +54,9 @@ enum FormatConfig {
     DynamicProto(MessageDescriptor),
     #[cfg(feature = "arrow-flight")]
     Arrow(Arc<ArrowSchema>),
+    /// Avro writer schema as JSON.
```
elenagaljak-db: we don't need this comment

rust/sdk/src/builder/stream_builder.rs
```diff
@@ -220,6 +225,15 @@ impl<'a> StreamBuilder<'a> {
         self
     }
 
+    /// Select Avro record format with the writer schema as JSON (Beta).
+    ///
+    /// Ingest pre-encoded datums with [`AvroBytes`](crate::AvroBytes).
```
elenagaljak-db: Myb we could move this "Ingest pre-encoded..." comment somewhere else?

rust/sdk/src/stream/grpc/connection.rs
```diff
@@ -36,6 +36,7 @@ impl ZerobusStream {
         table_properties: &TableProperties,
         headers_provider: &Arc<dyn HeadersProvider>,
         record_type: RecordType,
+        avro_schema_json: Option<String>,
```
elenagaljak-db: I'm guessing that we had to do this here because adding avro schema to table properties would be a breaking change?

rust/sdk/src/stream_configuration.rs
```diff
@@ -96,6 +96,12 @@ pub struct StreamConfigurationOptions {
     /// Default: RecordType::Proto
     pub record_type: RecordType,
 
+    /// Avro writer schema (JSON) for `RecordType::Avro` streams; set via
+    /// [`StreamBuilder::avro`](crate::StreamBuilder::avro).
+    ///
+    /// Default: `None`
+    pub avro_schema_json: Option<String>,
```
elenagaljak-db: I don't think this belongs in stream config options. That struct should contain configs that every stream has. I think its more suited in TableProperties

rust/sdk/src/stream/grpc/connection.rs
teodordelibasic-db: I don't think it would be, just checked and `TableProperties` is `pub(crate)`. Agreed that it feels off having the schema as additional arg here.

rust/sdk/src/stream/grpc/supervisor.rs
```diff
@@ -93,6 +93,7 @@ impl ZerobusStream {
                 let table_properties = table_properties.clone();
                 let headers_provider = Arc::clone(&headers_provider);
                 let record_type = options.record_type;
+                let avro_schema_json = options.avro_schema_json.clone();
```
elenagaljak-db: Ditto

rust/sdk/src/stream_configuration.rs
teodordelibasic-db: +1

rust/examples/avro/README.md
```diff
@@ -0,0 +1,13 @@
+# Avro ingestion example (Beta)
```
elenagaljak-db: We should try to keep the same style and format of the other examples READMEs

rust/sdk/src/stream/grpc/ingest.rs
```diff
@@ -50,7 +50,7 @@ impl ZerobusStream {
     /// ```
     pub async fn ingest_record_offset(
         &self,
-        payload: impl Into<EncodedRecord>,
+        payload: impl IntoEncodedRecord,
```
teodordelibasic-db: Eh I am not a fan of this, I see why we may need it, but I'll try to see if there is an alternative.

rust/sdk/src/stream/grpc/mod.rs
```diff
@@ -136,6 +139,9 @@ pub struct ZerobusStream {
     /// Resolved message descriptor for building dynamic-proto records, supplied by
     /// the builder. `None` for JSON and compiled-proto streams.
     dynamic_message_descriptor: Option<MessageDescriptor>,
+    /// Cached Avro schema, lazily parsed from avro_schema_json on first use.
+    #[cfg(feature = "avro")]
+    avro_schema_cache: OnceLock<Result<apache_avro::Schema, ZerobusError>>,
```
teodordelibasic-db: Why don't we parse the schema when building the stream and avoid having to use the `OnceLock`?

rust/.gitignore
irinatomic-db: It's a side effect of `examples/avro` being temporarily excluded from the workspace: the excluded crate builds its own lockfile that we don't track. It goes away once avro is ungated and the crate rejoins `members` (the root lock takes over). Added a comment explaining it.

rust/sdk/src/builder/stream_builder.rs
irinatomic-db: Removed

rust/sdk/src/builder/stream_builder.rs
irinatomic-db: Reworded the `.avro()` doc to cover both ingest paths (`AvroRecord` for objects, `AvroBytes` for pre-encoded) instead of the pre-encoded-only line.

rust/sdk/src/stream/grpc/connection.rs
irinatomic-db: Moved the schema into `TableProperties` and dropped the extra arg - it's read from `table_properties.avro_schema_json` now.

rust/sdk/src/stream/grpc/ingest.rs
```diff
@@ -110,9 +110,14 @@ impl ZerobusStream {
...
-        let encoded_batch = EncodedBatch::try_from_record(payload, self.options.record_type)
+        let schema = self.get_avro_schema()?;
+        let encoded_record = payload.into_encoded(
+            self.options.record_type,
+            schema.as_ref().map(|s| s as &dyn std::any::Any),
+        )?;
+        let encoded_batch = EncodedBatch::try_from_record(encoded_record, self.options.record_type)
             .ok_or_else(|| {
                 ZerobusError::InvalidArgument(
                     "Record type does not match stream configuration".to_string(),
```
elenagaljak-db: I agree with Teodor on the above, and also I'm not a huge fan of having a separate handling of avro here.

rust/sdk/src/stream/grpc/supervisor.rs
irinatomic-db: Fixed. Moved the `avro_schema_json` to `TableProperties`.

rust/sdk/src/stream_configuration.rs
irinatomic-db: Fixed. Moved the `avro_schema_json` to `TableProperties`.

rust/sdk/Cargo.toml
```diff
@@ -41,6 +41,7 @@ hyper-util.workspace = true
 smallvec.workspace = true
 bytes.workspace = true
 self_cell = { version = "1.2", optional = true }
+apache-avro = { version = "0.20", optional = true }
```
teodordelibasic-db, line 44: A slight problem here is that this `apache-avro` creates requires Rust 1.85 and the minimum advertised version for this crate is 1.70. Could we document this?

```toml
# Avro record format (Beta). apache-avro 0.20 requires Rust 1.85.
# Default builds keep rust-version 1.70.
avro = ["dep:apache-avro"]
```

rust/sdk/src/lib.rs
```diff
@@ -124,6 +126,9 @@ pub(crate) struct TableProperties {
     pub(crate) table_name: String,
     pub(crate) descriptor_proto: Option<prost_types::DescriptorProto>,
     pub(crate) message_descriptor: Option<MessageDescriptor>,
+    /// Avro writer schema (JSON) for `RecordType::Avro` streams; set via
+    /// [`StreamBuilder::avro`](crate::StreamBuilder::avro). `None` otherwise.
+    pub(crate) avro_schema_json: Option<String>,
 }
```
elenagaljak-db: Shouldn't this be under the flag too? I'm not sure how proper feature gating works in rust tho, so it maybe shouldn't. Also we don't need the comment

rust/sdk/src/record_types.rs
```diff
@@ -157,10 +179,96 @@ impl<T: serde::Serialize> From<JsonValue<T>> for EncodedRecord {
...
+/// ```
+#[cfg(feature = "avro")]
+pub struct AvroRecord<T: serde::Serialize>(pub T);
+
+/// Sealed trait for converting records into [`EncodedRecord`] with optional schema context.
+///
+/// This trait enables schema-aware encoding for Avro records while maintaining
+/// backward compatibility with existing record types through a blanket implementation.
+/// It is intentionally sealed to prevent external implementations.
+pub trait IntoEncodedRecord: Sized {
```
teodordelibasic-db: PTAL at this to see if it looks better maybe and if it is not a breaking change. It changes a public method signature, but it may be okay. We can discuss.

This change also made [`prepare_records`](https://github.com/databricks/zerobus-sdk/blob/9df57f669c2697cb7028c5c7ef83e06d9811796b/rust/sdk/src/stream/grpc/ingest.rs#L136) collect `Vec<EncodedRecord>` and then walk it again into `try_from_batch`. That used to be one pass into a `SmallVec`. It now runs for every `ingest_records_offset`, including default JSON/proto. The "collecting errors" comment is also wrong: `?` still returns on the first error.

Could we convert with no schema into a hidden `PreparedInput`, then fetch and apply the schema only when the payload is actually `AvroRecord`, and fold that fallible conversion directly into the `SmallVec`? `JsonValue` / `ProtoMessage` / `AvroBytes` stay `Ready`. A blanket `From` keeps `ingest_record_offset(JsonValue(...))` compiling. `PreparedInput` has to be `pub` (re-exported) so it can appear on the public ingest signature; `#[doc(hidden)]` keeps it out of rustdoc.

In [`record_types.rs`](https://github.com/databricks/zerobus-sdk/blob/9df57f669c2697cb7028c5c7ef83e06d9811796b/rust/sdk/src/record_types.rs#L217), drop the trait and the `Any` impls:

```rust
#[doc(hidden)]
pub enum PreparedInput {
    Ready(EncodedRecord),
    #[cfg(feature = "avro")]
    AvroObject(AvroPending),
}

#[cfg(feature = "avro")]
pub struct AvroPending {
    encode_fn: Box<dyn FnOnce(&apache_avro::Schema) -> crate::ZerobusResult<Vec<u8>> + Send>,
}

impl<T: Into<EncodedRecord>> From<T> for PreparedInput {
    fn from(value: T) -> Self {
        PreparedInput::Ready(value.into())
    }
}

#[cfg(feature = "avro")]
impl<T: serde::Serialize + Send + 'static> From<AvroRecord<T>> for PreparedInput {
    fn from(record: AvroRecord<T>) -> Self {
        PreparedInput::AvroObject(AvroPending {
            encode_fn: Box::new(move |schema| {
                apache_avro::to_value(&record.0)
                    .and_then(|v| v.resolve(schema))
                    .and_then(|resolved| apache_avro::to_avro_datum(schema, resolved))
                    .map_err(|e| crate::ZerobusError::AvroEncodingError(e.to_string()))
            }),
        })
    }
}

#[cfg(feature = "avro")]
impl AvroPending {
    fn encode(self, schema: &apache_avro::Schema) -> crate::ZerobusResult<Vec<u8>> {
        (self.encode_fn)(schema)
    }
}

impl EncodedBatch {
    pub(crate) fn try_from_encoded_iter<I>(
        iter: I,
        record_type: RecordType,
    ) -> crate::ZerobusResult<Self>
    where
        I: IntoIterator<Item = crate::ZerobusResult<EncodedRecord>>,
    {
        let mut iter = iter.into_iter();
        let (lower, upper) = iter.size_hint();
        let size_hint = upper.unwrap_or(lower);
        match record_type {
            RecordType::Json => {
                let mut vec = SmallVec::with_capacity(size_hint);
                for record in iter {
                    match record? {
                        EncodedRecord::Json(value) => vec.push(value),
                        _ => {
                            return Err(crate::ZerobusError::InvalidArgument(
                                "Record type does not match stream configuration".to_string(),
                            ))
                        }
                    }
                }
                Ok(EncodedBatch::Json(vec))
            }
            RecordType::Proto => { /* same fold */ }
            #[cfg(feature = "avro")]
            RecordType::Avro => { /* same fold */ }
            _ => Err(crate::ZerobusError::InvalidArgument(
                "Record type does not match stream configuration".to_string(),
            )),
        }
    }
}
```

In [`lib.rs`](https://github.com/databricks/zerobus-sdk/blob/9df57f669c2697cb7028c5c7ef83e06d9811796b/rust/sdk/src/lib.rs#L78), drop `IntoEncodedRecord` from the crate-root `pub use` and hide the new type:

```rust
#[doc(hidden)]
pub use record_types::PreparedInput;
pub use record_types::{
    EncodedBatch, EncodedBatchIter, EncodedRecord, JsonEncodedRecord, JsonString, JsonValue,
    ProtoBytes, ProtoEncodedRecord, ProtoMessage,
};
```

In [`ingest.rs`](https://github.com/databricks/zerobus-sdk/blob/9df57f669c2697cb7028c5c7ef83e06d9811796b/rust/sdk/src/stream/grpc/ingest.rs#L14), restore ingest to a schema-free bound:

```rust
use crate::{EncodedBatch, OffsetId, PreparedInput, ZerobusError, ZerobusResult};

pub async fn ingest_record_offset(
    &self,
    payload: impl Into<PreparedInput>,
) -> ZerobusResult<OffsetId> {
    let encoded_batch = self.prepare_record(payload)?;
    self.enqueue_prepared_batch(encoded_batch).await
}

pub async fn ingest_records_offset<I, T>(&self, payload: I) -> ZerobusResult<Option<OffsetId>>
where
    I: IntoIterator<Item = T>,
    T: Into<PreparedInput>,
{
    let encoded_batch = self.prepare_records(payload)?;
    // ...
}
```

Then `prepare_record` / `prepare_records` check the payload, and only the `AvroObject` arm gets a schema. `prepare_records` folds into the `SmallVec` in one pass (no extra `Vec`):

```rust
fn encode_prepared(
    &self,
    prepared: PreparedInput,
    schema: &mut Option<apache_avro::Schema>,
) -> ZerobusResult<EncodedRecord> {
    match prepared {
        PreparedInput::Ready(record) => Ok(record),
        #[cfg(feature = "avro")]
        PreparedInput::AvroObject(pending) => {
            if self.options.record_type != RecordType::Avro {
                return Err(ZerobusError::InvalidArgument(
                    "AvroRecord requires stream record type to be Avro".to_string(),
                ));
            }
            if schema.is_none() {
                *schema = Some(self.get_avro_schema()?.ok_or_else(|| {
                    ZerobusError::InvalidArgument(
                        "Avro schema required but not provided in stream configuration".to_string(),
                    )
                })?);
            }
            Ok(EncodedRecord::Avro(
                pending.encode(schema.as_ref().unwrap())?,
            ))
        }
    }
}

pub(crate) fn prepare_record(
    &self,
    payload: impl Into<PreparedInput>,
) -> ZerobusResult<EncodedBatch> {
    let mut schema = None;
    let encoded_record = self.encode_prepared(payload.into(), &mut schema)?;
    let encoded_batch = EncodedBatch::try_from_record(encoded_record, self.options.record_type)
        .ok_or_else(|| {
            ZerobusError::InvalidArgument(
                "Record type does not match stream configuration".to_string(),
            )
        })?;
    self.validate_ingest_payload(&encoded_batch)?;
    Ok(encoded_batch)
}

pub(crate) fn prepare_records<I, T>(&self, payload: I) -> ZerobusResult<EncodedBatch>
where
    I: IntoIterator<Item = T>,
    T: Into<PreparedInput>,
{
    let mut schema = None;
    let encoded_batch = EncodedBatch::try_from_encoded_iter(
        payload
            .into_iter()
            .map(|rec| self.encode_prepared(rec.into(), &mut schema)),
        self.options.record_type,
    )?;
    self.validate_ingest_payload(&encoded_batch)?;
    Ok(encoded_batch)
}
```

`get_avro_schema` stays, but it is only called from the `AvroObject` arm, so `AvroBytes` and JSON/proto never parse or clone the schema. Unit tests that call `into_encoded(...)` can become `PreparedInput::from(AvroRecord(order))` plus `pending.encode(&schema)`.

[`MultiplexedStream`](https://github.com/databricks/zerobus-sdk/blob/9df57f669c2697cb7028c5c7ef83e06d9811796b/rust/sdk/src/multiplexed_stream.rs#L407) can take the same bound. `EncodedRecord` still works through the blanket `From`. Pass the payload into `prepare_record` without a prior `.into()`:

```rust
pub async fn ingest_record(
    &self,
    payload: impl Into<PreparedInput>,
) -> ZerobusResult<MessageId> {
    self.check_closed()?;
    let idx = self.pick_substream();
    let stream = &self.streams[idx];
    let encoded_batch = stream.prepare_record(payload)?;
    self.enqueue_reserved(stream, idx, encoded_batch).await
}

pub async fn ingest_records<I, T>(&self, payload: I) -> ZerobusResult<Option<MessageId>>
where
    I: IntoIterator<Item = T>,
    T: Into<PreparedInput>,
{
    self.check_closed()?;
    let idx = self.pick_substream();
    let stream = &self.streams[idx];
    let encoded_batch = stream.prepare_records(payload)?;
    if encoded_batch.is_empty() {
        return Ok(None);
    }
    self.enqueue_reserved(stream, idx, encoded_batch)
        .await
        .map(Some)
}
```

rust/sdk/src/stream/grpc/ingest.rs
teodordelibasic-db: PTAL at [this comment](https://github.com/databricks/zerobus-sdk/pull/827/changes/BASE..8ecc3b626f3710197df71f25bdcfda5756ee84b3#r3956901573), maybe something like that can look cleaner then the current approach, although it is complicated as well.

teodordelibasic-db (commented):
Not sure if this is planned for a future PR, but `cargo test --workspace` and `make lint` (`clippy --all`) use default features. Every Avro unit test is `#[cfg(feature = "avro")]`. [`rust/tests/Cargo.toml`](https://github.com/databricks/zerobus-sdk/blob/9df57f669c2697cb7028c5c7ef83e06d9811796b/rust/tests/Cargo.toml#L43) enables `arrow-flight` but not `avro`. `examples/avro` is excluded from the workspace, so it has no CI step either. The green checks do not compile this feature.

rust/examples/avro/src/main.rs
```diff
@@ -0,0 +1,110 @@
...
+        stream
+            .ingest_record_offset(AvroRecord(order))
+            .await?;
+    }
+
+    // Wait once for every pending ack.
+    stream.flush().await?;
+    println!("  {} records acknowledged", 2);
+    Ok(())
+}
```
elenagaljak-db: Also, the example should be more linear. It's a bit confusing with nested ingest methods

elenagaljak-db (commented):
If we support both batch and single, can we make 2 examples single and batch?

rust/sdk/src/lib.rs
irinatomic-db: It's okay to keep it ungated since it's a plain `Option<String>`. It'll just stay `None` in default builds.

Will remove the comment.

rust/sdk/src/stream/grpc/mod.rs
teodordelibasic-db: Following up, could we parse in `.avro()` / `validate()` / `build()` and store the `Schema` on the stream, then borrow it on ingest? That drops the `OnceLock`, fails junk JSON before CreateStream, and avoids the deep `schema.clone()` that `get_avro_schema` does on every `prepare_record` today.

```rust
#[cfg(feature = "avro")]
avro_schema: Option<apache_avro::Schema>, // set at build, no OnceLock

fn writer_schema(&self) -> ZerobusResult<&apache_avro::Schema> {
    self.avro_schema.as_ref().ok_or_else(|| {
        ZerobusError::InvalidArgument(
            "Avro schema required but not provided in stream configuration".to_string(),
        )
    })
}
```

Keep sending the original `avro_schema_json` to the server either way. `avro_sets_format_and_validates` currently expects `.avro(r#"{"type":"record","name":"R","fields":[]}"#)` to succeed. After a parse-at-build change, `.avro("")` and `.avro("{")` should fail `validate()` / `build()`.

rust/examples/avro/src/main.rs
teodordelibasic-db: TBH in general I think we should drop these helper methods in examples that we currently have and instead just have a streamlined main method.

rust/sdk/src/builder/stream_builder.rs
teodordelibasic-db: We should add Avro to this error message as well.

rust/sdk/src/stream/grpc/ingest.rs
```diff
@@ -138,6 +153,41 @@ impl ZerobusStream {
...
+        if self.options.record_type != crate::databricks::zerobus::RecordType::Avro {
+            return Ok(None);
+        }
+
+        let schema_json = self
+            .table_properties
+            .avro_schema_json
+            .as_ref()
+            .ok_or_else(|| {
+                ZerobusError::AvroSchemaParseError(
```
teodordelibasic-db: This may feel more like an `InvalidArgument` error, WDYT?

teodordelibasic-db (commented):
`main.rs` in examples already shows `AvroRecord` first, then `AvroBytes`. The docs around it still describe only the pre-encoded path:

- [`examples/avro/README.md`](https://github.com/databricks/zerobus-sdk/blob/9df57f669c2697cb7028c5c7ef83e06d9811796b/rust/examples/avro/README.md#L1) first sentence: "Ingests pre-encoded Avro datums into an ephemeral stream using `AvroBytes`."
- [`StreamBuilder::avro()`](https://github.com/databricks/zerobus-sdk/blob/9df57f669c2697cb7028c5c7ef83e06d9811796b/rust/sdk/src/builder/stream_builder.rs#L230) rustdoc: "Ingest pre-encoded datums with `AvroBytes`."
- [`examples/README.md`](https://github.com/databricks/zerobus-sdk/blob/9df57f669c2697cb7028c5c7ef83e06d9811796b/rust/examples/README.md#L23) still says three formats and the examples table has no avro row.

Could we update those three to match the code (objects via `AvroRecord`, optional pre-encoded `AvroBytes`)?

rust/examples/avro/README.md
irinatomic-db: Rewritten to match the json/proto READMEs: Overview + Features, a data-passing table, single/batch sections with Running/Code Highlights, and Adapting for Your Custom Table.

rust/sdk/src/stream/grpc/ingest.rs
irinatomic-db: Dropped the `IntoEncodedRecord` trait entirely and went with your` PreparedInput` suggestion (thanks!). 

It then evolved further: the object path is now `AvroRecord(apache_avro::types::Value)` (the serde wrapper is gone), so there's no schema-blind serde step and no Any downcast.

rust/sdk/src/stream/grpc/ingest.rs
irinatomic-db: Reworked per the `PreparedInput` direction: ingest takes `impl Into<PreparedInput>`, and the only avro-specific branch left is encoding the `AvroValue` against the writer schema (which inherently needs the schema). The trait + `Any` handling is gone.

rust/sdk/src/stream/grpc/mod.rs
irinatomic-db: Done. The `OnceLock` is gone = the writer schema is parsed once at `build()` and stored in `TableProperties` next to the JSON wire form (same as `message_descriptor` sits beside `descriptor_proto`), so it's reused on ingest and survives recreation. Malformed JSON now fails at build().

rust/sdk/Cargo.toml
irinatomic-db: Documented on the feature line (`# Avro record format (Beta), opt-in. Requires Rust 1.85 (apache-avro 0.20).`), and noted in the README and changelog.

rust/sdk/src/record_types.rs
irinatomic-db: Thank you!

Adopted this - the trait/`Any` approach is gone. Also fixed the `prepare_records` regression you flagged: it now folds encoded records into the `SmallVec` in one pass (`try_from_encoded_iter`) instead of collecting a `Vec` first. 

One change from your sketch: the object payload is now `AvroRecord(apache_avro::types::Value)` (serde dropped) so the record path is exhaustive, which also let `PreparedInput::AvroObject` hold the value directly (no boxed closure).

rust/examples/avro/src/main.rs
irinatomic-db: Done - `single.rs` and `batch.rs` are now a single linear `main()` with no helper methods.

rust/sdk/src/builder/stream_builder.rs
irinatomic-db: Added `.avro()` to both the `validate()` and `build()` "record format is required" messages.

rust/sdk/src/stream/grpc/ingest.rs
irinatomic-db: I agree, changed

irinatomic-db:
I've updated the PR:

@teodordelibasic-db 
Added a CI check that actually exercises the avro feature - a `--features avro` clippy step and test step in `ci-rust.yml`, since the default `cargo test --workspace / make lint` runs never compiled it. Also updated the doc comments, and went with your suggested `PreparedInput` approach.

@teodordelibasic-db @elenagaljak-db 
**One design change to flag:** the object path is now `AvroRecord(apache_avro::types::Value)` rather than a serde-serializable struct. I had to do this for the object path to be exhaustive - a serde struct goes through a schema-blind `to_value` step that can't express several Avro types (fixed, decimal, bytes, complex/multi-branch unions), which would force clients back to raw bytes for those. `apache_avro::types::Value` can represent every Avro type we support, so the stream can encode any of them against the writer schema. 

The full set is the Avro → Delta type mapping in the [design doc](https://docs.google.com/document/d/1q-XEOvl9i6cfwhmZ_h_t0kreFvLbv-M0hSG5ZysbOB0/edit?tab=t.0#heading=h.bedyga214buk).

Can you plase take another look? Thank you!

rust/sdk/src/lib.rs
irinatomic-db: We don't need it to be under the flag. Since it's an `Option<String>` in default builds it will just be `None`

rust/README.md
```diff
@@ -50,6 +50,7 @@ The Zerobus Rust SDK provides a robust, async-first interface for ingesting larg
 - **Graceful Stream Management** - Proper flushing and acknowledgment tracking
 - **Acknowledgment Callbacks** - Receive notifications when records are acknowledged or encounter errors
 - **Arrow Flight Ingestion** (opt-in) — Stream Apache Arrow `RecordBatch` data directly to Zerobus using Arrow Flight's gRPC transport. Enable with `features = ["arrow-flight"]`; see [`examples/arrow/`](https://github.com/databricks/zerobus-sdk/tree/main/rust/examples/arrow).
+- **Avro Ingestion** *(Beta, opt-in)* — Ingest Avro records on ephemeral streams. Enable with `features = ["avro"]`, select via `.avro(schema_json)`. Build records as `AvroValue` and let the stream encode them against the writer schema (`AvroRecord`), or ingest pre-encoded datums via `AvroBytes`. See [`examples/avro/`](https://github.com/databricks/zerobus-sdk/tree/main/rust/examples/avro). Note: the `avro` feature requires Rust 1.85 (via `apache-avro`); default builds are unaffected. Server support is pending.
```
elenagaljak-db: There are more places where we mention json/proto that we should add to, like on lines 96-100, 199, 251, 406, 526, 616, 1080...  I probably missed something

rust/.gitignore
teodordelibasic-db: We actually probably shouldn't ever "ungate" Avro the same way we didn't "ungate" Arrow. The `features` flag isn't necessarily for gating a feature, but instead just lets users not pull in parts of the binary they are not interested in. Ideally we would have done this for more parts of the SDK.

rust/README.md
irinatomic-db: Thanks.

Added Avro to the README enumeration spots: the Quick Start intro + serialization list (96–101), the examples dir tree (199), the architecture overview (251), the Usage-guide approaches (406), a new "Avro Stream" subsection under Create a Stream (526), the ingest wrapper table (616), the examples run table (1080), and the build-from-source commands. All marked (Beta, opt-in).

One thing I deliberately left out: the source rustdoc that renders on docs.rs (e.g. `record_type`'s enum list, the `ingest_`* param docs, `stream/mod.rs`). Those sit on non-gated core items, so mentioning Avro there either needs per-line #[cfg] or leaks Beta detail into the default-build docs. My inclination is to hold those until Avro has official/server support (e.g. Public Preview) and update them then. WDYT?

elenagaljak-db: Yess, sounds good!

rust/sdk/src/lib.rs
```diff
@@ -124,6 +130,9 @@ pub(crate) struct TableProperties {
     pub(crate) table_name: String,
     pub(crate) descriptor_proto: Option<prost_types::DescriptorProto>,
     pub(crate) message_descriptor: Option<MessageDescriptor>,
+    pub(crate) avro_schema_json: Option<String>,
```
teodordelibasic-db: Ah so actually we need both the parsed and unparsed versions of the schema so we can send the raw one on create stream to the server and we use the parsed one for records? In that case I guess `avro_schema_json` should also be under the feature flag and maybe we can create a struct in the SDK that holds both of these just for readability, but I am fine either way.

rust/Cargo.toml
irinatomic-db: I clarified the comment. 

- The avro feature is permanent (never ungated) - it just lets users who don't need Avro skip apache-avro and its Rust 1.85 floor.
- The workspace exclusion (and this untracked lockfile) is the genuinely temporary part. ffi/jni currently build the sdk without avro and match `EncodedRecord`/`EncodedBatch` exhaustively, so making `examples/avro` a member would unify `avro` onto them and break those matches (`EncodedRecord::Avro(_)` not covered).

It goes away when ffi/jni gain Avro support (#830): they'll enable avro on the sdk dependency just like they already do for arrow-flight, add the match arms, and examples/avro rejoins members - at which point this lockfile is tracked like any other.

rust/sdk/src/record_types.rs
```diff
@@ -69,6 +78,19 @@ impl From<ProtoBytes> for EncodedRecord {
     }
 }
 
+/// Wrapper for a pre-encoded Avro binary datum (Beta).
```
teodordelibasic-db: Sort of a nit regarding just code organization and ordering, could we group the wrappers here by who encodes them? `AvroBytes` currently sits between `ProtoBytes` and `JsonString`, so the pre-encoded types are no longer together and Avro looks inserted into the old proto/JSON pair. `EncodedRecord` at the top is good. After that, something like:

- already encoded: `ProtoBytes`, `JsonString`, `AvroBytes`
- SDK encodes in `From`: `ProtoMessage`, `JsonValue`
- SDK encodes on the stream: `AvroRecord`
- then `PreparedInput` and `EncodedBatch`

`AvroRecord` is the only wrapper that cannot become an `EncodedRecord` by itself, so keeping it next to `PreparedInput` makes that split easier to see. Same file is fine.

rust/sdk/src/lib.rs
irinatomic-db: Yes. Reorganized the code so we have a
```rust
#[cfg(feature = "avro")]
pub(crate) struct AvroSchema {
    pub(crate) json: String,
    pub(crate) parsed: apache_avro::Schema,
}
```

rust/sdk/src/record_types.rs
irinatomic-db: Fixed : D

rust/sdk/src/builder/stream_builder.rs
teodordelibasic-db: This comment is outdated.

teodordelibasic-db (commented):
[`StreamConfigurationOptions.record_type`]([`rust/sdk/src/stream_configuration.rs:91`](https://github.com/databricks/zerobus-sdk/blob/2362006a901d60b134543fcf92f019df627bfe9d/rust/sdk/src/stream_configuration.rs#L91)) still lists Proto/Json/Unspecified only.

teodordelibasic-db (commented):
Let's also fix `ingest_record` / `ingest_records` in `multiplexed_stream.rs` as they still take `Into<EncodedRecord>`. 

```rust
pub async fn ingest_record(
    &self,
    payload: impl Into<PreparedInput>,
) -> ZerobusResult<MessageId> {
    self.check_closed()?;
    let idx = self.pick_substream();
    let stream = &self.streams[idx];
    let encoded_batch = stream.prepare_record(payload)?;
    self.enqueue_reserved(stream, idx, encoded_batch).await
}
```

rust/sdk/src/stream/grpc/ingest.rs
teodordelibasic-db: This should be updated.

teodordelibasic-db: This as well kind of due to `EncodedRecord` -> `PreparedInput`.

rust/sdk/src/lib.rs
```diff
@@ -119,11 +125,23 @@ pub enum StreamType {
 /// -`InvalidTableName`: table_name contains invalid characters or doesn't exist
 /// -`PermissionDenied`: insufficient permissions to write to the specified table
 /// -`InvalidArgument`: invalid or missing descriptor_proto or auth token
+/// The Avro writer schema in the two forms the SDK needs: the raw JSON sent to the server
```
teodordelibasic-db, line 119: This is inserted at a wrong place and it now intercepts the comment for `TableProperties` below.

rust/sdk/src/builder/stream_builder.rs
teodordelibasic-db: `validate()` rustdoc says it performs the same checks as `build()`, but it only checks that table, auth, and format are set. [`build()`](https://github.com/databricks/zerobus-sdk/blob/f7363a9101624ad1fca1106281a5a466342a25c8/rust/sdk/src/builder/stream_builder.rs#L450) already calls `validate()` first, then parses the writer schema, so `.avro("{").validate()` is `Ok(())` while `build()` returns `AvroSchemaParseError`. `validate` is `&self`, so it cannot store the parsed `AvroSchema` that `build()` still has to put on `TableProperties`.

Could we pick one of these?

1. Leave `validate()` as a required-fields check, and say in the rustdoc that Avro schema parsing happens in `build()`:

```rust
/// Returns `Ok(())` if table name, authentication, and format are all set.
/// Avro writer-schema parsing happens in `build()`, not here.
```

2. Parse in both `validate()` and `build()` (or a small helper both call). `validate` stays `&self`; `build()` parses again to store the result:

```rust
apache_avro::Schema::parse_str(json)?;
```

3. Make `.avro()` return `ZerobusResult<Self>` and store `AvroSchema` on `FormatConfig`. The chain picks up a `?`; `validate()` / `build()` no longer parse:

```rust
pub fn avro(mut self, schema_json: impl Into<String>) -> ZerobusResult<Self> {
    let json = schema_json.into();
    let parsed = apache_avro::Schema::parse_str(&json)?;
    self.format = Some(FormatConfig::Avro(AvroSchema { json, parsed }));
    Ok(self)
}
```

4. Keep `.avro()` returning `Self`, parse inside it, and stash `Result<AvroSchema, ZerobusError>` on `FormatConfig`. `validate()` / `build()` return that error:

```rust
enum FormatConfig { Avro(Result<AvroSchema, ZerobusError>), /* ... */ }
```

Options 3 and 4 still copy onto `TableProperties` for ingest and recovery; `FormatConfig` does not outlive `build()`.

We can discuss offline.

rust/NEXT_CHANGELOG.md
teodordelibasic-db, line 24: Enabling `avro` also adds `EncodedRecord::Avro` / `EncodedBatch::Avro` on enums that are not `#[non_exhaustive]` (that is why the example crate is excluded from the workspace). Could we add those match arms to the migration list, and mention the ingest bound change from `Into<EncodedRecord>` to `Into<PreparedInput>` under API Changes?

rust/sdk/src/builder/stream_builder.rs
irinatomic-db: I would go with the 1st option. 

That way it's consistent with proto - `validate()` doesn't deep-validate a descriptor either and `build()` still fails fast before any connection.

rust/NEXT_CHANGELOG.md
irinatomic-db: Added a migration bullet for the avro-only `EncodedRecord::Avro`/`EncodedBatch::Avro` variants, and an API Changes entry for the `Into<EncodedRecord>` → `Into<PreparedInput>` ingest bound.

.github/workflows/ci-rust.yml
```diff
@@ -54,6 +54,9 @@ jobs:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Lint code
         run: make lint
+      # The avro feature is off by default, so `make lint` above never compiles it.
```
teodordelibasic-db: I've uncovered a miss in our CI checks/make targets when reviewing this PR, hence published [[CI] Expand Rust lint and test coverage](https://github.com/databricks/zerobus-sdk/pull/850). Once it is merged, you can rebase your change on top of it and you can drop all of the changes in this file. Avro should run in CI out of the box. Feel free to double check after rebasing though.

rust/README.md
teodordelibasic-db: This one should be updated as well.

teodordelibasic-db: ditto.

teodordelibasic-db (dismissed):
LGTM once the remaining three comments are addressed and we double check CI. We should also probably not link internal Google Docs in the PR description, but no harm done I guess. Feel free to ping for reapproval.

irinatomic-db:
@teodordelibasic-db 

I've rebased onto main and removed my step from the CI job. The avro tests are running as part of the CI since you've added the `--all-features` arg. Also cleaned up the README and fixed the examples to encode the record with apache_avro instead of hand-written bytes.

teodordelibasic-db (approved):
LGTM

elenagaljak-db (approved):

PR #828, opened 2026-09-03
[PureGo] Add Avro record format (avro build tag)

irinatomic-db:
## PR stack

    - [avro/purego](https://github.com/databricks/zerobus-sdk/pull/828) <- _this PR_
    - [avro/rust-core](https://github.com/databricks/zerobus-sdk/pull/827) [MERGED]
        - avro/java
        - avro/typescript

Adds the Avro record format to the pure-Go SDK, behind the `avro` build tag. Ephemeral streams only; feature in development. Default builds are unchanged — tagged files aren't compiled and no Avro dependency is pulled in.

Declare the writer schema once at stream creation with `WithAvro(schemaJSON)`; it is parsed and validated there, so a malformed schema fails fast. Records are then ingested in one of two ways: pre-encoded bytes via `IngestRecordOffset([]byte)` / `IngestRecordsOffset([][]byte)` (raw Avro datums the caller already encoded), or record objects the SDK encodes via `IngestAvroRecordOffset(AvroRecord)` / `IngestAvroRecordsOffset([]AvroRecord)`, where `AvroRecord` is a `map[string]any`.

`AvroRecord` covers every Avro type in the Avro-to-Delta mapping, including unions (via the `Union("branch", v)` helper), `fixed`, `decimal`, and logical types, using hamba's native value conventions documented on the `AvroRecord` type.

Implementation notes: object encoding uses `github.com/hamba/avro/v2`, imported only in `//go:build avro` files, so the pre-encoded bytes path needs no Avro library and stays available without the tag. The writer schema is parsed once at stream creation and reused across recovery, with no per-record parsing. Object-encode failures return non-retryable errors.

## How is this tested?

```
cd purego

# Default (feature off) — tagged files skipped, no Avro dep
gofmt -l . && go build ./... && go vet ./... && go test ./...

# With the avro tag (all-types encode test, fail-fast test, gated WithAvro tests, e2e)
go build -tags avro ./... && go vet -tags avro ./... && go test -tags avro ./...

# Examples are a separate module; build the avro example with the tag
cd examples && go build -tags avro ./...
```

CI: `ci-purego.yml` has an `avro` job covering all of the above.


purego/zerobus/stream.go
```diff
@@ -21,39 +23,73 @@ type Stream struct {
...
 //
-// record is the raw record payload: serialized protobuf bytes for a proto
-// stream, or UTF-8 JSON bytes for a JSON stream.
+// record is []byte (pre-encoded: protobuf for a proto stream, UTF-8 JSON for a
+// JSON stream, or a raw datum for an Avro stream) or, with the avro build tag,
+// an AvroRecord the stream encodes against its writer schema.
 //
 // For throughput, queue records in a loop and call Flush once.
-func (s *Stream) IngestRecordOffset(record []byte) (int64, error) {
+func (s *Stream) IngestRecordOffset(record any) (int64, error) {
```
danilonajkov-db: I see we change []bytes to any to allow ingestion of AvroRecord.
I don't know go at all, but is it possible to leave it as bytes and have an overload for AvroRecord. It's weird to have the main ingest record accept any

purego/zerobus/avro_test.go
```diff
@@ -0,0 +1,48 @@
+//go:build avro
```
danilonajkov-db, line 1: we don't have any integration tests (manual) for pure-go?

purego/zerobus/avro_test.go
```diff
@@ -0,0 +1,48 @@
...
+		t.Fatalf("encode: %v", err)
+	}
+	if len(b) == 0 {
+		t.Fatal("want non-empty datum")
+	}
+
+	if _, err := s.encodeObjectRecord([]byte{0x01}); err == nil {
+		t.Fatal("want error for non-AvroRecord type")
+	}
+}
```
danilonajkov-db, line 98: maybe you can also add a test with ingesting raw avro bytes

purego/zerobus/stream.go
danilonajkov-db: how is it done for json for example?

irinatomic-db: JSON uses a dedicated `IngestJSONOffset` rather than widening the generic method, so I'll follow that: `IngestRecordOffset` stays `[]byte`, and Avro objects get a typed `IngestAvroRecordOffset(AvroRecord)` (+ batch).

purego/zerobus/avro_test.go
irinatomic-db: I'll add an integration test against the in-memory server covering both the object and raw-bytes paths.

PR #829, opened 2026-09-03
[PureC] Add pure-C SDK skeleton

danilotrninic-db:

The initial skeleton of the pure-C Zerobus SDK: a native,
dependency-free C implementation of the ingestion client. Code and build 
only in this first commit:

- Public API in `include/zerobus/`: opaque SDK/stream builders and handles,
  fixed-ABI status codes, and owned structured errors.
- Input validation and error/ownership plumbing in `src/`: UTF-8, JSON,
  table-name and endpoint-URL checks, transactional setters, secure-zeroed
  OAuth credentials.
- Build and tests: CMake, Makefile wrappers, `.clang-format`, offline unit
  tests, and a single-producer example.

## How is this tested?

`make verify` passes end to end: `clang-format`, the `-Wall -Wextra -Werror`
build, and the offline unit tests (4/4) plain and under AddressSanitizer/UBSan
and ThreadSanitizer/UBSan. The tests cover the input validators (UTF-8, JSON,
table name, endpoint URL), error ownership, and the public-API validation and
precondition behavior.

No integration or end-to-end tests yet.

### DIFF
--- a/purec/CMakeLists.txt
+++ b/purec/CMakeLists.txt
@@ -1,0 +1,72 @@
+# Zerobus Pure C SDK build: the library, examples, and tests.
+#
+# No third-party dependencies — the SDK is the public API, its input validation,
+# and error/ownership plumbing, so it needs nothing beyond a C99 compiler and
+# the C standard library. The networking core (transport, OAuth, record codec,
+# threading) is not implemented yet; it will add nghttp2 and OpenSSL.
+cmake_minimum_required(VERSION 3.13)
+project(zerobus_c LANGUAGES C VERSION 0.0.1)
+
+set(CMAKE_C_STANDARD 99)
+set(CMAKE_C_STANDARD_REQUIRED ON)
+# Reject compiler extensions: build as -std=c99, not -std=gnu99, so a GNU-ism
+# cannot slip in unnoticed against the documented C99 baseline.
+set(CMAKE_C_EXTENSIONS OFF)
+if(NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE Release)
+endif()
+
+# Hidden visibility by default: only ZEROBUS_API symbols are exported from a
+# shared build. ELF visibility only affects the dynamic symbol table, so the
+# static archive still carries its external zb_* helpers as globals.
+set(CMAKE_C_VISIBILITY_PRESET hidden)
+set(CMAKE_VISIBILITY_INLINES_HIDDEN ON)
+
+# Optional sanitizer, e.g. -DZEROBUS_SANITIZE=address (or thread, undefined).
+set(ZEROBUS_SANITIZE "" CACHE STRING "Runtime sanitizer: address|thread|undefined")
+
+# Treat compiler warnings as errors (used by `make lint`; off for normal builds).
+option(ZEROBUS_WERROR "Treat compiler warnings as errors" OFF)
+
+# Warning flags shared by the library, tests, and example. -Werror is appended
+# only under ZEROBUS_WERROR, so a normal build surfaces warnings without failing.
+set(ZEROBUS_WARN_FLAGS -Wall -Wextra)
+if(ZEROBUS_WERROR)
+  list(APPEND ZEROBUS_WARN_FLAGS -Werror)
+endif()
+
+set(ZEROBUS_SOURCES
+  src/error.c
+  src/utils.c
+  src/sdk.c
+  src/stream.c
+)
+
+add_library(zerobus_c STATIC ${ZEROBUS_SOURCES})
+add_library(zerobus::c ALIAS zerobus_c)
+
+target_include_directories(zerobus_c
+  PUBLIC  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+  PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/src)
+
+target_compile_options(zerobus_c PRIVATE ${ZEROBUS_WARN_FLAGS})
+
+# Same sources compiled with ZB_TESTING, which gives ZB_STATIC helpers external
+# linkage so unit tests can call them directly. Not shipped.
+add_library(zerobus_c_testing STATIC ${ZEROBUS_SOURCES})
+target_compile_definitions(zerobus_c_testing PUBLIC ZB_TESTING)
+target_include_directories(zerobus_c_testing
+  PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+         ${CMAKE_CURRENT_SOURCE_DIR}/src)
+target_compile_options(zerobus_c_testing PRIVATE ${ZEROBUS_WARN_FLAGS})
+
+if(ZEROBUS_SANITIZE)
+  foreach(lib zerobus_c zerobus_c_testing)
+    target_compile_options(${lib} PUBLIC -fsanitize=${ZEROBUS_SANITIZE} -fno-omit-frame-pointer -g)
+    target_link_options(${lib} PUBLIC -fsanitize=${ZEROBUS_SANITIZE})
+  endforeach()
+endif()
+
+enable_testing()
+add_subdirectory(examples)
+add_subdirectory(tests)
--- a/purec/Makefile
+++ b/purec/Makefile
@@ -1,0 +1,71 @@
+# Makefile for the Zerobus Pure C SDK: convenience wrappers around CMake/CTest.
+
+BUILD_TYPE ?= Release
+JOBS ?= 4
+
+SANITIZE ?=
+SANITIZE_FLAG := $(if $(SANITIZE),-DZEROBUS_SANITIZE=$(SANITIZE),)
+
+# Extra -D flags forwarded to cmake configure (lint sets -DZEROBUS_WERROR=ON).
+CMAKE_EXTRA ?=
+
+# Sanitizer builds use their own build dir so they don't clash with a normal
+# build dir. Override BUILD_DIR to force a specific path.
+BUILD_DIR ?= $(if $(SANITIZE),build-$(SANITIZE),build)
+
+# All hand-written C sources (excludes the CMake build tree).
+SOURCES := $(shell find include src examples tests -name '*.c' -o -name '*.h' 2>/dev/null)
+
+.PHONY: help configure build test examples lint fmt fmt-check clean check verify
+
+help:
+	@echo "Available targets:"
+	@echo "  make build      - Configure and build the SDK, examples, and tests"
+	@echo "  make test       - Build and run the offline test suite (ctest)"
+	@echo "  make examples   - Build the example binaries"
+	@echo "  make fmt        - Format all C sources with clang-format"
+	@echo "  make fmt-check  - Check formatting without modifying files"
+	@echo "  make lint       - Warnings-as-errors build (-Werror)"
+	@echo "  make check      - fmt-check + lint (static gates, no tests)"
+	@echo "  make verify     - Full workspace check: check + tests, plain and sanitized"
+	@echo "  make clean      - Remove the build directory"
+	@echo ""
+	@echo "  Add SANITIZE=address (or thread/undefined) to build+test with a"
+	@echo "  sanitizer, e.g. 'make test SANITIZE=address'."
+
+configure:
+	cmake -S . -B $(BUILD_DIR) -DCMAKE_BUILD_TYPE=$(BUILD_TYPE) \
+		-DCMAKE_EXPORT_COMPILE_COMMANDS=ON $(SANITIZE_FLAG) $(CMAKE_EXTRA)
+
+build: configure
+	cmake --build $(BUILD_DIR) -j $(JOBS)
+
+test: build
+	cd $(BUILD_DIR) && ctest --output-on-failure
+
+examples: build
+	@echo "Examples built under $(BUILD_DIR)/examples/"
+
+fmt:
+	clang-format -i $(SOURCES)
+
+fmt-check:
+	clang-format --dry-run --Werror $(SOURCES)
+
+# No clang-tidy here, so -Wall -Wextra is the linter: run the build with
+# -Werror. Its own tree keeps the flag out of the normal build's cache.
+lint:
+	$(MAKE) build BUILD_DIR=build-lint CMAKE_EXTRA=-DZEROBUS_WERROR=ON
+
+check: fmt-check lint
+
+# Full workspace check: static checks, then the suite plain and under all three
+# sanitizers. address and thread can't coexist, so undefined rides with each.
+verify:
+	$(MAKE) check
+	$(MAKE) test
+	$(MAKE) test SANITIZE=address,undefined
+	$(MAKE) test SANITIZE=thread,undefined
+
+clean:
+	rm -rf build build-*
--- a/purec/Makefile
+++ b/purec/Makefile
@@ -13,10 +13,24 @@ CMAKE_EXTRA ?=
 # build dir. Override BUILD_DIR to force a specific path.
 BUILD_DIR ?= $(if $(SANITIZE),build-$(SANITIZE),build)
 
+# Coverage uses its own build dir (matched by `clean`'s build-* glob).
+COV_DIR ?= build-cov
+COV_OBJ_DIR := $(COV_DIR)/CMakeFiles/zerobus_c_testing.dir/src
+
+# gcov must match the compiler that built the objects or it misreads the
+# coverage data (.gcno/.gcda). CMake builds with cc, so pick the gcov-N whose
+# major version matches cc, and fall back to plain gcov. Override with GCOV=...
+GCOV ?= $(shell \
+	major=$$(cc -dumpversion 2>/dev/null | cut -d. -f1); \
+	if command -v gcov-$$major >/dev/null 2>&1; \
+		then echo gcov-$$major; \
+		else echo gcov; \
+	fi)
+
 # All hand-written C sources (excludes the CMake build tree).
 SOURCES := $(shell find include src examples tests -name '*.c' -o -name '*.h' 2>/dev/null)
 
-.PHONY: help configure build test examples lint fmt fmt-check clean check verify
+.PHONY: help configure build test examples lint fmt fmt-check clean check verify coverage
 
 help:
 	@echo "Available targets:"
@@ -28,7 +42,8 @@ help:
 	@echo "  make lint       - Warnings-as-errors build (-Werror)"
 	@echo "  make check      - fmt-check + lint (static gates, no tests)"
 	@echo "  make verify     - Full workspace check: check + tests, plain and sanitized"
-	@echo "  make clean      - Remove the build directory"
+	@echo "  make coverage   - Coverage build+test, then gcov: read .gcov files for uncovered lines (#####)"
+	@echo "  make clean      - Remove the build dirs (build, build-*, and BUILD_DIR)"
 	@echo ""
 	@echo "  Add SANITIZE=address (or thread/undefined) to build+test with a"
 	@echo "  sanitizer, e.g. 'make test SANITIZE=address'."
@@ -62,10 +77,21 @@ check: fmt-check lint
 # Full workspace check: static checks, then the suite plain and under all three
 # sanitizers. address and thread can't coexist, so undefined rides with each.
 verify:
+	$(MAKE) clean
 	$(MAKE) check
 	$(MAKE) test
 	$(MAKE) test SANITIZE=address,undefined
 	$(MAKE) test SANITIZE=thread,undefined
 
+# Coverage: build and test with gcov instrumentation, then run gcov.
+# Read the generated .gcov files to see which lines are not covered.
+coverage:
+	rm -rf $(COV_DIR)
+	cmake -S . -B $(COV_DIR) -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_FLAGS="--coverage -O0 -g" \
+		-DCMAKE_EXE_LINKER_FLAGS="--coverage" $(CMAKE_EXTRA)
+	cmake --build $(COV_DIR) -j $(JOBS)
+	cd $(COV_DIR) && ctest --output-on-failure
+	cd $(COV_OBJ_DIR) && $(GCOV) *.gcno
+
 clean:
-	rm -rf build build-*
+	rm -rf $(BUILD_DIR) build build-*
--- a/purec/examples/json/main.c
+++ b/purec/examples/json/main.c
@@ -1,0 +1,139 @@
+/*
+ * Minimal end-to-end JSON ingestion example: build an SDK, open an
+ * OAuth-authenticated ephemeral JSON stream, queue a few records without
+ * waiting after each one, flush once, then close and free everything.
+ *
+ * NOTE: the SDK is under early development. The operations validate inputs and
+ * return the documented statuses but perform no network I/O yet, so this shows
+ * the API shape and lifecycle rather than delivering records.
+ *
+ * The table schema must match the two example JSON fields (id, message), or the
+ * records must be replaced with ones matching the selected table.
+ */
+#include <stdio.h>
+#include <string.h>
+
+#include <zerobus/zerobus.h>
+
+static zerobus_string_view_t string_view(const char *value)
+{
+    return (zerobus_string_view_t){value, strlen(value)};
+}
+
+static void print_error(const zerobus_error_t *error)
+{
+    if (error == NULL) {
+        fputs("zerobus: operation failed without details\n", stderr);
+        return;
+    }
+    zerobus_string_view_t message = zerobus_error_message(error);
+    fprintf(stderr, "zerobus: %.*s\n", (int)message.len, message.data);
+}
+
+int main(int argc, char **argv)
+{
+    if (argc != 6) {
+        fprintf(stderr,
+                "usage: %s <zerobus-endpoint> <uc-endpoint> "
+                "<table> <client-id> <client-secret>\n",
+                argv[0]);
+        return 2;
+    }
+
+    const char *records[] = {
+        "{\"id\":1,\"message\":\"hello\"}",
+        "{\"id\":2,\"message\":\"from C\"}",
+    };
+
+    zerobus_error_t *error = NULL;
+    zerobus_sdk_builder_t *sdk_builder = NULL;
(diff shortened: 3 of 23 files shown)


purec/src/error.h
teodordelibasic-db, line 1: Duplicate `error.h` file?

purec/src/utils.h
teodordelibasic-db, line 1: Should this file by in `purec/include/`?

purec/src/stream.c
```diff
@@ -0,0 +1,248 @@
...
+                       "sdk and out_builder must not be NULL");
+    }
+    zerobus_stream_builder_t *b =
+        (zerobus_stream_builder_t *)calloc(1, sizeof(*b));
+    if (b == NULL) {
+        return ZEROBUS_STATUS_OUT_OF_MEMORY;
+    }
+    b->sdk = sdk;
+    *out_builder = b;
+    return ZEROBUS_STATUS_OK;
```
teodordelibasic-db, line 52: Hm what do you think about returning `ZEROBUS_STATUS_UNIMPLEMENTED` in all of these places where the functions are not actually working yet? It's not that important in any case but just for the sake of consistency.

purec/include/zerobus/stream.h
```diff
@@ -0,0 +1,79 @@
...
+/*
+ * Ingest one JSON record: exactly one well-formed UTF-8 JSON value, sent
+ * verbatim (not reshaped to the table schema). Success means the record was
+ * copied and queued locally, not acknowledged by the server, so the caller may
+ * reuse the bytes on return.
+ *
+ * The record format is fixed at build time. Other formats add their own ingest
+ * functions.
+ */
+ZEROBUS_API zerobus_status_t ZEROBUS_CALL zerobus_stream_ingest_json_record(
```
teodordelibasic-db, line 64: Should this method have an `out` argument for offset as well?

purec/include/zerobus/stream.h
```diff
@@ -0,0 +1,79 @@
...
+ */
+ZEROBUS_API zerobus_status_t ZEROBUS_CALL zerobus_stream_builder_build(
+    const zerobus_stream_builder_t *builder, zerobus_stream_t **out_stream,
+    zerobus_error_t **out_error);
+
+ZEROBUS_API void ZEROBUS_CALL
+zerobus_stream_builder_free(zerobus_stream_builder_t *builder);
+
+/*
+ * Ingest one JSON record: exactly one well-formed UTF-8 JSON value, sent
```
teodordelibasic-db: I think a full recursive validation of JSON records is an overkill, especially at this moment. It increases thread stack usage and is on the hot path. For now just a cheap gate (non-empty, size, maybe UTF-8) should we enough. If you want we can leave a TODO comment to look into full record validation later, but it is not of high priority.

```c
    if (zb_is_empty(json_record)) {
        return zb_fail(out_error, ZEROBUS_STATUS_INVALID_ARGUMENT,
                       "record must be non-empty");
    }
    if (json_record.len > ZB_MAX_PAYLOAD_BYTES) {
        return zb_fail(out_error, ZEROBUS_STATUS_INVALID_ARGUMENT,
                       "record of %zu bytes exceeds the %u-byte limit",
                       json_record.len, (unsigned)ZB_MAX_PAYLOAD_BYTES);
    }
```

purec/include/zerobus/sdk.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include "zerobus/common.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/*
+ * SDK builder. The endpoint setters validate the URL and copy their inputs, so
+ * a malformed endpoint is rejected at set time. TLS is always enabled with
```
teodordelibasic-db: Since TLS is hardcoded on ([`sdk.c:64`](https://github.com/databricks/zerobus-sdk/blob/94168aede4700d6901159e19a6600bf7acc8c056/purec/src/sdk.c#L64) rejects `http://`), and there is no `disable_tls`. The dotted-host check at [`sdk.c:104`](https://github.com/databricks/zerobus-sdk/blob/94168aede4700d6901159e19a6600bf7acc8c056/purec/src/sdk.c#L104) also rejects `localhost`. Together that blocks a local plaintext server.

Could we allow plaintext, either by accepting `http://` or with a `disable_tls` setter, and drop the "must contain a dot" host rule? One of those TLS knobs is enough; they do not both need to exist.

purec/src/sdk.c
```diff
@@ -0,0 +1,213 @@
...
+    char *endpoint;
+    char *uc_endpoint;
+};
+
+/* ---- builder ----------------------------------------------------------- */
+
+zerobus_status_t zerobus_sdk_builder_new(zerobus_sdk_builder_t **out_builder,
+                                         zerobus_error_t **out_error)
+{
+    if (out_error != NULL) {
```
teodordelibasic-db: Every public entry point does `*out_error = NULL` on entry, and `zb_fail` overwrites the slot again without freeing. Two calls that reuse the same `zerobus_error_t *err` leak the first object even if the second succeeds. The tests free-then-NULL before reuse; a caller who does not will leak.

Could we document that `*out_error` must be NULL on entry, and refuse to clobber a live pointer instead of NULLing it?

```c
    if (out_error != NULL && *out_error != NULL) {
        return ZEROBUS_STATUS_INVALID_ARGUMENT;
    }
```

purec/include/zerobus/sdk.h
```diff
@@ -0,0 +1,48 @@
...
+
+ZEROBUS_API zerobus_status_t ZEROBUS_CALL
+zerobus_sdk_builder_build(const zerobus_sdk_builder_t *builder,
+                          zerobus_sdk_t **out_sdk, zerobus_error_t **out_error);
+
+ZEROBUS_API void ZEROBUS_CALL
+zerobus_sdk_builder_free(zerobus_sdk_builder_t *builder);
+
+/*
+ * Best-effort cleanup, cannot report failures. The SDK is borrowed by every
```
teodordelibasic-db, line 38: Each stream keeps a pointer to the SDK and the docs say the SDK must outlive every stream. `zerobus_sdk_free` does not know about live streams, so freeing the SDK first leaves those pointers dangling. Today the stream barely uses that pointer, but it will once the SDK owns TLS or a runtime.

Could the stream copy what it needs from the SDK and not keep the pointer? Then free order would not matter.

purec/examples/json/main.c
```diff
@@ -0,0 +1,139 @@
...
+        return;
+    }
+    zerobus_string_view_t message = zerobus_error_message(error);
+    fprintf(stderr, "zerobus: %.*s\n", (int)message.len, message.data);
+}
+
+int main(int argc, char **argv)
+{
+    if (argc != 6) {
+        fprintf(stderr,
```
teodordelibasic-db, line 32: Usage takes `<client-secret>` as `argv[5]`, which shows up in shell history and process listings. As the first example this will be copied. Could we load the secret from an environment variable (or stdin / a protected file) and keep it out of the usage string?

```c
    const char *client_secret = getenv("DATABRICKS_CLIENT_SECRET");
    if (client_secret == NULL) {
        fputs("DATABRICKS_CLIENT_SECRET is required\n", stderr);
        return 2;
    }
```

purec/include/zerobus/error.h
```diff
@@ -0,0 +1,28 @@
...
+extern "C" {
+#endif
+
+/* Failable functions take an optional `zerobus_error_t **out_error`. On
+ * failure it may be populated with an owned error, which the caller must free
+ * with zerobus_error_free. Passing NULL discards the error. */
+
+/* The returned view is borrowed from the error, valid only until it is freed.
+ */
+ZEROBUS_API zerobus_string_view_t ZEROBUS_CALL
```
teodordelibasic-db, line 41: Publicly the error is a message plus `zerobus_error_free`. Internally it stores `code` ([`src/error.h:10`](https://github.com/databricks/zerobus-sdk/blob/94168aede4700d6901159e19a6600bf7acc8c056/purec/src/error.h#L10)) and never exposes it. Every other SDK, and FFI `CResult`, has `is_retryable`. Callers who only keep the error object lose the status, and retry loops will have to guess from the message.

Could we add `zerobus_error_status()` and `zerobus_error_is_retryable()` now, even if most variants are unused? Accessors on an opaque type stay additive.

```c
ZEROBUS_API zerobus_status_t ZEROBUS_CALL
zerobus_error_status(const zerobus_error_t *error);
ZEROBUS_API bool ZEROBUS_CALL
zerobus_error_is_retryable(const zerobus_error_t *error);
```

purec/src/utils.c
```diff
@@ -0,0 +1,533 @@
...
+    *target = copy;
+    return true;
+}
+
+char *zb_strndup(const char *src, size_t len)
+{
+    if (src == NULL) {
+        return NULL;
+    }
+    char *out = (char *)malloc(len + 1);
```
teodordelibasic-db, line 257: This allocates `len + 1` then copies `len` bytes. If a string view has `len == SIZE_MAX` (a bad `{pointer, length}`, not a normal C string), that add wraps to 0, `malloc` returns a tiny buffer, and `memcpy` writes a huge amount into it.

Unlikely in day-to-day use; cheap to close. Could we fail when `len == SIZE_MAX` so the add cannot wrap?

```c
    if (src == NULL || len == SIZE_MAX) {
        return NULL;
    }
```

purec/include/zerobus/stream.h
```diff
@@ -0,0 +1,79 @@
+/*
+ * Zerobus Pure C SDK — stream builder, stream handle, and ingestion.
```
teodordelibasic-db, line 2: Could we document that ingest/flush/close are not thread-safe and one caller thread should be used per stream?

purec/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
...
+# Reject compiler extensions: build as -std=c99, not -std=gnu99, so a GNU-ism
+# cannot slip in unnoticed against the documented C99 baseline.
+set(CMAKE_C_EXTENSIONS OFF)
+if(NOT CMAKE_BUILD_TYPE)
+  set(CMAKE_BUILD_TYPE Release)
+endif()
+
+# Hidden visibility by default: only ZEROBUS_API symbols are exported from a
+# shared build. ELF visibility only affects the dynamic symbol table, so the
+# static archive still carries its external zb_* helpers as globals.
```
teodordelibasic-db, line 21: These are only used from their own `.c` (plus tests), so they can use the existing `ZB_STATIC` / `ZB_TESTING` split:

- `zb_error_newf`
- `zb_strndup`
- `zb_secure_zero`
- `zb_secure_free`

These are called from `sdk.c` and/or `stream.c`, so they cannot be `static` in a `.c` without becoming `static inline` in the header (or compiling as one translation unit). If they stay separate objects, a namespaced prefix would keep them from colliding as `zb_*` globals:

- `zb_fail`
- `zb_is_empty`
- `zb_is_valid_string`
- `zb_is_valid_json`
- `zb_table_name_is_valid`
- `zb_url_validate`
- `zb_replace_string`
- `zb_strdup_view`
- `zb_strdup`
- `zb_secure_free_cstr`

Could we hide the first group with `ZB_STATIC`, and either inline or prefix the second, so the shipped archive only exports `ZEROBUS_API` entry points?

purec/src/utils.c
```diff
@@ -0,0 +1,533 @@
...
+        rest = endpoint + 8;
+        is_https = true;
+    } else if (strncmp(endpoint, "http://", 7) == 0) {
+        rest = endpoint + 7;
+        is_https = false;
+    } else {
+        return ZB_URL_NO_SCHEME;
+    }
+
+    /* The authority ends at the first '/', '?' or '#'. */
```
teodordelibasic-db: Validation only checks scheme and host, then [`sdk.c:117`](https://github.com/databricks/zerobus-sdk/blob/94168aede4700d6901159e19a6600bf7acc8c056/purec/src/sdk.c#L117) stores the original string, including any path, query, or fragment. Tests currently treat `/path`, `?x=1`, and `#frag` as valid.

Once we append our own paths, that leftover suffix moves the request:

```text
stored:  https://host.example.com?x=1
append:  /oauth/token
actual:  https://host.example.com?x=1/oauth/token
```

A fragment is worse: the suffix never leaves the client, so a token POST can hit `/` instead of `/oauth/token`.

Could we reject path, query, and fragment, and store only `scheme://host[:port]`? The cases at [`utils_test.c:213`](https://github.com/databricks/zerobus-sdk/blob/94168aede4700d6901159e19a6600bf7acc8c056/purec/tests/src/utils_test.c#L213) and [`utils_test.c:289`](https://github.com/databricks/zerobus-sdk/blob/94168aede4700d6901159e19a6600bf7acc8c056/purec/tests/src/utils_test.c#L289) would need to become rejections.

teodordelibasic-db (commented):
Should we wire pure C into CI now or did you plan it for a later PR? The list of things to be done to integrate into CI:

- `purec/**` path filter
- `ci-purec.yml`
- Updating the `gate` job so it `needs` a Pure C job.

Easiest is to look at pure Go.

```yaml
            purec:
              - 'purec/**'
              - '.github/workflows/ci-purec.yml'
```

```yaml
  purec:
    needs: changes
    if: needs.changes.outputs.purec == 'true'
    uses: ./.github/workflows/ci-purec.yml
```

purec/src/error.h
danilotrninic-db: Not a duplicate — it's the public/internal header split. `include/` holds the public API: installed headers that declare only the `zerobus_*` functions a consumer is meant to use, over opaque handle types. Headers in `src/` are internal — not installed, used only within the library to share struct layouts and `zb_*` helpers across source files.

For error specifically:
- `include/zerobus/error.h` is the consumer-facing view: the opaque `zerobus_error_t` plus `zerobus_error_message` and `zerobus_error_free`.
- `src/error.h` is the implementation side: the `struct zerobus_error` layout (`code`/`message`/`message_len`) and the `zb_error_newf`/`zb_fail` constructors that every entry point uses.

I'll write this convention up in the documentation.

purec/src/utils.h
danilotrninic-db: I'd keep it in `src/` — same reasoning as the `error.h` thread.

purec/include/zerobus/sdk.h
danilotrninic-db: Fair point — today the stream only reads const endpoint config from the SDK, so copying that at build time works.

It's the shared state the SDK will own later that's the problem. For example, the token cache: the SDK owns and frees it, so a stream can't just hold a pointer to it (that just relocates the same dangling-pointer bug), and copying it defeats the point of a shared cache.

So my plan was to reference-count the SDK handle instead: free order stops mattering, and teardown runs when the last stream/builder is gone. It needs an atomic count, so I've left it for when we implement the internals with the concurrency library. There are still a few details to work out — it's not a fully formed plan yet. Happy to compare the two once we get there.

danilotrninic-db:
I'd like to do CI as a small follow-up right after this lands, to keep this PR focused on the skeleton — possibly together with the documentation write-up.

purec/src/stream.c
danilotrninic-db: Done. `ingest` and `flush` now return `ZEROBUS_STATUS_UNIMPLEMENTED` — those are the two operations that are genuinely no-ops until the core lands. I kept them returning the bare status without touching `*out_error`, so callers (and the tests) don't have to free anything and a reused `err` stays `NULL`.

I deliberately left the object-producing calls returning `OK`: the builders, the setters, both `build`s (they hand back a usable handle) and `close` (its observable contract — marked closed, later ingest rejected, idempotent — is real; only the internal flush/teardown is deferred). The example was reworked so an `UNIMPLEMENTED` from ingest/flush is the expected dev-build outcome rather than a failure path.

purec/include/zerobus/stream.h
danilotrninic-db: Added. `zerobus_stream_ingest_json_record` now takes a `zerobus_offset_t *out_offset` (new typedef in `common.h`, mirroring `zerobus_status_t`). It's there purely to complete the function's API — it does nothing today and it will only be used once the core lands.

purec/include/zerobus/stream.h
danilotrninic-db: Agreed, I dropped record JSON validation entirely. A record is now checked only for non-empty and size, its structure and encoding are the server's to reject. The recursive validator is gone, so nothing walks the record on the ingest path. `is_valid_utf8` stays, but it now runs only on the short config strings (endpoint, table, credentials), never on records.

purec/include/zerobus/sdk.h
danilotrninic-db: Fixed, the SDK now simply accepts both schemes: `http://` alongside `https://`. I also dropped the "host must contain a dot" rule, so single-label hosts like `localhost` pass. That unblocks running against a local shinkansen over plaintext h2c, and both the ingest and UC endpoints now share the same rule. A `no_tls()` setter might make sense later, once the transport lands.

purec/src/sdk.c
danilotrninic-db: Fixed across every entry point. Instead of `*out_error = NULL` on entry, each function now refuses to clobber a live pointer: if `out_error != NULL && *out_error != NULL` it returns `ZEROBUS_STATUS_INVALID_ARGUMENT` and leaves the slot untouched.

purec/examples/json/main.c
danilotrninic-db: Done. The example now reads the secret from `getenv("DATABRICKS_CLIENT_SECRET")` (and errors out if it's unset) instead of taking it as `argv[5]`. The usage string drops `<client-secret>` and names the env var instead.

purec/include/zerobus/error.h
danilotrninic-db: Added both. `zerobus_error_status()` returns the error's stored code (`ZEROBUS_STATUS_UNKNOWN` for a NULL error).

`zerobus_error_is_retryable()` is a **temporary placeholder**: it treats a fixed set of codes as non-retryable (`INVALID_ARGUMENT`, `UNAUTHENTICATED`, `PERMISSION_DENIED`, `NOT_FOUND`, `UNIMPLEMENTED`) and everything else as retryable. That classification is temporary and **will change once the transport core lands** and defines retryability precisely.

purec/src/utils.c
danilotrninic-db: Taken. `zb_strndup` now returns `NULL` when `len == SIZE_MAX`, before the `len + 1` can wrap to 0. For consistency I also rewrote the `is_valid_utf8` truncation check from `i + extra >= len` to `extra >= len - i`, so that add can't wrap either.

purec/include/zerobus/stream.h
danilotrninic-db: Documented. `stream.h` now states up front that a stream and its builder are not thread-safe: one caller thread per stream, no concurrent ingest/flush/close on the same stream, with internal synchronization arriving alongside the transport core.

purec/CMakeLists.txt
danilotrninic-db: As we discussed offline, no change on this — recapping the reasoning for the thread:

The first four (`zb_error_newf`, `zb_strndup`, `zb_secure_zero`, `zb_secure_free`) are internal helpers meant for cross-file use — unused from other files today, but I expect that to change as the SDK fills in, so I'd rather not make them `static`. If they stay unneeded, the fix is to delete them, not demote them.

The wider goal is that only `ZEROBUS_API`-decorated functions get exported and the `zb_*` helpers stay internal, which isn't fully enforced yet, and is best handled together with the cross-platform export work.

purec/src/utils.c
danilotrninic-db: Fixed. `zb_url_validate` now rejects any path, query, fragment, or even a trailing slash (`ZB_URL_HAS_PATH`), so the stored endpoint is guaranteed to be a bare `scheme://host[:port]` origin that's safe to append `/oidc/v1/token` to.

purec/src/sdk.c
```diff
@@ -0,0 +1,188 @@
...
+    char *endpoint;
+    char *uc_endpoint;
+};
+
+/* ---- builder ----------------------------------------------------------- */
+
+zerobus_status_t zerobus_sdk_builder_new(zerobus_sdk_builder_t **out_builder,
+                                         zerobus_error_t **out_error)
+{
+    if (out_builder != NULL) {
```
teodordelibasic-db, line 24: Maybe it makes sense to flip these two checks.

The live-error check runs after `*out_builder` is already set to NULL. Same order in [`sdk_builder_build`](https://github.com/databricks/zerobus-sdk/blob/8deecd92e755fc254c3167c7d30c9608c52759dd/purec/src/sdk.c#L131), [`stream_builder_new`](https://github.com/databricks/zerobus-sdk/blob/8deecd92e755fc254c3167c7d30c9608c52759dd/purec/src/stream.c#L35), and [`stream_builder_build`](https://github.com/databricks/zerobus-sdk/blob/8deecd92e755fc254c3167c7d30c9608c52759dd/purec/src/stream.c#L117).

If a caller still holds a builder (or SDK/stream) and passes a live `*out_error`, we drop the only pointer to that object without `*_free`. The tests start from a NULL handle, so they miss it.

Could we check the live error before touching the out-handle?

```c
    if (out_error != NULL && *out_error != NULL) {
        return ZEROBUS_STATUS_INVALID_ARGUMENT;
    }
    if (out_builder != NULL) {
        *out_builder = NULL;
    }
```

purec/src/error.c
```diff
@@ -0,0 +1,112 @@
...
+        return false;
+    }
+    switch (error->code) {
+    case ZEROBUS_STATUS_INVALID_ARGUMENT:
+    case ZEROBUS_STATUS_UNAUTHENTICATED:
+    case ZEROBUS_STATUS_PERMISSION_DENIED:
+    case ZEROBUS_STATUS_NOT_FOUND:
+    case ZEROBUS_STATUS_UNIMPLEMENTED:
+        return false;
+    default:
```
teodordelibasic-db, line 102: `FAILED_PRECONDITION` falls through to `return true`. We already use that code for ingest/flush after close and for build with missing endpoint, table, or OAuth. Retrying those with unchanged state cannot succeed, and they will not become transient later.

Could we treat `FAILED_PRECONDITION` as non-retryable, and say in the header that this table is still provisional?

```c
    case ZEROBUS_STATUS_FAILED_PRECONDITION:
    case ZEROBUS_STATUS_UNIMPLEMENTED:
        return false;
```

purec/Makefile
```diff
@@ -0,0 +1,97 @@
...
+# -Werror. Its own tree keeps the flag out of the normal build's cache.
+lint:
+	$(MAKE) build BUILD_DIR=build-lint CMAKE_EXTRA=-DZEROBUS_WERROR=ON
+
+check: fmt-check lint
+
+# Full workspace check: static checks, then the suite plain and under all three
+# sanitizers. address and thread can't coexist, so undefined rides with each.
+verify:
+	$(MAKE) clean
```
teodordelibasic-db, line 83: `verify` always runs `clean`, and `clean` is `rm -rf $(BUILD_DIR) build build-*` with `BUILD_DIR` documented as overridable. An override of `.` (or another source path) deletes the tree. The path is also unquoted.

Could we keep automatic cleanup inside known `build` / `build-*` dirs under `purec`, and quote the path?

```make
clean:
        rm -rf build build-*
```

This is meant just to avoid someone shooting themselves in the foot.

teodordelibasic-db (commented):
LGTM, a couple of nits.

purec/src/error.c
danilotrninic-db: Done. `FAILED_PRECONDITION` is non-retryable now, and the header notes the classification is provisional until the transport core defines retryability precisely.

purec/Makefile
danilotrninic-db: Done. `clean` no longer references `$(BUILD_DIR)`. On the unquoted-path point, I quoted the path expansions across `configure`/`build`/`test`/`coverage` too, so a `BUILD_DIR` with spaces now works (`cmake --build my project` was broken before). Help text now notes that a custom build dir named `build-*` is still cleaned.

purec/src/sdk.c
danilotrninic-db: First, why it works this way today. Each entry point NULLs `*out_X` before doing anything else, which keeps one simple guarantee: after any non-success return, the out-handle is NULL (for every handle except `out_error`). It's a standard C shape: `handle == NULL` on its own tells the caller the call failed. Doing the NULL first is what makes that hold on every path.

You've flagged that this also NULLs a caller's live handle before the guard returns. I see three ways to go:

1. **Keep it as-is.** `*out_X` is NULL after any failure, as today.
2. **Your suggestion — reorder.** Move the `*out_error` guard above the NULL, so a call rejected for a live `*out_error` leaves the handle untouched.
3. **Drop the NULLing.** Never write `*out_X` on failure — validate touching nothing and write it only on success.

I lean toward 3, for a few reasons:

- The reorder (2) only half-closes the gap: it protects the handle on the guard path, but any other failure path after it still NULLs the handle and then returns, so a live handle is still lost there. Only (3) leaves the handle untouched on *every* failure path.
- The uniform-NULL guarantee (1) buys less than it looks. The caller learns success or failure from the status and error, not from the handle, so `handle == NULL` is really only consulted after the status is already known — at which point its value on a failed call doesn't matter.
- (3) gives the two out-params a clean symmetric contract: each carries the outcome for one result — the error on failure, the handle on success — so the function writes whichever is relevant and leaves the other untouched. (On the OOM path we already return a bare status with no error object, so status is the authoritative channel there regardless.) It's also the familiar realloc-style rule of not destroying the caller's resource on a failed call.

What do you think?

teodordelibasic-db: Agreed, 3 is the better contract: write `*out_builder` only on success, leave it alone on every failure, live-`err` guard first. Status (and `*out_error`) is how the caller knows the call failed; they should not infer that from a NULLed handle.

Could we do that in this PR on `sdk_builder_new`, `sdk_builder_build`, `stream_builder_new`, and `stream_builder_build`? A test that starts with a live builder and a live error, and checks the builder pointer is unchanged, would lock it in.

danilotrninic-db: Done. The four factory functions now write the handle out-param only on success and leave it untouched on every failure path — the `*out_X = NULL` on entry is gone, so a failed call never overwrites whatever the caller left in the variable (and the pre-guard write goes with it). I documented the rule in `error.h` next to the `*out_error` contract, and added tests that seed the handle non-NULL and check each failure path preserves it.

teodordelibasic-db (approved):
LGTM

PR #851, opened 2026-09-10
[Python] Add Avro record format (avro feature)

irinatomic-db:
## PR stack

    - [avro/rust-core](https://github.com/databricks/zerobus-sdk/pull/827) [MERGED]
        - avro/java
        - [avro/python](https://github.com/databricks/zerobus-sdk/pull/851) <- _this PR_
        - avro/typescript

Adds the Avro record format to the Python SDK. Avro is compiled into the extension unconditionally (the `avro` feature is enabled on the Rust `databricks-zerobus-ingest-sdk` dependency), so it works on default `maturin build` / PyPI wheels — there is no cargo feature to select. Ephemeral streams only; feature in development.

Declare the writer schema once at stream creation with `TableProperties(table, avro_schema=<json>)`, then ingest through the existing `ingest_record_offset` / `ingest_records_offset`. An Avro stream accepts:
- a **record object** (dict) — encoded to a datum natively with `fastavro` against the writer schema;
- **pre-encoded Avro bytes** — passed through as-is.

The record format is **inferred from the schema on `TableProperties`** — a descriptor means proto, an `avro_schema` means Avro, neither means JSON. That inferred format is the source of truth for how each payload is coerced: every format accepts its native object or its pre-encoded bytes, and a payload that doesn't match the stream (e.g. a `str` on a proto stream) raises `TypeError` rather than being silently mis-encoded. `TableProperties` exposes it as a read-only `record_format` property (`"proto"`, `"json"`, or `"avro"`).

`StreamConfigurationOptions.record_type` is **optional** and defaults to the new `RecordType.UNSPECIFIED`, which means "infer from the schema" (existing code is unaffected). A client may still pass an explicit `record_type` (`PROTO` / `JSON` / `AVRO`), but it does **not** select the format — it is checked against the schema-inferred format and must agree, surfacing any mismatch as an early `ValueError` at stream creation. `RecordType` gains `UNSPECIFIED` and `AVRO`.

`fastavro` powers the dict-encoding path and installs via the `[avro]` extra (`pip install databricks-zerobus-ingest-sdk[avro]`); pre-encoded bytes need no extra. Both the sync and async (aio) trees are supported. See the README for the fastavro value conventions (an explicit `(branch, value)` tuple disambiguates an ambiguous union, `Decimal` scale, naive `datetime` treated as UTC).

## How is this tested?

```
cd python
make dev                      # builds the extension; installs fastavro for the dict path
pytest tests/                 # incl. tests/test_avro.py, tests/test_record_format.py
make lint && make fmt
```

Avro is always compiled in, so the Avro tests run on a default build (no `--all-features`).


python/rust/src/common.rs
```diff
@@ -68,13 +76,27 @@ pub struct TableProperties {
 
     // Internal field - stores the parsed DescriptorProto
     pub(crate) descriptor_proto: Option<prost_types::DescriptorProto>,
+
+    #[cfg(feature = "avro")]
+    #[pyo3(get)]
+    pub avro_schema: Option<String>,
```
teodordelibasic-db: For consistencies sake we can make this `pub(crate)` as well.

python/rust/src/common.rs
teodordelibasic-db: Preexisting, but we can make this `pub(crate)` as well.

python/rust/Cargo.toml
```diff
@@ -8,6 +8,9 @@ rust-version = "1.88"
 name = "_zerobus_core"
 crate-type = ["cdylib"]
 
+[features]
+avro = ["databricks-zerobus-ingest-sdk/avro"]
```
teodordelibasic-db: `#[cfg(feature = "avro")]` in `python/rust` is not something pip can select. The `[avro]` extra only installs `fastavro`. Release and abi3 `maturin build` use default features, so PyPI wheels reject `avro_schema`, while `make dev` (`--all-features`) is green. Arrow Flight is already always on the SDK dependency; Avro is the only cfg in this crate.

Could we add `avro` next to `arrow-flight` on the SDK dep, delete the cfg, and keep `[avro]` for fastavro only?

```toml
databricks-zerobus-ingest-sdk = { path = "../../rust/sdk", version = "2.8.0", features = ["arrow-flight", "avro"] }
```

python/rust/src/common.rs
```diff
@@ -68,13 +76,27 @@ pub struct TableProperties {
...
 }
 
 #[pymethods]
 impl TableProperties {
     #[new]
```
teodordelibasic-db, line 86: Passing both `descriptor_proto` and `avro_schema` is allowed. The stream opens as proto (descriptor wins), Python still Avro-encodes because `avro_schema` is set, and those bytes are accepted as proto. `record_format` reports `"proto"`.

Could we reject both at construction, and encode from the opened format rather than “`avro_schema` is not None”?

```rust
if rust_descriptor.is_some() && avro_schema.is_some() {
    return Err(PyValueError::new_err(
        "descriptor_proto and avro_schema are mutually exclusive",
    ));
}
```

A constructor-only test can cover this with no server: `TableProperties("t", descriptor, avro_schema="{...}")` should raise `ValueError`. Today that call succeeds.

python/NEXT_CHANGELOG.md
teodordelibasic-db, line 18: The intended model matches the code: `UNSPECIFIED` (the new default) infers from `TableProperties`; a set `PROTO` / `JSON` / `AVRO` must match that inference. Callers who omit `record_type` still get the same stream as today (no descriptor → JSON, descriptor → proto).

The break is narrower: `record_type` used to be ignored, so `StreamConfigurationOptions(record_type=RecordType.PROTO)` on a JSON table still opened JSON. That is the old documented default, and it now raises `ValueError` at [`resolve_format`](https://github.com/databricks/zerobus-sdk/blob/e7badced282b85667aae686edb45e5f4d93f931b/python/rust/src/common.rs#L267) before connect. The inverse (`JSON` on a descriptor table) fails the same way. v1.9.0 leaves Breaking Changes empty and puts this under API Changes.

Could we document that explicit-mismatch case under Breaking Changes with a migration note (omit the field, or set `UNSPECIFIED` / the type that matches `TableProperties.record_format`)? README and `config.py` still say the default is `PROTO` and the field is ignored, which would produce exactly this `ValueError` on JSON streams.

python/zerobus/sdk/shared/avro.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def _parse_schema(schema_json: str) -> Dict[str, Any]:
+    """Parse a schema JSON string and return a fastavro parsed schema."""
+    fastavro = _require_fastavro()
+    schema_dict = json.loads(schema_json)
+    return fastavro.parse_schema(schema_dict)
+
+
+def _encode(parsed_schema: Dict[str, Any], record: Dict[str, Any]) -> bytes:
```
teodordelibasic-db, line 86: The fidelity rules say naive `datetime` values are treated as UTC, but `_encode` hands them straight to `fastavro.schemaless_writer`. With `TZ=America/Los_Angeles`, naive `datetime(1970, 1, 1)` round-trips as `1970-01-01T08:00:00+00:00`, not the UTC epoch. The current test only checks that some bytes came back, and CI is usually UTC, so it cannot see the shift.

Could we walk the record and attach `timezone.utc` to naive datetimes before calling fastavro, or reject naive values and change the contract?

```python
if isinstance(value, datetime) and value.tzinfo is None:
    value = value.replace(tzinfo=timezone.utc)
```

`test_encode_logical_types` can assert the decoded instant with `TZ=America/Los_Angeles` set. Today it only checks `isinstance(..., bytes)`.

python/rust/src/common.rs
```diff
@@ -442,47 +552,78 @@ impl StreamConfigurationOptions {
...
+        Ok(bytes)
+    } else {
+        Err(PyErr::new::<pyo3::exceptions::PyTypeError, _>(
+            "Avro payload must be bytes (pre-encoded by the Python layer)",
+        ))
     }
 }
 
-pub(crate) fn extract_record_payloads(payloads: &Bound<'_, PyAny>) -> PyResult<Vec<EncodedRecord>> {
+pub(crate) fn extract_record_payloads(
```
teodordelibasic-db, line 605: The format-strict rewrite dropped the old non-list fallbacks (`Vec<Vec<u8>>` → proto, `Vec<String>` → JSON) and replaced them with `try_iter` so a tuple or other sequence of records still works. That also iterates a `dict` (keys) and a `str` (characters). On a JSON stream, `ingest_records_offset({"a": 1})` ingests the key `"a"`, and `ingest_records_offset('{"a":1}')` ingests characters. A list of records is unchanged.

Could we keep iteration for tuples and generators, but reject `str`, bytes-like objects, and mappings before that?

```rust
if payloads.cast::<pyo3::types::PyString>().is_ok()
    || payloads.cast::<PyBytes>().is_ok()
    || payloads.cast::<PyDict>().is_ok()
{
    return Err(PyErr::new::<pyo3::exceptions::PyTypeError, _>(
        "payloads must be a list of records",
    ));
}
```

python/zerobus/sdk/sync/zerobus_sdk.py
```diff
@@ -313,18 +335,22 @@ def create_stream(
             options: Optional stream configuration
             headers_provider: Optional custom headers provider (if set, overrides OAuth)
         """
+        from zerobus.sdk.shared.avro import parse_table_schema
```
teodordelibasic-db: Both public `create_stream` wrappers call `parse_table_schema`, which imports fastavro and parses the writer schema before the native stream exists. That runs even when the caller will ingest only pre-encoded bytes, so an Avro stream cannot be opened without the extra. The PR description says the bytes path needs no extra.

Could we store the raw schema on the wrapper and parse lazily only when a non-bytes payload needs encoding? Same for [`aio/zerobus_sdk.py:359`](https://github.com/databricks/zerobus-sdk/blob/e7badced282b85667aae686edb45e5f4d93f931b/python/zerobus/sdk/aio/zerobus_sdk.py#L359).

python/zerobus/sdk/shared/avro.py
```diff
@@ -0,0 +1,101 @@
+"""
+Avro record format support for the Zerobus SDK.
+
+Requires fastavro to be installed:
+    pip install databricks-zerobus-ingest-sdk[avro]
+
+Records can be passed as dicts (encoded to bytes via fastavro) or pre-encoded bytes.
+
+Fidelity rules:
```
teodordelibasic-db: The module says ambiguous unions such as `[int, long]` need an explicit `(branch_name, value)` tuple. There is no check; fastavro accepts a plain value and picks a matching branch. [`sync_example_avro.py`](https://github.com/databricks/zerobus-sdk/blob/e7badced282b85667aae686edb45e5f4d93f931b/python/examples/sync_example_avro.py#L1) claims pre-encoded bytes, unions, and logical types, then only sends dicts, and `reading_time` is `datetime.now().timestamp() * 1000` (a float) for a `long` / `timestamp-millis` field. The async example uses `int(...)`.

Could we document fastavro's value model (or encode in Rust so the rules match apache-avro), and fix the example?

python/README.md
teodordelibasic-db: README, [`config.py`](https://github.com/databricks/zerobus-sdk/blob/e7badced282b85667aae686edb45e5f4d93f931b/python/zerobus/sdk/shared/config.py#L55), and [`_zerobus_core.pyi`](https://github.com/databricks/zerobus-sdk/blob/e7badced282b85667aae686edb45e5f4d93f931b/python/zerobus/_zerobus_core.pyi#L126) still say `record_type` defaults to `RecordType.PROTO` and does not select the format. The implementation defaults to `UNSPECIFIED` and raises on mismatch. Following the current table produces a `ValueError` on JSON streams. Avro is an install snippet only; it is missing from “Choose Your Serialization Format”, `TableProperties`, and [`examples/README.md`](https://github.com/databricks/zerobus-sdk/blob/e7badced282b85667aae686edb45e5f4d93f931b/python/examples/README.md#L176).

Could we update those surfaces to default `UNSPECIFIED`, document that a non-unspecified value must match `TableProperties.record_format`, and add Avro next to proto/JSON/Arrow?

.github/workflows/ci-python.yml
```diff
@@ -76,7 +76,7 @@ jobs:
           .venv/Scripts/python -m pip install --upgrade pip
           test -f LICENSE || cp ../LICENSE LICENSE
           .venv/Scripts/python -m pip install -r requirements-dev.lock
-          .venv/Scripts/python -m maturin develop
+          .venv/Scripts/python -m maturin develop --all-features
```
teodordelibasic-db: If Avro is compiled into the extension unconditionally, this flag is no longer needed.

python/rust/Cargo.toml
irinatomic-db: Enabled `avro` on the `databricks-zerobus-ingest-sdk` dependency next to arrow-flight and deleted the avro cargo feature along with every `#[cfg(feature = "avro")]` in the crate. Release/abi3 `maturin build` now compiles Avro in, so PyPI wheels accept `avro_schema`.

The `[avro]` extra still only installs `fastavro` for the object path. Verified a plain `maturin develop` (default features) accepts `avro_schema` and reports `record_format == "avro"`.

python/rust/src/common.rs
irinatomic-db: Done. 

`TableProperties.__new__` now rejects the combination with `ValueError("descriptor_proto and avro_schema are mutually exclusive")`. With them mutually exclusive, `avro_schema is not None` unambiguously means an Avro stream, so the encode path stays consistent with the opened format. Added the constructor-only test as well.

python/NEXT_CHANGELOG.md
irinatomic-db: Moved it to Breaking Changes with a migration note (omit record_type - it defaults to UNSPECIFIED = or set it to match `TableProperties.record_format`). Also updated the docs that still said PROTO/ignored.

python/zerobus/sdk/shared/avro.py
irinatomic-db: Fixed. 

I checked the Rust path in first - AvroRecord takes raw `apache_avro::Value::TimestampMillis(i64)` (UTC millis since epoch, no tz logic). To match that, the encoder now stamps naive datetimes as UTC before `fastavro`, recursing through nested records/arrays and union (branch, value) tuples. 

Added a `TZ=America/Los_Angeles` test that encodes naive `datetime(1970,1,1)` and asserts it decodes to the UTC epoch (it shifted 8h before).

python/rust/src/common.rs
irinatomic-db: Done. 

Kept `try_iter` so tuples/generators still work, but reject `str`, bytes-like, and mappings up front with `TypeError("payloads must be a list of records")`.

python/zerobus/sdk/sync/zerobus_sdk.py
irinatomic-db: Done. 

The wrapper now holds an `AvroEncoder` that stores the raw schema and parses it lazily on the first dict/object payload - the pre-encoded-bytes path never imports `fastavro`. Applied to both sync and aio

python/zerobus/sdk/shared/avro.py
irinatomic-db: Reworded the value-conventions note to describe fastavro's actual behavior: it writes the first branch whose type matches, and an explicit `(branch_name, value)` tuple is only needed to disambiguate an ambiguous union. 

Went with documenting the fastavro model rather than re-encoding in Rust. Fixed the examples too - the `timestamp-millis` field now passes a `datetime` (was datetime.now().timestamp() * 1000, a float), and trimmed the sync example's docstring to what it actually demonstrates.

python/README.md
irinatomic-db: Updated all three surfaces. README, config.py, and _zerobus_core.pyi now document `record_type` defaulting to `UNSPECIFIED` and that a non-UNSPECIFIED value must match `TableProperties.record_format`. Added Avro to "Choose Your Serialization Format", the options/TableProperties docs, and examples/README.md.

.github/workflows/ci-python.yml
irinatomic-db: Good point. Dropped it.

python/rust/src/common.rs
teodordelibasic-db: The str/bytes/dict guard here never runs on an Avro stream: `ingest_records_offset` calls [`encode_batch`](https://github.com/databricks/zerobus-sdk/blob/dffa49b7755a6d9a36853f191d903548bd2db4fc/python/zerobus/sdk/shared/avro.py#L112) first, which does `[self.encode(p) for p in payloads]`. A dict iterates keys, a str iterates characters, `bytes`/`bytearray` iterate integers, and Rust then sees a list so it does not raise `TypeError("payloads must be a list of records")`. On a writer schema that can encode those elements, that silently becomes several records.

The native guard is also still exact-type only. `bytearray`, `memoryview`, and `collections.UserDict` skip it and hit `try_iter` on JSON too.

Could we reject strings, bytes-like objects, and mappings in `encode_batch` the same way, keep the Rust check, and broaden that check past concrete `str`/`bytes`/`dict`?

```python
def encode_batch(self, payloads):
    if isinstance(payloads, (str, bytes, bytearray, memoryview, dict)):
        raise TypeError("payloads must be a list of records")
    return [self.encode(p) for p in payloads]
```

`AvroEncoder.encode_batch({"a": 1})` (and a bare `str` / `bytes`) can assert `TypeError` with no server. Today a record schema usually explodes inside fastavro; after the guard it should be that TypeError.

python/zerobus/sdk/shared/avro.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def _parse_schema(schema_json: str) -> Dict[str, Any]:
+    """Parse a schema JSON string and return a fastavro parsed schema."""
+    fastavro = _require_fastavro()
+    schema_dict = json.loads(schema_json)
+    return fastavro.parse_schema(schema_dict)
+
+
+def _utc_naive_datetimes(value: Any) -> Any:
```
teodordelibasic-db, line 66: Every dict ingest now rebuilds every nested `dict`/`list`/`tuple` even when there is no datetime, then fastavro walks the record again. Could we copy-on-change (return the original object when nothing was naive) or only walk logical timestamp fields?

```python
if isinstance(value, dict):
    changed = False
    out = {}
    for k, v in value.items():
        nv = _utc_naive_datetimes(v)
        changed = changed or nv is not v
        out[k] = nv
    return out if changed else value
```

python/tests/test_avro.py
```diff
@@ -0,0 +1,219 @@
...
+
+Skipped if fastavro is not installed (it powers the dict-encoding path).
+"""
+
+import json
+import time
+
+import pytest
+
+# fastavro powers the object-encoding path; skip these tests when it is absent.
```
teodordelibasic-db, line 12: Nothing constructs `AvroEncoder`. The skip-if-no-fastavro also hides `TableProperties(..., avro_schema=...)`, which does not import fastavro. Lazy parse can regress with a green suite.

Could we unit-test bytes passthrough against an unparsable schema (proves `_parsed` stays `None`), and move the `record_format == "avro"` assertion into `test_record_format.py` so it always runs?

```python
from zerobus.sdk.shared.avro import AvroEncoder

def test_bytes_passthrough_does_not_parse_schema():
    enc = AvroEncoder("not-json")
    assert enc.encode(b"\x00") == b"\x00"
    assert enc._parsed is None
```

python/tests/test_avro.py
```diff
@@ -0,0 +1,219 @@
...
+        )
+        parsed = _parse_schema(schema_json)
+        from datetime import date, datetime
+
+        record = {"date_field": date(2025, 1, 1), "ts_millis": datetime(2025, 1, 1, 0, 0, 0)}
+        encoded = _encode(parsed, record)
+        assert isinstance(encoded, bytes)
+
+
+class TestAvroStreamIntegration:
```
teodordelibasic-db: The class constructs `TableProperties` or calls `_encode` and never goes through `ZerobusStream`. Could we drop the “stream integration” name unless it actually uses the public wrapper?

python/zerobus/sdk/shared/avro.py
teodordelibasic-db: [`_utc_naive_datetimes`](https://github.com/databricks/zerobus-sdk/blob/dffa49b7755a6d9a36853f191d903548bd2db4fc/python/zerobus/sdk/shared/avro.py#L65) only walks concrete `dict`, `list`, and `tuple`. fastavro still accepts other mappings and sequences. A `collections.UserDict` with naive `datetime(1970, 1, 1)` under `TZ=America/Los_Angeles` round-trips as `1970-01-01T08:00:00+00:00`, not the UTC epoch. The new test only covers a plain dict.

Could we normalize `collections.abc.Mapping` and sequence types (or reject containers the walker does not handle)?

```python
from collections.abc import Mapping, Sequence

if isinstance(value, Mapping):
    return {k: _utc_naive_datetimes(v) for k, v in value.items()}
if isinstance(value, Sequence) and not isinstance(value, (str, bytes, bytearray)):
    return type(value)(_utc_naive_datetimes(v) for v in value)
```

python/README.md
teodordelibasic-db: The [`TableProperties` snippet](https://github.com/databricks/zerobus-sdk/blob/dffa49b7755a6d9a36853f191d903548bd2db4fc/python/README.md#L549) is still JSON and Protobuf only (no `avro_schema=`, no mutual exclusion). [`examples/README.md`](https://github.com/databricks/zerobus-sdk/blob/dffa49b7755a6d9a36853f191d903548bd2db4fc/python/examples/README.md#L3) intro, `pip install -e ".[arrow]"`, the run list, and “two serialization formats” omit `sync_example_avro.py` / `async_example_avro.py`. [`_zerobus_core.pyi`](https://github.com/databricks/zerobus-sdk/blob/dffa49b7755a6d9a36853f191d903548bd2db4fc/python/zerobus/_zerobus_core.pyi#L29) still says `avro_schema` exists only with the cargo feature, which was deleted. README line 110 still reads as “Avro requires the extra,” which contradicts the bytes path and the changelog.

Could we add the Avro constructor next to proto, wire the examples README (install `[avro]`, run commands), and say the extra is only for dict encoding via fastavro?

python/NEXT_CHANGELOG.md
teodordelibasic-db: Tests check enum values and the `UNSPECIFIED` default, not the `ValueError` at `create_stream`. `resolve_format` runs before connect; [`test_exceptions.py`](https://github.com/databricks/zerobus-sdk/blob/dffa49b7755a6d9a36853f191d903548bd2db4fc/python/tests/test_exceptions.py#L63) already uses `create_stream` without a live server.

Could we add `StreamConfigurationOptions(record_type=RecordType.PROTO)` on `TableProperties("catalog.schema.table")` and assert `ValueError` matching `conflicts with the schema`?

```python
def test_record_type_proto_rejected_on_json_table():
    sdk = ZerobusSdk("https://example.zerobus.cloud.databricks.com", "https://example.cloud.databricks.com")
    opts = StreamConfigurationOptions(record_type=RecordType.PROTO, recovery=False)
    with pytest.raises(ValueError, match="conflicts with the schema"):
        sdk.create_stream("id", "secret", TableProperties("catalog.schema.table"), opts)
```

teodordelibasic-db:
[`README.md:102`](https://github.com/databricks/zerobus-sdk/blob/dffa49b7755a6d9a36853f191d903548bd2db4fc/README.md#L102)

The repo README ingestion list has no Avro, and the PR description still talks about an off-by-default cargo feature and `maturin --features avro`, both of which were removed. Could we add Avro next to JSON / Protobuf / Arrow here, and drop that cargo-feature wording from the PR body?

python/NEXT_CHANGELOG.md
irinatomic-db: Added `test_record_type_proto_rejected_on_json_table` to `python/tests/test_record_format.py`: `create_stream(record_type=PROTO)` on `TableProperties("catalog.schema.table")` (with recovery=False) asserts `ValueError` matching "conflicts with the schema", raised before connecting - no server needed.

python/zerobus/sdk/shared/avro.py
irinatomic-db: `_utc_naive_datetimes` now keys off `collections.abc.Mapping`/`Sequence` instead of concrete dict/list/tuple, so UserDict etc. are normalized. 

Added `test_naive_datetime_coerced_in_mapping_subclass` (UserDict) in `python/tests/test_record_format.py`.

python/rust/src/common.rs
irinatomic-db: Both done: 

1. `AvroEncoder.encode_batch` (`shared/avro.py`) rejects str/bytes/bytearray/memoryview/mappings with the same `TypeError,
2. Native `extract_record_payloads` guard now also catches bytearray, memoryview, and mapping subclasses (via hasattr("keys"), covering UserDict). 

Added `test_encode_batch_rejects_single_record` covering all those types in `python/tests/test_record_format.py`.

python/README.md
irinatomic-db: Got the rest now: 
1. added the `TableProperties(..., avro_schema=...)` snippet with the mutual-exclusion note (`python/README.md`)
2. wired `python/examples/README.md` (intro, `[arrow,avro]` install, run list, "three serialization formats" + an Avro section)
3. fixed the `_zerobus_core.pyi` comment (the cargo feature is gone), 
4. reworded `README:110` so the extra is only for dict encoding - bytes need no extra.

python/zerobus/sdk/shared/avro.py
irinatomic-db: `_utc_naive_datetimes` is now copy-on-change - it returns the original object when nothing was naive, so records without datetimes aren't rebuilt.

python/tests/test_avro.py
irinatomic-db: Added `test_bytes_passthrough_does_not_parse_schema` to `python/tests/test_record_format.py` (runs without fastavro): `AvroEncoder("not-json").encode(b"\x00")` returns the bytes and asserts `_parsed is None`. Moved the `record_format == "avro"` assertion there too (`test_avro_format_from_schema`).

python/tests/test_avro.py
irinatomic-db: Renamed the class to `TestAvroEncodingFlow` and reworked it to exercise `AvroEncoder` directly (dict encode + mixed dict/bytes batch); the creation assertion moved to `test_record_format.py`.

irinatomic-db:
> [`README.md:102`](https://github.com/databricks/zerobus-sdk/blob/dffa49b7755a6d9a36853f191d903548bd2db4fc/README.md#L102)
> 
> The repo README ingestion list has no Avro, and the PR description still talks about an off-by-default cargo feature and `maturin --features avro`, both of which were removed. Could we add Avro next to JSON / Protobuf / Arrow here, and drop that cargo-feature wording from the PR body?

Added Avro to the repo README ingestion list, scoped as beta ("currently available in the Rust and Python SDKs; support in the other SDKs is in development") so it doesn't imply the SDKs that don't have it yet. And updated the PR description to drop the off-by-default cargo-feature / `maturin --features avro` wording — Avro is compiled in unconditionally now.

README.md
```diff
@@ -101,6 +101,7 @@ Pick the record format that matches your data.
 
 - JSON: schema-free ingestion. Pass a JSON string or a native object (dict, map, and so on) and the SDK serializes it. No compilation step. Good for getting started or dynamic schemas.
 - Protocol Buffers: strongly-typed, schema-validated ingestion. More compact on the wire than JSON. A typical choice for production workloads that are not already producing Arrow.
+- Avro (beta): row-oriented binary encoding against a declared writer schema. Pass a record object or pre-encoded Avro bytes. Currently available in the Rust and Python SDKs; support in the other SDKs is in development.
 - Arrow Flight: Apache Arrow `RecordBatch` data over the Arrow Flight protocol. Best when the workload is columnar or batched, or the application already produces Arrow (pyarrow, [arrow-rs](https://github.com/apache/arrow-rs), DataFusion, Polars).
 
 JSON and Protocol Buffers share one stream API, available in every SDK. Arrow Flight is a separate columnar API, available in the SDKs listed below.
```
teodordelibasic-db: We could probably update this file a bit more, i.e. this sentence,  the table below and expand/add a section below the table.

python/README.md
```diff
@@ -107,13 +107,21 @@ installs `pyarrow` 22.0.0 or later, which is the first release with 3.14 wheels.
...
+```
+
 ## Quick Start
 
 ### Choose Your Serialization Format
 
 1. **Protocol Buffers** (Recommended) - Strongly-typed schemas with compact binary encoding. More efficient over the wire and the best choice for production and high-throughput workloads.
 2. **JSON** - Simple, no schema compilation needed. Good for getting started or quick prototyping, but each record carries higher per-record overhead (text serialization plus UTF-8 validation), so it is slower than Protocol Buffers for high-volume ingestion.
-3. Arrow Flight - High-throughput columnar ingestion for applications that already produce Arrow data. Install the `arrow` extra and see the [sync](examples/sync_example_arrow.py) or [async](examples/async_example_arrow.py) example.
+3. **Avro** (beta) - Row-oriented binary encoding against a declared writer schema. Install the `avro` extra and see the [sync](examples/sync_example_avro.py) or [async](examples/async_example_avro.py) example.
```
teodordelibasic-db: the pip optional extra `[avro]` only installs `fastavro`. Avro itself is already in the wheel.

Two ingest paths:

- dict records: encoded with `fastavro`, so `[avro]` is needed
- pre-encoded Avro bytes: passed through with no `fastavro`, so a plain `pip install databricks-zerobus-ingest-sdk` is enough

Lines [110-111](https://github.com/databricks/zerobus-sdk/blob/1321f2fc0156dbdbc62b3728b9772a322ed08d41/python/README.md#L110) already say that. 

Could we match that here with something like: dicts need `fastavro` via `[avro]`; pre-encoded bytes need no extra.

README.md
irinatomic-db: Added it to the table below and avro has it's own passage now as well

python/README.md
irinatomic-db: Expanded the test to explain 2 ingest paths and whether they need to install fastavro or not

python/rust/src/common.rs
```diff
@@ -442,47 +531,87 @@ impl StreamConfigurationOptions {
...
-    } else if let Ok(json_list) = payloads.extract::<Vec<String>>() {
-        for json in json_list {
-            out.push(EncodedRecord::Json(json));
-        }
-    } else {
+    } else if payloads.cast::<PyString>().is_ok()
+        || payloads.cast::<PyBytes>().is_ok()
+        || payloads.cast::<PyByteArray>().is_ok()
+        || payloads.cast::<PyMemoryView>().is_ok()
+        || payloads.hasattr("keys")?
```
teodordelibasic-db: ingest_records_offset` on a pandas Series of records should ingest those records. JSON/proto raise `TypeError("payloads must be a list of records")` and Avro ingests one record per Series value, with no error, so we have the same call, with two wrong answers.

This is because Rust treats [`hasattr("keys")`](https://github.com/databricks/zerobus-sdk/blob/1321f2fc0156dbdbc62b3728b9772a322ed08d41/python/rust/src/common.rs#L603) as “this is one dict.” A Series has `.keys` (it is `.index`) but it is not one mapping. Avro [`encode_batch`](https://github.com/databricks/zerobus-sdk/blob/1321f2fc0156dbdbc62b3728b9772a322ed08d41/python/zerobus/sdk/shared/avro.py#L123) uses `collections.abc.Mapping`, so it iterates the Series and Rust never sees the original object.

We can drop `hasattr("keys")` and use `Mapping`, same as `encode_batch`. A dict / `UserDict` still raises; a Series stays a batch.

```rust
let mapping = payloads.py().import("collections.abc")?.getattr("Mapping")?;
if payloads.cast::<PyString>().is_ok()
    || payloads.cast::<PyBytes>().is_ok()
    || payloads.cast::<PyByteArray>().is_ok()
    || payloads.cast::<PyMemoryView>().is_ok()
    || payloads.is_instance(&mapping)?
{
```

python/zerobus/sdk/shared/avro.py
```diff
@@ -0,0 +1,130 @@
...
+    def __init__(self, schema_json: str):
+        self._schema_json = schema_json
+        self._parsed: Optional[Dict[str, Any]] = None
+
+    def _schema(self) -> Dict[str, Any]:
+        if self._parsed is None:
+            self._parsed = _parse_schema(self._schema_json)
+        return self._parsed
+
+    def encode(self, payload: Any) -> bytes:
```
teodordelibasic-db, line 116: `ingest_record_offset(memoryview(avro_bytes))` should send those bytes through, same as `bytes` / `bytearray`. Instead the view is treated as an object record, and encoding fails (or the datetime walker turns it into a list of ints). `ingest_records_offset(memoryview(...))` already TypeErrors, which is correct for a batch call.

This happens because [`encode()`](https://github.com/databricks/zerobus-sdk/blob/1321f2fc0156dbdbc62b3728b9772a322ed08d41/python/zerobus/sdk/shared/avro.py#L116) only passthroughs `bytes`/`bytearray`. A `memoryview` falls into `_encode`. The walker then treats it as a `Sequence` because [`avro.py:78`](https://github.com/databricks/zerobus-sdk/blob/1321f2fc0156dbdbc62b3728b9772a322ed08d41/python/zerobus/sdk/shared/avro.py#L78) excludes `str`/`bytes`/`bytearray` but not `memoryview`. `encode_batch` already lists `memoryview` next to `bytes`.

We can treat `memoryview` as bytes-like in `encode()`, and skip it in the Sequence walk.

```python
def encode(self, payload):
    if isinstance(payload, (bytes, bytearray, memoryview)):
        return bytes(payload)
    return _encode(self._schema(), payload)
```

```python
if isinstance(value, Sequence) and not isinstance(value, (str, bytes, bytearray, memoryview)):
```

`AvroEncoder("not-json").encode(memoryview(b"\x00"))` can assert `b"\x00"` and `_parsed is None`, same as [`test_bytes_passthrough_does_not_parse_schema`](https://github.com/databricks/zerobus-sdk/blob/1321f2fc0156dbdbc62b3728b9772a322ed08d41/python/tests/test_record_format.py#L40).

teodordelibasic-db (commented):
I think only these 2 remain.

python/rust/src/common.rs
irinatomic-db: Thanks! 

A good reminder (for me in the future) that clients use a wider range of container types than I'd mapped. Guard now uses `isinstance(collections.abc.Mapping)` instead of `hasattr("keys")`: `dict`/`UserDict` still raise, a `Series` iterates as a batch - consistent with `encode_batch`.

python/zerobus/sdk/shared/avro.py
irinatomic-db: Thanks!

I'd modeled bytes-like as just `bytes`/`bytearray` and hadn't accounted for `memoryview`. `encode()` now passes it through too, and the datetime walker excludes it. Extended the passthrough test to cover it.

teodordelibasic-db (approved):
LGTM

PR #856, opened 2026-09-18
[PureC] Add CI, logging, and harden tests

danilotrninic-db:

Adds CI coverage for the Pure C SDK, makes existing tests handle failed preconditions safely, and introduces opt-in internal logging.

### CI

- Add a reusable `ci-purec.yml` workflow with separate formatting, warnings-as-errors build/test, ASan+UBSan, and TSan+UBSan jobs.
- Wire it into `push.yml`.

### Test hardening

- Add `REQUIRE` for preconditions that must succeed before a test can continue. Failures jump to cleanup, `CHECK` remains nonfatal for independent assertions.
- Apply it to stream and secure-memory tests so failed setup does not lead to invalid accesses or skipped cleanup.
- Update coverage collection to include nested source directories.

### Logging

Add internal logging for SDK lifecycle events and errors. Configurable through two environment variables: `ZEROBUS_LOG_LEVEL` selects `off` (default), `error`, `warn`, `info`, `debug`, or `trace` and `ZEROBUS_LOG_FILE` redirects output from stderr to a file. Configuration is read once on first use and stays fixed for the run.

## How is this tested?

Added unit tests for log filtering and file output, and manually checked configuration and error handling. The full test suite and sanitizer checks pass locally. The new CI workflow still needs to be verified on GitHub.


purec/src/internal/log.c
```diff
@@ -0,0 +1,87 @@
...
+            }
+        }
+    }
+    config.initialized = true;
+    errno = saved_errno;
+}
+
+bool zb_log_enabled(enum zb_log_level level)
+{
+    if (!config.initialized) {
```
teodordelibasic-db: I see the next PR says logger synchronization is a follow-up, but if you want maybe we could protect first-use initialization with `pthread_once` here? Suppressed messages still initialize the shared configuration, so separate SDK instances can race even with logging disabled.

For example, replace the `config.initialized` check with:

```c
#include <pthread.h>

static pthread_once_t log_once = PTHREAD_ONCE_INIT;

bool zb_log_enabled(enum zb_log_level level)
{
    (void)pthread_once(&log_once, zb_log_init);
    return level > ZB_LOG_OFF && level <= ZB_LOG_TRACE && level <= config.level;
}
```

The `initialized` field and its assignment can then be removed. `pthread_once` makes other callers wait for initialization to finish and makes the initialized configuration visible to them. This works with C99 but needs the CMake `Threads::Threads` dependency, which #868 already introduces. When integrating that PR, the once primitive can live behind its private concurrency interface.

teodordelibasic-db (dismissed):
LGTM, if you want you can address the comment either here or in a follow up.

purec/src/internal/log.c
danilotrninic-db: Added `pthread_once` for initialization and the `Threads::Threads` dependency. I also added a mutex around the complete write, so messages from different threads don’t interleave.

teodordelibasic-db (approved):

